Issue #11, opened 2021-12-31
select vs sel


d-chambers:
**Description**
slack message from Jin:

It seems that the select function cannot take a list of sample locations.

![image](https://user-images.githubusercontent.com/11671536/147839828-1d7c2a5a-f5df-4235-a022-0d775c1120a8.png)

d-chambers:
@jinwar,

This is indeed a bug. However, `select` should raise an error when passed an array of sample locations. I don't want to support this behavior because of how easily it breaks the assumption of even sampling in each dimension. Consider the following example:

```python
import numpy as np

import dascore as dc
# get example patch
patch = dc.get_example_patch()

# load time array, take irregular chunks of it.
time = dc.coords['time']
time_indices = np.array([1,2,3,5,12,40)
new_time = time[time_indices]

# now calling `patch.select` with new_time will create an unevenly sampled patch in the time dimension
patch_uneven = patch.select(time=new_time)
 
# However, if we restrict `select` to only a tuple of ranges (start, stop), it cannot change the sampling interval
patch_event = patch.select(time=(time[0], time[10]) )
```

It's not that we *cant* support patches with uneven sampling rates, it will just require more work since they break many of the patch methods/transforms (eg fft, pass_filter).

Moreover, If we supported both styles it introduces ambiguity with 2 length sequences. For example:

```python
patch.select(time=(time[0], time[10])) 
```
Might mean return a patch with only time slices from time[0] and time[10], or it might mean return a range for all time samples between time[0] and time[10]. Of course, we could go the numpy route where different objects have different meanings (eg, tuples vs lists vs arrays) but now it is getting very complicated.

d-chambers:
I think there should be an `iselect` method which simply works on ranges of indices though, which may alleviate or remove the need for using arrays in `select`. 

```python
import dascore as dc

patch = dc.get_example_patch()

# slice patch along first 10 samples
out = patch.iselect(time=(0, 10)) 
```

jinwar:
The function I would like to have is a resampling of the data at certain time/spacing points. I added a set and iset methods just to pass the arguments to the xarray.sel and xarray.iset. We may want to think what we can implement these. 

Here is what I have for my low-frequency processing package:

            lfDAS = DASdata.pass_filter(time=(None,1/dt/2*0.9))\
                            .sel(time=time_grid[data_end-buff_size:new_data_end-buff_size]
                                    ,method='nearest')

This will also allow increasing the sampling rate.

[...]

PR #12, opened 2022-01-03
Add pickle and improve filter

jinwar:
adding small changes to the code


dascore/core/patch.py
```diff
@@ -30,16 +30,19 @@ def __init__(
...
+                return
+            dims = dims if dims is not None else list(coords)
+            mixer = _AttrsCoordsMixer(attrs, coords, dims)
+            attrs, coords = mixer()
+            # get xarray coords from custom coords object
+            if isinstance(coords, Coords):
+                coords = coords._coords
+            self._data_array = DataArray(
+                data=data, dims=dims, coords=coords, attrs=attrs
+            )
```
d-chambers, line 45: Hey @jinwar, so I added a few tests in tests/core/patch.py. I decided to create a new test case (class) since there will likely be several tests around empty patches (a common edge case).

It looks like this:

```python
class TestEmptyPatch:
    """Tests for empty patch objects."""

    def test_empty_patch_str(self):
        """An empty patch should have both a str and repr of nonzero length."""
        patch = Patch()
        assert len(str(patch))
        assert len(repr(patch))

    def test_has_attrs(self):
        """An empty test patch should have attrs with null(ish) values."""
        patch = Patch()
        attrs = patch.attrs
        assert len(attrs)
        # Check that default values exist but are null (of appropriate dtype)
        names = ['time', 'distance']
        fields = ['d_{x}', '{x}_min', "{x}_max"]
        for name in names:
            for field in fields:
                value = attrs[field.format(x=name)]
                assert pd.isnull(value)
```

dascore/core/patch.py
```diff
@@ -139,12 +142,19 @@ def attrs(self) -> FrozenDict:
...
     detrend = dascore.proc.detrend
     pass_filter = dascore.proc.pass_filter
     aggregate = dascore.proc.aggregate
     abs = dascore.proc.abs
 
+    # --- io funcs
+
+    save_pickle = dascore.utils.save_pickle
+    load_pickle = dascore.utils.load_pickle
+
```
d-chambers, line 157: We should create a new io module for pickle. However, I was just thinking about refactoring this so we can leave this until we refactor. 

Normally IO would look like this (in ObsPy)
```python
import dascore as dc
patch = dc.read(picke_path, format='pickle')  # format is optional, it can auto-detect
patch.write("output_path.pkl", "pickle")
```

dascore/core/patch.py
d-chambers: I had to refactor the `Patch.__init__` method slightly to get the tests to pass.

d-chambers:
Hey @jinwar, sorry about this but there were a few file ending issues (differences between linux and windows) that mangled some of the files, which is why it shows so many files changed, but it should be good when we merge. 

I added some tests for the empty patch and started a io.pickle module. However, I just had a good idea for refactoring the IO to allow for easily adding parsers (as we discussed last week) so I am going to merge this for now then refactor a bit in a subsequent PR.

d-chambers:
I moved the `isel` and `sel` methods to a new PR #14 as I think they merit some more consideration in light of the conversation in #11

PR #14, opened 2022-01-12
sel and isel

d-chambers:
A pull request implementing `sel` and `isel` on `Patch` object which just passes info to `xarray.sel` or `xarray.isel`.

PR #19, opened 2022-01-20
Resample

d-chambers:
Adds a few methods around resampling and interpolation between samples.

I am hoping this PR will circumvent the need to expose `isel` or `sel` directly (as in #11 and #14). 

For example, this (from #11):
```python
     lfDAS = DASdata.pass_filter(time=(None,1/dt/2*0.9))\
                      .sel(time=time_grid[data_end-buff_size:new_data_end-buff_size]
                              ,method='nearest')
```
I think can be replaced by something like this:

```python
import numpy as np
import dascore as dc

# downsample to 1 ms
patch = (
    dc.get_example_patch()
    .resample(time=np.datetime64(1, 'ms')
)
```

There is also an interpolate that will behave similarly to `sel` (but I think the name will be more descriptive of what it does).

```python
import numpy as np
import dascore as dc

# interpolate new grid points for distance dim
distance = [10, 15, 20, 25, 30]
patch = (
    dc.get_example_patch()
    .interpolate(distance=distance, method='nearest')
)

```


d-chambers:
@jinwar Do you think this will work instead of `sel` and `isel` for your use case?

d-chambers:
I am going to merge this for now because I need some of the utils for another branch. We can change later if needed (we are still early in the development phase)

jinwar:
@d-chambers In my case I need to get the time sample value at a certain points of time, which seems could be accomplished by a recombination of resample and interpolate

jinwar:
@d-chambers I agree that having sel and isel could risk the even temporal sampling assumption of Patch. I'm ok with removing it in future.

jinwar:
Your current resample does not apply an anti-aliasing filter if the new sampling interval is larger than the current one. We could add an option to the function.

d-chambers:
> Your current resample does not apply an anti-aliasing filter if the new sampling interval is larger than the current one.

This gets kinda deep in the DSP weeds but I found a nice SO post on that [here](https://dsp.stackexchange.com/questions/62177/downsample-resample-vs-antialias-fitlering-decimation). It isn't that resample doesn't handle the aliasing issue, it just might not be the best approach depending on the application.

d-chambers:
I think we should also add a method based on scipy's  [resample_poly](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample_poly.html). Eventually, we should add a section in the docs all about changing sampling rates.  This section should show plots of the different approaches and provide links for additional resources. I will certainly need to dust off my DSP textbook before undertaking that though.

PR #20, opened 2022-02-05
Dascore format

d-chambers:
This PR adds support for the dasdae format. The format is experimental and loosely based on the [adaptable seismic data format](https://academic.oup.com/gji/article/207/2/1003/2583765) 

As it stands now the format looks like the following:
```
/root
/root.attrs
    __format__ = "DASDAE"  
    __DASDAE_version__ = 'x.y.z'  # version str
/root/waveforms/
    DAS__{net}__{sta}__{tag}__{start}__{end}
        data   # patch data array
        data.attrs
        _coords_{coord_name}  # each coordinate array is saved here
    DAS__{net}__{sta}__{tag}__{start}__{end}.attrs
        _attrs_{attr_nme}  # each patch attribute
        _dims  # a str of 'dim1, dim2, dim3'
 ```
following asdf convention, we can then add event information, station metadata, etc under different name spaces (eg `/root/events`) .

Of course these are all implementation details; all the user sees is this:

```python
import dascore 

patch = dascore.get_example_patch()
patch.io.write('some_path.h5', 'dasdae')
patch2 = dascore.read('some_path.h5')
```

The internal DASDAE format version string will allow us to iterate quickly while maintaining backwards compatibility (if we really need to).

### DIFF
--- a/dascore/constants.py
+++ b/dascore/constants.py
@@ -43,13 +43,14 @@ class PatchSummaryDict(TypedDict):
     distance_max: float
     instrument_id: str
     dims: str
+    tag: str
 
 
 # The expected attributes for the Patch
 DEFAULT_PATCH_ATTRS = {
     "d_time": np.NaN,
     "d_distance": np.NaN,
-    "data_type": "",
+    "data_type": "DAS",
     "data_units": "",
     "category": "",
     "time_min": np.datetime64("NaT"),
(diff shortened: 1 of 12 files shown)


d-chambers:
hmmm... odd the CI isn't triggering

d-chambers:
Looks like GH actions is at least partially down. I will check back later.

image

d-chambers:
Even though the file format will still probably need some tweaking, I am going to merge for now so I can work on the spool.

jinwar:
We can start with this format. Once iris publish their das data format
white paper we may want to implement some attributes later on.

PR #32, opened 2022-08-17
Spool class

d-chambers:
This PR implements the long anticipated `Spool` class, a container for `Patch` instances. Implemented here are the `MemorySpool` and `FileSpool` for working with in-memory and on-disk patches, respectively. 

It also includes some small bug fixes and tweaks to some of the API. 

TODO
---------
- [x] Update documentation (add spool section)
- [x] Additional indexing tests
- [ ] ~~Look into lazy, single file spool for formats which support multiple patches in one file~~ (better for different PR)
- [x] Create a simple `spool` function for getting the correct spool subclass from various inputs


d-chambers:
This turned out to be a larger PR than I would have liked, with many changes/bug fixes. There are certainly still some improvements I would like to make but I will work to address these in future PRs so this can get merged into master.

PR #40, opened 2022-08-25
Add support for TDMS format

aissah:


d-chambers:
Hey @aissah, 

The changes I made had to do with making this compatible with master, fixing a few warnings,  and renaming silixa to tdms as we discussed. I also added tdms to the entry points in the pyproject.toml file.

 Let's wait and see how the coverage report looks to make sure we don't have too much untested code in here, then I will try to take a closer look at it tomorrow but I don't anticipate there will be any major issues.

Since I re-based this branch to master you will have to ["force pull'](https://stackoverflow.com/a/8888015/3645626) the branch back to your local machine if you want to make additional changes.

Also, @eileenrmartin mentioned we ought to consider including the Silixa licence info from the original script you looked at as a reference. The right place for this would be the dascore.io.tdms.__init__.py docstring.  What do you think?  

Really great work!

d-chambers:
Ah, it looks like there might be some code coverage issues (meaning there are parts of the code that don't get executed during the test run) as this dropped our total test coverage down by almost a percentage point. 

The biggest part is in dascore.io.tdms.utils lines 346 to 396. You can see these aren't executed by clicking on the code coverage report then "show diff" for that file. Any idea how we can test this code path?

aissah:
The metadata provides start and end offsets in the binary file for the raw data. But it seemed from the original script that the raw data might sometimes be provided in two segments (2 start and 2 stop offsets). Those lines between 346 and 396 provide for the situation where there are two segments which isn't the case for any of the tdms files I have come across. 

One solution would be to modify the sample file to make sure that area is covered. Or I could make that portion into a function that gets called in each so that the current sample file still goes through it. I will explore both potential solutions and see what I come up with.

As for the license info, I can come by and show you what is in there and we can figure out what to do with it.

d-chambers:
Hey @aissah, nice job!

I re-based to master (yet again) then I added tests for the other TDMS file Silixa provided and made a few small tweaks. It should be good to merge once CI is green.

d-chambers (approved):

PR #90, opened 2022-12-20
Median filter

jinwar:


d-chambers (approved):
Hey @jinwar Looks great! 

We will add more specific tests if bugs crop up, but I don't expect many will since it just wraps scipy's function.

jinwar:
Sounds great!

PR #92, opened 2022-12-20
Adding standardize function

ahmadtourei:
Added standardize function to processing methods

### DIFF
--- a/tests/test_proc/test_basic.py
+++ b/tests/test_proc/test_basic.py
@@ -1,6 +1,7 @@
 """Tests for basic patch functions."""
 
 import numpy as np
+import pandas as pd
 import pytest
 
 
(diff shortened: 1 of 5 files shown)


tests/test_proc/test_basic.py
```diff
@@ -66,3 +67,12 @@ def test_max(self, random_patch):
...
         assert np.allclose(norm, 1)
+
+
+class TestStandarize:
+    """Tests for standardization."""
+
+    def test_base_case(self, random_patch):
+        """Ensure runs with default parameters."""
+        out = random_patch.standardize("time")
+        assert not np.any(pd.isnull(out.data))
```
d-chambers, line 78: Can you also add an assert that all the standard deviations along the time dimension are unity?

d-chambers (changes requested):
I think we just need one small change to the tests than it should be good to go.

tests/test_proc/test_basic.py
ahmadtourei: Done!

d-chambers (approved):

PR #93, opened 2022-12-20
Sobel

SeunghooKim:
Create a sobel filter using scipy.ndimage.sobel


d-chambers (changes requested):
A few suggestions:

1) We don't actually need kwargs here, usually they are used for specifying a quantity related to a dimension. For example in `Patch.pass_filter` kwargs are used to specify the dimension *and* the values for it (`patch.pass_filter(time=(1, 10))`) . When we don't need kwargs for that kind of flexibility, it is much better to have explicit parameters.

So the signature for `sobel_filter` can just be:

```python
def sobel_filter(patch: PatchType, dim: str,  mode='reflect', cval=0.0):
```

Then just copy the mode and cval arguments description from scipy for inclusion in the docs. 

You can get the axis corresponding to the selected dimension like this:

`axis = patch.dims.index(dim)`

2) Add some explanation in the docstring that the sobel filter is only applied along one axis, then add a third example of applying it along both time and space. I found [this so question](https://stackoverflow.com/questions/7185655/applying-the-sobel-filter-using-scipy) helpful. The example might look something like this:

```
    >>>  # 3. Apply sobel filter over space and time
    >>> sobel_time_space = pa.sobel_filter('time',).sobel_filter('space')
```

Great job being thorough on testing!

d-chambers (approved):
Looks good!

PR #105, opened 2023-01-10
Fk transform

SeunghooKim:
## Checklist

I have (if applicable):


d-chambers:
Hey @SeunghooKim,

Thanks for revisiting this. There are a few small points we probably need to get nailed down here before merging. I don't have a lot of experience with f-k analysis, so maybe @eileenrmartin, @aaronjgirard, or @jinwar  can weigh in. We may even want to spend some time on the design of this transform (and its inverse) in our next developer meeting.

1. Right now it looks like you are modifying coordinates for the frequency axis but not the other axis. They both need to change since we apply the dft to both axis. 

2.  We may want to perform an [fft_shift](https://numpy.org/doc/stable/reference/generated/numpy.fft.fftshift.html) so the zero frequency appears in the center of the coordinate array as most would expect.

3. It looks like you are following what I did with the rft transform in naming the new dimension "frequency_{dim}" which works for generic dimensions. This is probably the best, most general, approach but I would be curious if anyone else has ideas for naming the new dimensions.

There is an xarray package called [xrft](https://github.com/xgcm/xrft) which might be a useful reference.

d-chambers:
Hey @SeunghooKim, sorry for sitting on this soooooo long. 

It turns out the implementation was quite a bit more complicated than we anticipated when placing this item on the sprint board back in December. We had to account for proper scaling, frequency shifting, unit transformations, and making the transform reversible (which has complexities related to going from absolute time, in datetime, to frequency, then back to absolute time). 

Anyway, I finally found some time to tackle this in #192. I would love to get your feedback, both on the new notes section of the documentation and the method interfaces. We can always change the master branch before a release is made if you have suggestions on how to improve it.

Thanks again for your efforts though, and sorry for setting you up with such a complicated task on this one.

PR #102, opened 2022-12-23
add python 3.11 to workflow

d-chambers:
## Description

This PR adds python 3.11 to the CI matrix. 3.11 has been out for a few months now and I think all the scientific libraries DASCore depends on now support it.

## Checklist

I have (if applicable):

- [x] referenced the GitHub issue this PR closes.
- [x] documented the new feature with docstrings or appropriate doc page.
- [x] included a test. See [testing guidelines](https://dascore.org/contributing/testing.html).
- [x] your name has been added to the contributors page (docs/contributors.md).
- [x] added the "ready_for_review" tag once the PR is ready to be reviewed.

PR #109, opened 2023-01-19
Fix doc build

d-chambers:
## Description

This PR attempts to fix the github action which builds and deploys the documentation. The issue may be related to py3.11 (see #102).  

The failing action with debug output if found [here](https://github.com/DASDAE/dascore/actions/runs/3953963361/jobs/6771601791)

## Checklist

I have (if applicable):

- [ ] removed the debug code from the build deploy action (self reminder)


d-chambers:
I ended up having to replace the panflute filter with something a little more hacky, but probably faster and more robust to changes in pandocs json output. Also dropped panflute as a documentation dependency.

PR #133, opened 2023-05-15
Optasense io support

d-chambers:
## Description

Adds support for optasense V2 data format, which is used by the Quantx interrogator. 

## Checklist

I have (if applicable):


d-chambers:
Hey @JacksonZhu90,

Good work on this. I made a few tweaks. I removed some of the comments that only pertained to the Terra15 parser (no need to confuse our future selves) and I changed the distance array from channel number to cable distance. I will highlight this in the review so you can double check you agree with it. 

Also, we want to make sure the name is right because once we merge this we can't change it. We might get @eileenrmartin to help here. What is the correct name for this format? Is "Optasense V2" ok or is this format specific to the quantx interrogator so "Quantx V2" might be better?

d-chambers:
So I heard from @eileenrmartin today. It will be best to call this format QuantX v2 as it is different from the ODH3/ODH4 formats. Do you want to make that change @JacksonZhu90 or should I?

JacksonZhu90:
I am on my way to Houston. Could you make that change @d-chambers ?

d-chambers:
Sure thing, happy to wrap it up. I just didn't want to steal your thunder. Have a nice trip.

d-chambers:
Ok, there will probably be future bug fixes to do once we test this on additional QuantX files, but for now it looks good and I am confident we have the name right. Anything else we can address in future PRs.

Nice work @JacksonZhu90.

🥳

PR #146, opened 2023-06-01
Windows compatibility tweak: astype(int)

shawnboltz:
## Description

This PR makes some small compatibility tweaks for Windows by replacing any instances of `.astype(int)` with `.astype(np.int64)`. This is necessary because numpy will assume a smaller-sized int by default on Windows which leads to overflow errors (ex., when dealing with dates).

It should be noted that I just assumed that np.int64 was appropriate in all cases where I came across it, but because I'm not at all familiar with the codebase I don't know if it would be more appropriate to use something smaller (ex., for something memory intensive).

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/proc/aggregate.py
+++ b/dascore/proc/aggregate.py
@@ -50,7 +50,7 @@ def aggregate(
     # update coords with new dimension (use mean)
     coords = patch.coords
     if dim == "time":  # need to convert time to ints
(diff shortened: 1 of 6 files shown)


shawnboltz:
@d-chambers 
This is ready for your review.

PR #195, opened 2023-07-20
Io refactor

d-chambers:
## Description

This PR refactors the IO modules/machinery. In addition to basic code clean-up it does the following:

1. Creates a common test-suite for basic IO tests that all FiberIO instances should pass. This allows new IO modules to just plug into these tests so new contributors don't have to write many of their own tests.

2. FiberIO methods can now specify the type parameter they require. 

For example, before a read method might have looked like this:

```python
def read(path):
    with open(path, 'rb') as resource:
        ...
```

Can now do this:

```python
from dascore.io import BinaryReader

def read(resource: BinaryReader):
    ....
```

Of course the former still works, but with the latter we can now reuse open-file handles and the fiber methods are simplified.

3. Updates the documentation to reflect  these changes.

4. Remove redundant test code which is now handled by the common test code.

## Checklist

I have (if applicable):


d-chambers:
One unforeseen side-effect, some of the prototypes I used here don't work with python 3.9. Python 3.9's EoL isn't until 2025 but I think we probably still drop it because:

1. We don't yet have any packages that depend on DASCore (as listed in the github interface)
2. It is easy for users to create a python 3.10 or 3.11 virtual env
3. Since python 3.12 comes out in October, then we will support the 3 newest versions of python soon.

PR #219, opened 2023-08-12
Attrs refactor

### LINKED ISSUE #163: DASCore index to support missing time or distance values
## Description
Currently the indexing mechanism for the `DirectorySpool` requires all values in "time_max", "time_min", "d_time", "distance_max", "distance_min", "d_distance" to be non-null values. However, this doesn't need to be the case, and its conceivable some patches may not have even sampling in both time/distance, or may have a completely different second dimension name "eg channel_number". We need to make sure the indexing mechanism is flexible enough to handle these cases.

### LINKED ISSUE #194: Can't get gauge length from attributes - PRODML v. 2.0 format
## Description
I can see "GaugeLength" in the attributes list using HDFView software. However, I can't get the value using: ```patch_0.attrs['gauge_length']```
Please note that I can get some other attributes such as sampling interval or channel spacing.

Data format: ONYX - PRODML v. 2.0

## Example
```python
import dascore as dc
sp = dc.spool(data_path)
patch_0 = sp[0]
gauge_length = patch_0.attrs['gauge_length']
```

### Error:
```python
AttributeError                            Traceback (most recent call last)
Cell In[3], line 12
      8 print(patch_0.attrs)
     11 # get sampling rate, channel spacing, and gauge length
---> 12 gauge_length = patch_0.attrs['gauge_length']
     13 print("Gauge length = ", gauge_length)
     14 channel_spacing = patch_0.attrs['d_distance']

File [~/anaconda3/envs/py10/lib/python3.10/site-packages/dascore/core/schema.py:94](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/ahmadtourei/coding/fervo/iDAS_stimulation_stage8/LF_DAS/~/anaconda3/envs/py10/lib/python3.10/site-packages/dascore/core/schema.py:94), in PatchAttrs.__getitem__(self, item)
     93 def __getitem__(self, item):
---> 94     return getattr(self, item)

AttributeError: 'PatchAttrs' object has no attribute 'gauge_length'
```

## Versions
 - OS [e.g. Ubuntu 20.04]: Ubuntu 20.04
 - DasCore Version [e.g. 0.0.5]: 0.0.12
 - Python Version [e.g. 3.10]: 3.10

### LINKED ISSUE #207: support for datetime and timedelta
## Description
DASCore currently doesn't support python `datetime` or `timedelta` objects. It should.

## Example
```python

from datetime import datetime

import numpy as np

import dascore as dc

py_dt = datetime.now()
py_td = datetime.now() - py_dt

# these both raise an error
dc.to_datetime64(py_dt)
dc.to_timedelta64(py_td)

# they should also work with lists
dc.to_datetime64([py_dt] * 10)
dc.to_timedelta64([py_td] * 10)

# and arrays
dc.to_datetime64(np.array([py_dt] * 10))
dc.to_timedelta64(np.array([py_td] * 10))
```

### LINKED ISSUE #209: Better unit display
## Description
Currently the unit display is a bit overly verbose. For example, "meter" rather than "m" is displayed in plots and such. [This SO post](https://stackoverflow.com/questions/65681490/format-pint-unit-as-short-form-symbol) shows how to configure pint to be more concise.

### LINKED ISSUE #213: Time axis label in waterfall plots
## Description
The time axis label in the waterfall plot should just read "time" not "time(s)". This should also work for other dimensions not named "time" but are datetime64 types.

### LINKED ISSUE #221: Could not index PRODML v. 2.0 format
## Description
The index file is not created in the data directory. So, I could not get the patches out of the spool.
Data format: PRODML v. 2.0 format

Please note that no error occurred after getting the spool and indexing got to 100%:
```python
import dascore as dc

sp = dc.spool(data_path)
```

## Versions
 - OS [e.g. Ubuntu 20.04]: Ubuntu 20.04
 - DasCore Version [e.g. 0.0.5]: 0.0.13.dev25+g8ebe6e6
 - Python Version [e.g. 3.10]: 3.10

d-chambers:
## Description
This PR refactors how attributes work in DASCore and fixes a few other small issues. 

Admittedly, this PR got a bit too large as I was working without internet on a roadtrip which made it harder to do a series of smaller PRs.

Here are the highlights:

1. Rather than containing {dim}_min, {dim}_max, d_{dim} there is now a coordinate dictionary attached to the attributes. This allows for arbitrarily named dimensions to be tracked rather than just time and distance.

2. In light of 1, attrs.{dim}_min, attrs.{dim}_max, attrs.d_{dim} are still accessible, but they are just properties which fetch the correct attributes from the coordinates dictionary. attrs.d_{dim} is deprecated in favor of attrs.{dim}_step which is more clear and fits better when attribute names are sorted.

3. `PatchSummary` is no longer a public facing class, but `PatchAttrs` can just be to summarize a patch.

4. Now only time is used to index a patch by `DirectorySpool` and other dimensional queries are fully supported. 

5. DASCore's `DirectorySpool` now keeps track of where index files for each directory, so if a non-default index is specified it will be used with each new DirectroySpool without having to specify it. Also, if the user can't write to the data directory, a new, suitable location is chosen automatically.  

6. `Patch.__init__` and `Patch.new` now behave in a more structured way. 

6. Fixes #213 
7. Fixes #209
8. Fixes #207 
9. Fixes #194 
10. Fixes #163 
11. Fixes #221 

## Checklist

I have (if applicable):


d-chambers:
Merging for now, I am feeling much better about DASCore's internals. Hopefully no more *huge* PRs like this one.

Issue #220, opened 2023-08-16
patch ordering in a spool


ahmadtourei:
## Description
The patches seem to not be in order.

## Example
```python
import dascore as dc

sp = dc.spool(data_path)

# print the contents of first 5 patches
content_df = sp.get_contents()
content_df.head()
```

## Expected behavior
The first 5 paths were:
```UTC_20230322_030824.631.h5```
.
.
.
```UTC_20230322_031224.631.h5```

Instead of:
```UTC_20230322_030024.631.h5```
.
.
.
```UTC_20230322_030424.631.h5```

I also printed the ```content_df```:
```python
    cable_id  d_distance                 d_time data_category    data_type  \
0               2.041904 0 days 00:00:00.001000                strain_rate   
1               2.041904 0 days 00:00:00.001000                strain_rate   
2               2.041904 0 days 00:00:00.001000                strain_rate   
3               2.041904 0 days 00:00:00.001000                strain_rate   
4               2.041904 0 days 00:00:00.001000                strain_rate   
..       ...         ...                    ...           ...          ...   
235             2.041904 0 days 00:00:00.001000                strain_rate   
236             2.041904 0 days 00:00:00.001000                strain_rate   
237             2.041904 0 days 00:00:00.001000                strain_rate   
238             2.041904 0 days 00:00:00.001000                strain_rate   
239             2.041904 0 days 00:00:00.001000                strain_rate   

              dims  distance_max  distance_min file_format file_version  \
0    time,distance   2608.323975   -264.636139      PRODML          2.1   
1    time,distance   2608.323975   -264.636139      PRODML          2.1   
2    time,distance   2608.323975   -264.636139      PRODML          2.1   
3    time,distance   2608.323975   -264.636139      PRODML          2.1   
4    time,distance   2608.323975   -264.636139      PRODML          2.1   
..             ...           ...           ...         ...          ...   
235  time,distance   2608.323975   -264.636139      PRODML          2.1   
236  time,distance   2608.323975   -264.636139      PRODML          2.1   
237  time,distance   2608.323975   -264.636139      PRODML          2.1   
238  time,distance   2608.323975   -264.636139      PRODML          2.1   
239  time,distance   2608.323975   -264.636139      PRODML          2.1   

    instrument_id network                                 path station tag  \
0                          /UTC_20230322_030824.631.h5               
1                          /UTC_20230322_030924.631.h5               
2                          /UTC_20230322_031024.631.h5               
[...]

PR #222, opened 2023-08-17
Sort method for spool

ahmadtourei:
## Description
This PR implements the spool.sort method into DASCore. 
Related issue: #220

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/core/spool.py
+++ b/dascore/core/spool.py
@@ -330,41 +330,33 @@ class DataFrameSpool(BaseSpool):
         """
         {doc}
         """
-        try:
-            df = self._df
-            inst_df = self._instruction_df
+        df = self._df
+        inst_df = self._instruction_df
 
-            # make sure a suitable attribute is entered
-            attrs = set(df.columns)
-            if attribute not in attrs:
-                # make sure we can also cover coordinate names instead of the attribute
-                if f"{attribute}_min" in attrs:
-                    attribute = f"{attribute}_min"
-                else:
-                    msg = (
-                        "Invalid attribute. "
-                        "Please use a valid attribute such as: 'time'"
-                    )
-                    raise IndexError(msg)
+        # make sure a suitable attribute is entered
+        attrs = set(df.columns)
+        if attribute not in attrs:
+            # make sure we can also cover coordinate names instead of the attribute
+            if f"{attribute}_min" in attrs:
+                attribute = f"{attribute}_min"
+            else:
+                msg = (
+                    "Invalid attribute. " "Please use a valid attribute such as: 'time'"
+                )
+                raise IndexError(msg)
 
-            # get a mapping from the old current index to the sorted ones
-            sorted_df = df.sort_values(attribute).reset_index(drop=True)
-            old_indices = df.index
-            new_indices = np.arange(len(df))
-            mapper = pd.Series(new_indices, index=old_indices)
+        # get a mapping from the old current index to the sorted ones
+        sorted_df = df.sort_values(attribute).reset_index(drop=True)
+        old_indices = df.index
+        new_indices = np.arange(len(df))
+        mapper = pd.Series(new_indices, index=old_indices)
 
-            # swap out all the old values with new ones
-            new_current_index = inst_df["current_index"].map(mapper)
-            new_instruction_df = inst_df.assign(current_index=new_current_index)
+        # swap out all the old values with new ones
+        new_current_index = inst_df["current_index"].map(mapper)
+        new_instruction_df = inst_df.assign(current_index=new_current_index)
 
-            # create new spool from new dataframes
-            return self.new_from_df(df=sorted_df, instruction_df=new_instruction_df)
-        except IndexError:  # Catch only IndexError
-            msg = "Invalid attribute. Please use a valid attribute such as: 'time'"
-            raise IndexError(msg)
-        except Exception:
-            msg = f"spool of type {self.__class__} has no sort implementation"
-            raise NotImplementedError(msg)
+        # create new spool from new dataframes
+        return self.new_from_df(df=sorted_df, instruction_df=new_instruction_df)
 
     @compose_docstring(doc=BaseSpool.get_contents.__doc__)
     def get_contents(self) -> pd.DataFrame:
--- a/dascore/core/spool.py
+++ b/dascore/core/spool.py
@@ -99,7 +99,16 @@ class BaseSpool(abc.ABC):
     def sort(self, attribute) -> Self:
         """
         Sort the Spool based on a specific attribute.
+
+        Parameters
+        ---------------
+        atribute
+            The attribute or coordinate used for sorting. If a coordinate name
+            is used, the sorting will be based on the minimum value.
         """
+        raise NotImplementedError(
+            f"spool of type {self.__class__} has no sort implementation"
+        )
 
     @abc.abstractmethod
     def __len__(self) -> int:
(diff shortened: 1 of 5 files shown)


dascore/core/spool.py
```diff
@@ -94,6 +95,12 @@ def get_contents(self) -> pd.DataFrame:
         Get a dataframe of the patches that will be returned by the spool.
         """
 
+    @abc.abstractmethod
```
d-chambers, line 98: We need to think about whether we want to require all spools to implement sort. I am thinking about a spool that interfaces with a remote server (imagine if you are trying to interact with all of EarthScope's DAS data for example); maybe it doesn't make sense to require it to to know how to sorts its own massive contents. 

For now, rather than making this an abstractmethod, which will require all subclasses to implement `sort`, let's change it to just raise a `NoteImplementedError`: 

```python
def sort(self, attribute) -> Self:
    """
    Sort the Spool based on a specific attribute.
    """
    raise NoteImplementedError(f"spool of type {self.__class__} has no sort implementation")
```

Then if the subclass implements sort it will work fine, otherwise this Error will get raised.

dascore/core/spool.py
```diff
@@ -309,6 +316,37 @@ def select(self, **kwargs) -> Self:
         )
         return out
 
+    @compose_docstring(doc=BaseSpool.sort.__doc__)
+    def sort(self, attribute) -> Self:
+        """Sort the Spool based on a specific attribute."""
```
d-chambers, line 321: The `compose_docstring` decorator allows you to reuse strings in the documentation. You don't appear to be using it here though. So perhaps you want this?

```python
@compose_docstring(doc=BaseSpool.sort.__doc__)
def sort(self, attribute) -> Self:
    """
    {doc}
    """
```

Also, We should add more to the docs explaining the parameters and such. See note above.

dascore/core/spool.py
```diff
@@ -309,6 +316,37 @@ def select(self, **kwargs) -> Self:
...
+        df = self._df
+        inst_df = self._instruction_df
+
+        # make sure we can also cover "time" and "distance" as attributes
+        if attribute == "time":
+            attribute = "time_min"
+        elif attribute == "distance":
+            attribute = "distance_min"
+
```
d-chambers, line 330: I would like to get away from "time" and "distance" being special so we can support transformed patches that might have differently named dimensions. Let's remove this and add more general logic.

dascore/core/spool.py
```diff
@@ -309,6 +316,37 @@ def select(self, **kwargs) -> Self:
...
+        # make sure a suitable attribute is entered
+        attrs = set(df.columns)
+        if attribute not in attrs:
+            msg = "Invalid attribute. Please use a valid attribute such as: 'time'"
+            raise IndexError(msg)
```
d-chambers, line 335: So here something like this:

```python 
if attribute not in attrs:
    if f"{attribute}_min" in attrs:
        attribute = f"{attribute}_min"
    else:
        msg = "Invalid attribute. Please use a valid attribute such as: 'time'"
        raise IndexError(msg)
```

d-chambers (changes requested):
Great work @ahmadtourei! Just a couple of things to tweak and it should be good to go.

tests/test_core/test_spool.py
```diff
@@ -224,6 +225,53 @@ def test_multiple_range_selects(self, adjacent_spool_no_overlap):
             assert patch.attrs["distance_max"] <= distance_max
 
 
+class TestSort:
+    """
+    Tests for sorting spools.
+    """
+
+    # A dataframe which represents contents as they will be output
+    _df: pd.DataFrame = CacheDescriptor("_cache", "_get_df")
```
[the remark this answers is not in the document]
ahmadtourei: To convert the spool to a DataFrame (```df = sorted_spool._df```) and then test:
```python
assert df["time_min"].is_monotonic_increasing
```

tests/test_core/test_spool.py
```diff
@@ -224,6 +225,53 @@ def test_multiple_range_selects(self, adjacent_spool_no_overlap):
...
+        """
+        try:
+            diverse_spool.sort("dummy_attribute")
+        except IndexError as e:
+            assert (
+                str(e)
+                == "Invalid attribute. Please use a valid attribute such as: 'time'"
+            )
+        else:
+            assert False, "Expected an IndexError but no exception was raised."
```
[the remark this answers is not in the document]
ahmadtourei: Nice!

dascore/core/spool.py
ahmadtourei: Agree.

dascore/core/spool.py
ahmadtourei: Agree.

dascore/core/spool.py
```diff
@@ -94,6 +95,12 @@ def get_contents(self) -> pd.DataFrame:
         Get a dataframe of the patches that will be returned by the spool.
         """
 
+    @abc.abstractmethod
+    def sort(self, attribute) -> Self:
+        """
+        Sort the Spool based on a specific attribute.
+        """
```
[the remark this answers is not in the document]
ahmadtourei: Yes, I just realized that I forgot to add this yesterday!

dascore/core/spool.py
ahmadtourei: Good idea!

ahmadtourei: I included this raise using a try/except block. However, I'm not sure how I can test it. For testing, I need a spool whose type has no sort implementation. Do you have any suggestions?

tests/test_core/test_spool.py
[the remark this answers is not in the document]
ahmadtourei: Got it! Thanks for the explanation!
```df = sorted_spool.get_contents()``` is what I missed.

File: dascore/core/spool.py (hunk 1 above)
Comment (ahmadtourei) [reply dascore/core/spool.py:98]: I think just a dummy spool for testing will work.

Comment (d-chambers) [reply dascore/core/spool.py:98]: If you just want to make sure it raises for the sake of testing you can call it directly and pass an instance of a subclass spool to it. 

```python 
def test_base_spool_sort_raises(self, random_spool):
    with pytest.raises(...):
        BaseSpool.sort(random_spool, 'time')
```

ahmadtourei: The test failed. Here is the test:
```python
    def test_base_spool_sort_raises(self, random_spool):
        expected_str = "spool of type"
        with pytest.raises(NotImplementedError, match=expected_str):
            BaseSpool.sort(random_spool, 'time')
```

Error:
```python
    def test_base_spool_sort_raises(self, random_spool):
        expected_str = "spool of type"
>       with pytest.raises(NotImplementedError, match=expected_str):
E       Failed: DID NOT RAISE <class 'NotImplementedError'>

tests/test_core/test_spool.py:243: Failed
```

ahmadtourei: I wanted to add a test like the one below:

```python
    def test_sorting_not_implemented(self):
        """
        Test sorting a not supported spool type.
        """
        dummy_spool = ...
        expected_str = "spool of type"
        with pytest.raises(NotImplementedError, match=expected_str):
            dummy_spool.sort("time")
```

d-chambers: > The test failed

Weird, what does `BaseSpool.sort` look like? Can you paste it here?

ahmadtourei: @abc.abstractmethod
    def sort(self, attribute) -> Self:
        """
        Sort the Spool based on a specific attribute.

        Parameters
        ---------------
        atribute
            The attribute or coordinate used for sorting. If a coordinate name is used, the sorting will be based on the minimum value.
        """

ahmadtourei: And here is the DataFrameSpool.sort:

```python
   @compose_docstring(doc=BaseSpool.sort.__doc__)
    def sort(self, attribute) -> Self:
        """
        {doc}
        """
        try:
            df = self._df
            inst_df = self._instruction_df

            # make sure a suitable attribute is entered
            attrs = set(df.columns)
            if attribute not in attrs:
                # make sure we can also cover coordinate names instead of the attribute
                if f"{attribute}_min" in attrs:
                    attribute = f"{attribute}_min"
                else:
                    msg = "Invalid attribute. Please use a valid attribute such as: 'time'"
                    raise IndexError(msg)

            # get a mapping from the old current index to the sorted ones
            sorted_df = df.sort_values(attribute).reset_index(drop=True)
            old_indices = df.index
            new_indices = np.arange(len(df))
            mapper = pd.Series(new_indices, index=old_indices)

            # swap out all the old values with new ones
            new_current_index = inst_df["current_index"].map(mapper)
            new_instruction_df = inst_df.assign(current_index=new_current_index)

            # create new spool from new dataframes
            return self.new_from_df(df=sorted_df, instruction_df=new_instruction_df)
        except IndexError:  # Catch only IndexError
            msg = "Invalid attribute. Please use a valid attribute such as: 'time'"
            raise IndexError(msg)
        except Exception:
            msg = f"spool of type {self.__class__} has no sort implementation"
            raise NotImplementedError(msg)
```

d-chambers: Ok, so did you forget this line 
```python
raise NoteImplementedError(f"spool of type {self.__class__} has no sort implementation")
```
in `BaseSpool.sort`? That would explain why it isn't raising an exception in this code:

```python
def test_base_spool_sort_raises(self, random_spool):
    expected_str = "spool of type"
    with pytest.raises(NotImplementedError, match=expected_str):
        BaseSpool.sort(random_spool, 'time')
```

ahmadtourei: > Ok, so did you forget this line

Yes. I had not added this line. It works now. Thanks very much for walking me through this!

d-chambers (approved):

d-chambers:
Good to go. Feel free to press the green squash and merge button 🎉

PR #231, opened 2023-08-26
read pubdas TDMS

d-chambers:
## Description
 So it appears the TDMS version, or at least the attributes, are different from our TDMS test file and the ones in pubdas. This PR fixes the issue by changing how distances are determined.

@aissah, do you mind reviewing this? 

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/io/tdms/utils.py
+++ b/dascore/io/tdms/utils.py
@@ -104,7 +104,6 @@ def _get_version_str(tdms_file, LEAD_IN_LENGTH=28) -> str:
     # lead_in is 28 bytes:
     # [string of length 4][int32][int32][int64][int64]
     fields = struct.unpack("<4siiQQ", lead_in)
-
     # TODO: validate file
     if fields[0].decode() in "TDSm":
         version_str = str(fields[2])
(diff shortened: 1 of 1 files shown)


aissah:
I am a bit confused about what the issue is. We had one test file that you got from Silixa, I think and another that I made from the FORESEE dataset. Which of them is mismatched with that on PUBDAS?
Also, the FORESEE dataset was reformatted and uploaded as hdf5. Which other dataset on PUBDAS uses TDMS, maybe Silixa somehow allows customization of attributes for different datasets?
Lastly, is it that some of the attributes are missing or they are saved under different "keys"?

d-chambers:
> Which of them is mismatched with that on PUBDAS?

Maybe both? The problem is that the Fossa TDMS files don't have "Start Distance (m)" and "Stop Distance (m)" in their attrs and I didn't see anything similar. But all the TDMS files I have seen so far do have "StartPosition[m]" and "n_channels" which prompted the change [here](https://github.com/DASDAE/dascore/pull/231/files#).

I am not sure if you can add custom attrs to the TDMS file or if we are just seeing a slightly different version of TDMS data.

 I just wanted your feedback to see if you know of any issues that might crop up by doing it this way.

aissah:
That seems like a reasonable fix to me. We'll just have to wait and see 😅.

d-chambers:
> We'll just have to wait and see 

Ya, or actually read the TDMS spec 😝. Will merge for now since the change still allows us to read our test files **and** the TDMS files in original PubDAS datasets.

PR #299, opened 2023-11-16
embarrassingly mpi4py recipe

ahmadtourei:
## Description

This PR provides a recipe and example codes for using the mpi4py library along with DASCore for parallelization purposes. 

## Checklist

I have (if applicable):


d-chambers:
A few suggestions:

1. We probably don't want to require mpi4py to build the docs because of this recipe, so we can tell quarto to just print the code and not run it by adding this to the start of the file:
```
---
title: "Parallelization using Open MPI"
execute:
  eval: false
---
```

2. It would be good to include instructions on how to run the script, and modify the codeblock so the filename is shown. You can do this by including a `filename` in the code fence like this:

````
```{.python filename="mpi_spool.py"}
... 
```
`````

Then show how to run it:

```bash
mpiexec -n 4 python mpi_spool.py
```


3. We talked about `spool.split` , and I still think that approach might be a little more clear but I will leave it up to you to decide. This is what the split code looks like:

```{.python filename="mpi_spool.py"}
import sys

import dascore as dc

from mpi4py import MPI

# Initiate MPI
comm = MPI.COMM_WORLD
rank = comm.Get_rank()
size = comm.Get_size()

# Load and split the spool, one patch per process
spool_list = list(
    dc.get_example_spool("random_das")
    .split(count=size)  # Evenly (as possible) split spool into count parts
)

# If more processes than patches exit excess processes
if rank >= len(spool_list):
    print(f"Exiting excess process {rank}")
    sys.exit(0)

# Get the spool that belongs to this process
sp = spool_list[rank]

for patch in sp:
    print(f"rank: {rank}, patch: {patch}")
    ...
```

d-chambers:
Even more pythonic version:

```python
import sys

import dascore as dc

from mpi4py import MPI

# Initiate MPI
comm = MPI.COMM_WORLD
rank = comm.Get_rank()
size = comm.Get_size()

# Load and split the spool, one patch per process
spool_list = list(
    dc.get_example_spool("random_das")
    .split(count=size)  # Evenly (as possible) split spool into count parts
)

try:
    sp = spool_list[rank]   # Get the spool that belongs to this process
except IndexError:
    print(f"Exiting excess process {rank}")
    sys.exit(0)

for patch in sp:
    print(f"rank: {rank}, patch: {patch}")
    ...
```

ahmadtourei:
Hey @code-cullison, Please have a look at this PR and let me know if you have any comments. 
Regarding DASK and Multiprocessing, feel free to add instructions and example codes here or on a separate recipe so we can also work on them at some point.

d-chambers (approved):

d-chambers:
Hey @ahmadtourei  or @code-cullison, I just added a bit on using `concurrent.futures` with `Spool.map` to give users a few options, and made some of the headings more concise. Take a look, and when you are happy with it smash the merge button.
🥳

code-cullison:
Derrick, per our conversation last week, I tested if individual ranks can open and write to separate files -- indeed they can. I think I was was convoluting stdin/stdout and parallel file access. For stdin/stdout, I've had issues when I've tried having ranks other than 'root' use these pipes. For parallel access to the SAME file, then MPI-IO needs to be used.

Note, if the .split() results in memory copies, then each rank will make the same copy(ies).  If the split only changes the "overseeing" data structure (e.g. returns a different iterable) then I think the split idea is good. Otherwise, for large files or large spools, the memory copying by individual ranks could negatively impact performance.  I'm assuming code was also tested using MPI?

PR #300, opened 2023-11-21
added to_pandas and test

code-cullison:
## Description

Added functionality for converting Patches to pandas.DataFrame

## Checklist

I have (if applicable):


code-cullison (commented):
I added df_to_patch and the related test. I'm guessing that there will need to be a discussion regarding attrs.  

One thought is that it be required that attrs be a PatchAttr.  In contrast, if we let it be a dictionary, then the creation of the coords and dims could be a problem.

Also, if we attrs=None, then how does when build the coords and dims, or should it be required that the df already be in some dascore acceptable form?  This will affect the testing function

Another possibility is to add a utility function to create PatchAttr beforehand?

d-chambers (changes requested):
Looks good so far, I left a few suggestions. We probably need a few more short test cases for `patch_to_df`. The ones I can think of are
- Test attrs as dictionary
- Test attrs as a PatchAttrs object
- Test either case when attrs doesn't have dims but they are on the dataframe
- Test when the dataframe doesnt have dims but they are on the patch
- Test that an error is raised when no dims are found. `pytest.raises` will be useful for this.
Regarding the attrs/dims interaction, here is what I am thinking.

If dims are in the attrs dict those should be given priority, otherwise we use the ones extracted from the names. Something like this?:

```python
dims = [df.index.name, df.columns.name]
coords = {dims[0]: df.index.to_numpy(), dims[1]: df.columns.to_numpy()}
data = df.to_numpy()

if attrs is not None:
    # allow for patch attributes or a dictionary, but defer to PatchAttrs parsing logic
    attrs = atts if isnstance(attrs, PatchAtts) else PatchAttrs(**attrs)
    if not attrs.dims:  # no dims found in attrs, apply the ones from the dataframes
        attrs = attrs.update(dims=dims)
    if not patch.dims:
        ....
        # raise an exception here since we can't get the dim names
    dims = attrs.dim_tuple  # now we are sure attrs has correct dims for patch
 
return Patch(data=data, coords=coords, dims=dims, attrs=attrs)
```

d-chambers:
By the way, congrats on getting number 300 for this PR. I feel like you deserve some kind of door prize. If we get DASDAE/DASCore stickers printed I will save the first one for you 😄

d-chambers:
Hey @code-cullison, I just polished this a bit so we can merge it. Thanks for working on it.

PR #301, opened 2023-11-21
Added RSF writer and tests

aaronjgirard:
## Description

Added a method to write out RSF files and included tests for the code. 

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/io/rsf/__init__.py
+++ b/dascore/io/rsf/__init__.py
@@ -3,18 +3,17 @@ RSF format support module.
 
 Notes
 -----
--
+- output has been tested on Madagascar version 3.1-git
 
 
 Examples
 --------
 import dascore as dc
-from dascore.utils.downloader import fetch
 
-# get the path to a segy file.
-path = fetch("conoco_segy_1.sgy")
+# get the path to a random DAS file.
+patch = dc.get_example_patch()
 
-segy_patch = dc.spool(path)[0]
+patch.io.write("test_out.rsf","rsf",data_path="test_out.rsf")
 """
 
 from .core import RSFV1
--- a/dascore/io/rsf/__init__.py
+++ b/dascore/io/rsf/__init__.py
@@ -1,0 +1,20 @@
+"""
+RSF format support module.
+
+Notes
+-----
+-
+
+
+Examples
+--------
+import dascore as dc
+from dascore.utils.downloader import fetch
+
+# get the path to a segy file.
+path = fetch("conoco_segy_1.sgy")
+
+segy_patch = dc.spool(path)[0]
+"""
+
+from .core import RSFV1
--- a/dascore/io/rsf/core.py
+++ b/dascore/io/rsf/core.py
@@ -60,14 +60,20 @@ class RSFV1(FiberIO):
         axis_names = patch.dims
         axis_units = [patch.get_coord(x).units.units for x in patch.dims]
 
-        ## we are casting to float32 to ensure that esize=4 for m8r
-        data = patch.data.astype(np.float32)
+        data = patch.data
         dtype = np.dtype(data.dtype)
-        file_esize = dtype.itemsize
 
         if np.issubdtype(dtype, np.integer):
+            ## we are casting to float32 to ensure that esize=4 for m8r
+            data = patch.data.astype(np.float32)
+            dtype = np.dtype(data.dtype)
+            file_esize = dtype.itemsize
             file_formt = 'data_format="native_float"'
         elif np.issubdtype(dtype, np.floating):
+            ## we are casting to float32 to ensure that esize=4 for m8r
+            data = patch.data.astype(np.float32)
+            dtype = np.dtype(data.dtype)
+            file_esize = dtype.itemsize
             file_formt = 'data_format="native_float"'
         else:
             raise ValueError("Data format is not integer or floating.")
--- a/dascore/io/rsf/core.py
+++ b/dascore/io/rsf/core.py
@@ -9,6 +9,7 @@ import numpy as np
 
 import dascore as dc
 from dascore.io.core import FiberIO
+from dascore.utils.time import to_float
 
 RSFKEYS_WRITE = ("in", "esize", "data_format")
 
@@ -28,7 +29,7 @@ class RSFV1(FiberIO):
     #   like hide it in header file when writing out an RSF file.
     ########################
 
-    def write(self, patch, path, data_path=None, **kwargs):
+    def write(self, spool, path, data_path=None, **kwargs):
         """
         Write a patch to RSF format.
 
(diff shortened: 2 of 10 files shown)


dascore/io/rsf/core.py
```diff
@@ -0,0 +1,100 @@
...
+        data = patch.data.astype(np.float32)
+        dtype = np.dtype(data.dtype)
+        file_esize = dtype.itemsize
+
+        if np.issubdtype(dtype, np.integer):
+            file_formt = 'data_format="native_int"'
+        elif np.issubdtype(dtype, np.floating):
+            file_formt = 'data_format="native_float"'
+        else:
+            raise ValueError("Data format is not integer or floating.")
```
d-chambers, line 63: Since we now explicitly convert to float32 dtype I think these lines can just be `file_format = "native_float"` right?

dascore/io/rsf/core.py
```diff
@@ -0,0 +1,100 @@
...
+        Write a patch to RSF format.
+
+        if no data_path is NOT specified, the header and binary will be
+        packed together
+        data_path needs to be bindata_file.rsf or /location/of/bindata_file.rsf
+        (NO '@')
+
+        path needs to be hdr_file.rsf or /location/of/hdr_file.rsf
+
+        spool needs to have a single patch in it
```
d-chambers, line 43: So this will generate a nice API doc page, we need a parameter section:

```python

Parameters
----------------
spool
    The input spool to convert to rsf, must have exactly one patch. 
path
    Path to create the rsf file
data_path
    If data and rsf header information are to be separate, the location of the data file. Needs to be 
    bindata_file.rsf or /location/of/bindata_file.rsf (no '@')

Notes
--------
- Patch datatype is converted to float32 for compatibility with Madagascar (may be able to keep dytpe in the future)
```

dascore/io/rsf/__init__.py
```diff
@@ -0,0 +1,20 @@
+"""
+RSF format support module.
+
+Notes
+-----
+-
```
d-chambers, line 6: delete empty section, or add some notes here if you have any. Maybe the version of Madagascar you used to test the output?

dascore/io/rsf/core.py
```diff
@@ -0,0 +1,100 @@
...
+    preferred_extensions = "rsf"
+    # also specify a version so when version 2 is released you can
+    # just make another class in the same module named rsfV2.
+    version = "1"
+
+    ########################
+    # Add a way to read in rsf files.
+    # May need a way to get absolute time stamp,
+    #   like hide it in header file when writing out an RSF file.
+    ########################
```
d-chambers, line 30: Is this resolved?

tests/test_io/test_rsf/test_rsf.py
```diff
@@ -0,0 +1,42 @@
...
+
+import dascore as dc
+from dascore.io.rsf import RSFV1
+
+
+class TestRsfWrite:
+    """testing the rSF write out function."""
+
+    def test_write_nopath(self, random_patch, tmp_path):
+        """Test write function."""
```
d-chambers, line 16: maybe:

"""test write function with no data path specified"""

tests/test_io/test_rsf/test_rsf.py
```diff
@@ -0,0 +1,42 @@
...
+
+        assert path.exists()
+        test_data = random_patch.data.astype(np.float32)
+        dtype = np.dtype(test_data.dtype)
+        file_esize = dtype.itemsize
+        datasize = test_data.size * file_esize
+        assert os.path.getsize(path) >= datasize
+
+    def test_write_path(self, random_patch, tmp_path):
+        """Test write function."""
```
d-chambers, line 29: "Test write function with separate data path"

d-chambers (changes requested):
A couple small things then should be good to merge when you are happy with it.

aaronjgirard:
Thanks, those are good changes. I have gone ahead and made them. We still
do not have test coverage for a file that is not int or float input
(frequency, for example), so one line of code is not covered in the tests.
Do you have an example test dataset with complex numbers, for example?


*Aaron Girard*

dascore/io/rsf/core.py
aaronjgirard: yes and removed

d-chambers:
> Do you have an example test dataset with complex numbers, for example?

We can use `patch.dft` to generate one like this: `complex_patch = random_patch.dft("time")`

But I think we are not supporting that now? [This line](https://github.com/DASDAE/dascore/blob/712b6a5b845fda8abc21ff3c8f81e011f5b29414/dascore/io/rsf/core.py#L64) explicitly converts any data type to float 32. If this is what you want to do you can essentially remove lines 68-73 and just have `file_formt = 'data_format="native_float"'`

aaronjgirard:
> But I think we are not supporting that now? [This line](https://github.com/DASDAE/dascore/blob/712b6a5b845fda8abc21ff3c8f81e011f5b29414/dascore/io/rsf/core.py#L64) explicitly converts any data type to float 32. If this is what you want to do you can essentially remove lines 68-73 and just have `file_formt = 'data_format="native_float"'` which will remove the untested lines and make codecov happy.

The only problem then is in line 72-73, if someone ever tries to export a complex valued data set it will surely export but I have no idea what a complex value DASCORE data set converted to float32 will do. 

Unless you know that that will be ok, I figure just give a test data set in not int and not float so that the tests cover the line ```raise ValueError("Data format is not integer or floating.")```

d-chambers:
> The only problem then is in line 72-73, if someone ever tries to export a complex valued data set it cannot export because patch = spool[0] in [line 56](https://github.com/DASDAE/dascore/blob/712b6a5b845fda8abc21ff3c8f81e011f5b29414/dascore/io/rsf/core.py#L56) cannot export and indeed fails when the following is added in the test:

This is a bug. I will open another issue. 

The dtype conversion will probably fail if it is complex, but should work fine with ints (it will just cast the ints to floats, which is probably fine).

d-chambers:
see #303.

Once we get that fixed we can either explicitly support the complex type or just raise an error if the array dtype is complex. You can check it with `np.issubdtype(dtype, np.complexfloating)`

aaronjgirard:
it looks like #304 fixed the complex values to spool issue, so I changed the RSF write and the test. However, now there another issue for writing integers to spool.

```
import dascore as dc

random_patch = dc.get_example_patch()
int_patch = random_patch.data.astype(np.int32)
spool = dc.spool(int_patch) # this raises error: "msg = 'Could not get spool from: {obj}" 
```

d-chambers:
I think that is actually the expected behavior. The problem is `int_patch = random_patch.data.astype(np.int32)` returns a numpy array (not a patch) with dtype int32, and dascore doesnt know how to make a spool from that.

Try this:

```python
import numpy as np

import dascore as dc

int_patch = random_patch.new(data=np.ones_like(random_patch.data)) 
```

d-chambers (approved):

PR #341, opened 2024-01-10
real-time proc codes and recipe

ahmadtourei:
## Description

This PR provides codes and a demonstration of real-time processing.

## Checklist

I have (if applicable):


docs/recipes/real_time_proc.qmd
```diff
@@ -0,0 +1,107 @@
...
+# Define path for saving results
+output_data_folder = '/path/to/desired/output/directory'
+
+# Get a spool to work on
+sp = dc.get_example_spool().sort("time").update()
+```
+
+## Set real-time processing parameters (if needed)
+
+In this section, we define the window size and step size required for [rolling](https://dascore.org/api/dascore/proc/rolling/rolling.html) mean processing. With a sampling interval of 10 seconds, the cutoff frequency (Nyquist frequency) is determined to be 0.05 Hz. Additionally, we establish the desired wait time after each run by using the `sleep_time_mult` parameter, which acts as a multiplier coefficient for the number of seconds in each patch.
```
d-chambers, line 33: The link should use DASCore's internal linking:

```
[rolling](`dascore.proc.rolling.rolling`)
```

That way each doc build will link to its own version of the rolling page rather than pointing to this static URL.

docs/recipes/real_time_proc.qmd
```diff
@@ -0,0 +1,107 @@
+---
+title: "Real-Time Processing"
+execute:
+  eval: false
+---
+
+
+This recipe serves as an example to showcase the real-time processing capability of DASCore. Here, we demonstrate how to use DASCore to perform rolling mean processing on a spool in real time for edge computing purposes.
```
d-chambers, line 8: I suggest using the term "near real-time batch processing". True real-time usually implies some kind of high frequency streaming.

docs/recipes/real_time_proc.qmd
```diff
@@ -0,0 +1,107 @@
...
+
+```{python}
+# Start the for loop for real-time processing
+i = 0
+while True:
+    initial_run = (i == 0)
+    run_num = i+1
+    print(f"\nRun number: {run_num}")
+
+    # Select a updated sub-spool
```
d-chambers, line 61: I think you need "Select an updated sub-spool"

docs/recipes/real_time_proc.qmd
```diff
@@ -0,0 +1,107 @@
...
+    sp = dc.spool(data_path).sort("time").update()
```
d-chambers, line 62: I am not 100% sure, but  I don't think update will preserve the sorting, so I suggest you change to:

```python
sp = dc.spool(data_path).update().sort("time")
```

Also, any reason not to define the spool outside of the while loop so we don't have to init it each iteration?

docs/recipes/real_time_proc.qmd
```diff
@@ -0,0 +1,107 @@
...
+    len_updated_sp = len(sp)
+
+    # Get number of seconds in the first patch 
+    sampling_interval = sp[0].attrs['d_time']
```
d-chambers, line 66: We are trying to move away from using attrs to get coord info. I suggest:

```python 
sp[0].coords.step("time")
```

d-chambers (approved):
A few suggestions on improve conciseness but looks good otherwise.

d-chambers:
Also, probably goes without saying, but before merging, it would be good to manually run the code to make sure it works 😄. With the exec: false setting the doc build won't actually verify that for us.

docs/recipes/real_time_proc.qmd
ahmadtourei: Oops I forgot to change this from my old codes.

docs/recipes/real_time_proc.qmd
```diff
@@ -0,0 +1,107 @@
...
+    num_sec = len(sp[0].coords["time"]) * sampling_interval
```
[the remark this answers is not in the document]
ahmadtourei: So, actually:  
```python
num_sec = sp[0].coords.max("time") - sp[0].coords.min("time") + sampling_interval
```
since `sp[0].coords.min("time")` is equal to patch's start time + `sampling_interval`

ahmadtourei:
> Also, probably goes without saying, but before merging, it would be good to manually run the code to make sure it works 😄. With the exec: false setting the doc build won't actually verify that for us.


Thank you for the comments! Please feel free to re-open if any further improvement is needed.

PR #382, opened 2024-05-31
Adding roll function for patches

Nik-P2:
## Description
Adding in roll function for patches
based on [numpy roll](https://numpy.org/doc/stable/reference/generated/numpy.roll.html) function
an example of this function
```python
import dascore as dc
# roll time dimension 5 elements
rolled_patch = patch.roll(time=5, samples=True)
```

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/proc/basic.py
+++ b/dascore/proc/basic.py
@@ -613,7 +613,7 @@ def pad(
 
 
 @patch_function()
-def roll(patch, samples=False, **kwargs):
+def roll(patch, samples=False, update_coord=False, **kwargs):
     """
     Roll patch array elements along a given dimension.
 
@@ -622,7 +622,9 @@ def roll(patch, samples=False, **kwargs):
     patch
         input patch
     samples
-        If True value indicates coordinate or value of dimension
+        if True, value indicates coordinate or value of dimension
+    update_coord
+        if True, updates coord based on rolled amount
     **kwargs
         specifies dimension and number of elements to roll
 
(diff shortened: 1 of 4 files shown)


d-chambers (approved):
Looks great!

dascore/proc/basic.py
```diff
@@ -606,3 +610,36 @@ def pad(
...
+    >>> rolled_patch2 = patch.roll(distance=30, samples=False)
+    """
+    dim, axis, input_value = get_dim_value_from_kwargs(patch, kwargs)
+    arr = patch.data
+    coord = patch.get_coord(dim)
+    value = coord.get_sample_count(input_value, samples=samples)
+
+    roll_arr = np.roll(arr, value, axis=axis)
+
+    return patch.new(data=roll_arr)
```
ahmadtourei, line 645: Do we need to update the coord as well? Or maybe just adding an optional update_coord argument so when is set True, the function will roll the coord as well?

d-chambers: > Do we need to update the coord as well? Or maybe just adding an optional update_coord argument so when is set True, the function will roll the coord as well

Ya, good point. Normally `Patch.roll` would be used to create a new patch that will broadcast with the old one. For example, in cross correlation. In this case you wouldn't want to coords to update, you are purposefully "rolling" the data around stationary coordinates. However, I could also see a case where you might want to roll the coords as well, so an optional argument to do so makes sense.

d-chambers: @Nik-P2, do you think you can implement this? 

You need to add an optional keyword called `update_coord` or `roll_coord` which is `False` by default. Then if it is `True` do something like this (haven't tested but this should be enough to get you started):

```python
...
if roll_coord:  # Handle rolling coordinate if needed
    coord = patch.get_coord(dim)
    rolled_coord_array = np.roll(coord.values, value, axis)
    new_coord = coord.update(values=rolled_coord_array)
    patch = patch.update_coords(**{dim: new_coord})

return patch.new(data=roll_arr)
```

Make sure to write another test case that runs this logic.

Nik-P2: Yup, I will work on that now.

d-chambers:
Hey @Nik-P2,
If you look at the CI that failed (by clicking on the details button above) I see [this error message](https://github.com/DASDAE/dascore/actions/runs/9334950556/job/25693775391?pr=382)  when running the docstests. I think I told you wrong above, the dimensional coordinates are always 1D, so you might try replacing this line:

```python
np.roll(coord.values, value, axis=axis)
```

with this one:
```python
np.roll(coord.values, value, axis=0)
```

and see if it works.

dascore/proc/basic.py
ahmadtourei: Looks great. Thanks, @Nik-P2 !

d-chambers:
All green, feel free to press the squash and merge button.

PR #385, opened 2024-06-04
Map plot

aissah:
## Description

I added "plot_map", a visualization function that plots the layout of a cable colored by a particular parameter. 

## Checklist

I have (if applicable):


d-chambers:
Nice work @aissah, it looks very configurable. Do you mind posting an example of the output here so we can see what it looks like?

d-chambers:
Looks great!

d-chambers:
Except, why is latitude in ft and longitude in m?

ahmadtourei:
A quick suggestion: since we don't use `plot` in viz's modules (wiggle, waterfall, and spectrogram), I suggest using something like `map_fiber` instead to make it consistent.

aissah:
> Except, why is latitude in ft and longitude in m?

This was an arbitrary choice 😅 illustrating that they can have some units. Would you recommend I change it?

> A quick suggestion: since we don't use `plot` in viz's modules (wiggle, waterfall, and spectrogram), I suggest using something like `map_fiber` instead to make it consistent. What do you think?

I think this is a good idea. What do you think @d-chambers ?

d-chambers:
> I think this is a good idea. What do you think @d-chambers 

Ya, that sounds like a better name to me!

d-chambers:
> This was an arbitrary choice 😅 illustrating that they can have some units. Would you recommend I change it?

Nope, that's perfect. I just wanted to make sure it will use the correct units when we do an actual plot.

aissah:
> > I think this is a good idea. What do you think @d-chambers
> 
> Ya, that sounds like a better name to me!

Done! I think we might be ready to merge after the checks

d-chambers (approved):

PR #364, opened 2024-04-16
update correlate

ahmadtourei:
## Description

This PR adds four new capabilities to the [correlate](https://dascore.org/api/dascore/proc/correlate/correlate.html) function:

- [x] patch can be in the frequency domain, thereby eliminating the need for the Fourier transform step.
- [x] keeps the patch in the frequency domain after computing the correlation if the idft argument is set to False.
- [x] an step_size argument to skip receivers.
- [ ] compute correlation over an array of master sources (passed as kwargs) instead of one master source, which results in a 3D patch.

## Checklist

I have (if applicable):

- [x] referenced the GitHub issue this PR closes.
- [x] documented the new feature with docstrings or appropriate doc page.
- [x] included a test. See [testing guidelines](https://dascore.org/contributing/testing.html).
- [x] your name has been added to the contributors page (docs/contributors.md).
- [ ] added the "ready_for_review" tag once the PR is ready to be reviewed.

PR #400, opened 2024-06-12
Multi source correlate

d-chambers:
## Description

This PR changes the correlate function in a few ways:

1. Multiple sources can now be selected in a single go
2. Removed the `lag` parameter as it was redundant (eg the user should do a `select` on the return value)
3. Greatly simplified the internals to simply use `patch.dft`
4. Adds the `Patch.conj` method for getting the complex conjugate of a patch
5. Adds the `Patch.correlate_shift` for undoing the fft-related shift and scaling
6. Fixes a few small bugs with coords/units. 

supersedes #364 

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/core/coords.py
+++ b/dascore/core/coords.py
@@ -1591,7 +1591,7 @@ def get_coord(
         if isinstance(data, BaseCoord):  # just return coordinate
             return data
         if not isinstance(data, np.ndarray):
-            data = array(data)
+            data = np.atleast_1d(data)
         if np.size(data) == 0:
             dtype = dtype or data.dtype
             return CoordPartial(shape=data.shape, units=units, step=step, dtype=dtype)
--- a/dascore/core/coords.py
+++ b/dascore/core/coords.py
@@ -1371,9 +1371,8 @@ class CoordMonotonicArray(CoordArray):
         stop = self._get_index(v2, forward=True)
         new_stop = stop if stop is not None and stop < len(self) else None
         # We need to add 1 to end so 1 sample get selected if start == stop
-        if new_start is not None and new_stop is not None:
-            eq_values = self.values[new_start] == self.values[new_stop]
-            if eq_values and self.values[new_start] == v1:
+        if new_stop is not None:
+            if self.values[new_stop] == v2:
                 new_stop = new_stop + 1
         out = slice(new_start, new_stop)
         if self._slice_degenerate(out):
@@ -1403,7 +1402,7 @@ class CoordMonotonicArray(CoordArray):
         # right_ok = (right < len(self)) & (right < 0)
         left = np.searchsorted(values, new_value, side="left")
         left_ok = (left < len(self)) & (left > 0)
-        eq = left_ok & (values.take(left, mode='clip') == new_value)
+        eq = left_ok & (values.take(left, mode="clip") == new_value)
         out = right if forward else left
         # where equal it should also be left values. This makes the function
         # behavior consistent with BaseCoord._get_index.
--- a/dascore/core/coords.py
+++ b/dascore/core/coords.py
@@ -1023,6 +1023,9 @@ class CoordRange(BaseCoord):
                 stop = start + step * length
             if pd.isnull(step):
                 step = (stop - start) / length
+                # handle conversion to integer if other values are ints.
+                if isinstance(start, int) and isinstance(stop, int):
+                    step = int(step) if np.isclose(np.round(step), step) else step
         if step != 0:
             int_val = int(np.ceil(np.round((stop - start) / step, 1)))
             stop = start + step * int_val
@@ -1039,7 +1042,7 @@ class CoordRange(BaseCoord):
             raise CoordError(msg)
         # Note: dtype was a property before but it messed up model
         # serialization.
-        values["dtype"] = np.asarray(start).dtype
+        values["dtype"] = np.asarray(start + step).dtype
         return values
 
     def __getitem__(self, item):
@@ -1174,7 +1177,9 @@ class CoordRange(BaseCoord):
         if is_datetime64(self.start) or is_timedelta64(self.start):
             out = np.arange(self.start, self.stop, self.step)
         else:
-            out = np.linspace(self.start, self.stop - self.step, num=len(self))
+            out = np.linspace(
+                self.start, self.stop - self.step, num=len(self), dtype=self.dtype
+            )
         # again, due to round-off error the array can one element longer than
         # anticipated. The slice here just ensures shape and len match.
         return array(out[: len(self)])
--- a/dascore/core/coords.py
+++ b/dascore/core/coords.py
@@ -815,11 +815,11 @@ class BaseCoord(DascoreBaseModel, abc.ABC):
         if samples:
             return array if is_array(value) else array[0]
         # otherwise get forward and backward inds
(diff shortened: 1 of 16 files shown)


dascore/core/coords.py
```diff
@@ -1109,14 +1118,22 @@ def _get_index(self, value, forward=True):
...
+        out = fraction.astype(np.int64)
+        lt_forward = (out < 0) & forward
+        gt_back = (out >= len(self)) & (not forward)
+        bad_values = lt_forward | gt_back
+        if not input_is_array and np.any(bad_values):
             return None
-        return out
+        # Bad array values should be mined out so min finds them later.
+        elif np.any(bad_values):
+            out[bad_values] = -9999
```
ahmadtourei, line 1135: Needs a test here.

dascore/proc/basic.py
```diff
@@ -570,70 +588,83 @@ def pad(
...
         either an integer or a tuple (before, after).
+        In addition, the following strings are supported:
+
+        "fft" - pad to the next fast fft length along the given dimension by
+        adding values to the end of the axis.
+
+        "correlate" - prepare the coordinate for correlation/convolution in
+        the frequency domain by pading to the next fast fft length after
+        2*n - 1 where n is the current dimension length by adding values
+        to the end of the axis.
```
ahmadtourei, line 604: Nice!

dascore/proc/correlate.py
```diff
@@ -3,143 +3,204 @@
...
     samples : bool, optional (default = False)
         If True, the argument specified in kwargs refers to the *sample* not
         value along that axis. See examples for details.
     **kwargs
-        Additional arguments to specify cross correlation dimension and the
-        master source, to which
-        we cross-correlate all other channels/time samples.
+        Specifies correlation dimension and the master source(s), to which
+        we want to cross-correlate all other channels/time samples.If the
+        master source is an array, the function will compute cc for all
```
ahmadtourei, line 116: define cc for the first time here
cross-correlation (cc)

dascore/proc/correlate.py
```diff
@@ -3,143 +3,204 @@
...
+    patch whose target dimension has already been transformed with the
+    [`Patch.dft`](`dascore.transform.fourier.dft`) method, otherwise the dft
+    will be performed. If the input has already been transformed,
+    [`Patch.correlation_shift`](`dascore.proc.correlate.correlate_shift`)
+    is useful to undo dft artefacts after the idft is applied.
+
+    While a 2D patch is required for input, a 3D patch is returned where the
+    3rd dimesion cooresponds to the source rows/columns. For the case of a
+    single source, the [`Patch.squeeze`](`dascore.Patch.squeeze`) method
+    can be helpful to remove length 1 dimensions.
```
ahmadtourei, line 103: I suggest we squeeze under the hood for a single source case so the patch is already 2D as expected.

dascore/proc/correlate.py
```diff
@@ -3,143 +3,204 @@
...
     >>> # Use 2nd channel (python is 0 indexed) along distance as master channel
     >>> cc_patch = patch.correlate(distance=1, samples=True)
-
+    >>>
     >>> # Example 4
     >>> # Correlate along time dimension
     >>> cc_patch = patch.correlate(time=100, samples=True)
+    >>>
+    >>> # Example 5
+    >>> # An example pipeline of frequency domain correlation.
```
ahmadtourei, line 158: # An example pipeline of frequency domain correlation and and an array of sources

dascore/proc/correlate.py
```diff
@@ -3,143 +3,204 @@
...
+    >>> # and the 10 m channel as the master channel. The new patch has dimensions
+    >>> # (lag_time, distance, source_distance)
+    >>> cc_patch = (
+    ...     patch.correlate(distance = 10 * m)
+    ...     .select(lag_time=(-2, 2))
+    ... )
+    >>>
     >>> # Example 3
     >>> # Use 2nd channel (python is 0 indexed) along distance as master channel
     >>> cc_patch = patch.correlate(distance=1, samples=True)
```
ahmadtourei, line 151: It'd be great to adjust this example to showcase how one can skip receivers.

```python
    >>> # Example 3
    >>> #  Correlate along distance having the 5th channel (python is 0 indexed) as a master channel
    >>> #  and skip every other channel for memory efficiency 
    >>> cc_patch = patch.decimate(distance=2).correlate(distance=4, samples=True)
```

dascore/proc/correlate.py
```diff
@@ -3,143 +3,204 @@
...
-
+    >>>
     >>> # Example 4
     >>> # Correlate along time dimension
```
ahmadtourei, line 154: Correlate along time dimension (useful for template matching applications)

Just trying to help search engines :)

ahmadtourei (approved):
Hey @d-chambers, this looks great! I just added a few comments to improve the docs. Thanks for your help in improving the correlate function which will be super helpful in DASDAE's future das-masw package!

dascore/proc/correlate.py
d-chambers: I went back and forth on this. 

On the one hand, it is easier to deal with a 2D patch. On the other though, having the third dimension preserves the information about which channel was used as the source, then the user gets to choose if they want to throw this away with a squeeze or not.

ahmadtourei: Sounds good!

dascore/proc/correlate.py
```diff
@@ -147,15 +147,20 @@ def correlate(
     ... )
     >>>
     >>> # Example 3
-    >>> # Use 2nd channel (python is 0 indexed) along distance as master channel
-    >>> cc_patch = patch.correlate(distance=1, samples=True)
+    >>> # Correlate along distance having the 2nd channel (python is 0 indexed)
+    >>> # as a master channel and skip every other channel to reduce memory
+    >>> cc_patch = (
+    ...     patch.decimate(distance=2, filter_type=None)
```
ahmadtourei, line 153: Just a minor thing but decimating along distance by a factor of 2 results in a patch with a half number of channels. So, having kwargs of `distance=1` basically means correlating the third channel of the original patch before decimation with other channels.

dascore/proc/correlate.py
ahmadtourei: Sorry for back and forth on this but wouldn't be actually better to use select instead of decimate?

```python
distance = patch.get_array("distance")
cc_patch = patch.select(distace=distance[::2]).correlate(distance=2, samples=True)
```

Also, I think there is an issue with `Patch.select` as it does not actually select when an array is passed. I'm on the master branch and refreshed my installation.
```python
import dascore as dc

pa = dc.get_example_patch()

distance = pa.get_array("distance")
selected_pa = pa.select(distace=distance[::2])

assert pa.shape == (300,2000)
assert len(distance) == 300
assert len(distance[::2]) == 150
assert selected_pa.shape == (150,2000) # AssertionError here
```

ahmadtourei: > wouldn't be actually more efficient to use select here instead of decimate?

I think both methods would work in different cases where aliasing may or may not be an issue. We can leave this to the user's choice.

d-chambers:
I just realized the `lag` parameter was already merged (#364 didn't introduce it), so we should go through a deprecation cycle before simply removing it.

PR #406, opened 2024-07-03
Fix rolling center

ahmadtourei:
## Description
This PR fixes two issues regarding the `center=True` in the `Patch.rolling` function when `engine="numpy"`:
1. The first issue was related to using a wrong `-window//2` roll size instead of `-(window//2)` in `np.roll` function to roll NaN values. This could cause unequal NaNs at the beginning and end of the rolled patch.
2. The second issue was related to using `-(window//2)` instead of `-(num_nans//2)` inside the `np.roll` that led to the wrong number of indices to roll when the `step` size is not equal to `None` (and therefore the size of the padded array is different than the original patch's data).

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/proc/rolling.py
+++ b/dascore/proc/rolling.py
@@ -300,7 +300,7 @@ def rolling(
         engines = {"numpy": _NumpyPatchRoller, "pandas": _PandasPatchRoller}
         if cls := engines.get(engine):
             return cls
-        if step < 10 and len(patch.dims) < 2:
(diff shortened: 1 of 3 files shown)


d-chambers:
Hey @ahmadtourei,

I just looked over the code and it looks good. Also, thanks for correcting some of the misleading comments.

d-chambers (approved):

PR #412, opened 2024-07-16
add notes for using apply with rolling

ahmadtourei:
## Description

This PR improves `Patch.rolling` function's docs. It adds notes and an example of how to apply a desired function to the patch along with the rolling operator. 

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/proc/rolling.py
+++ b/dascore/proc/rolling.py
@@ -251,6 +251,10 @@ def rolling(
 
     Notes
     -----
+    Here we have included some useful notes for the rolling function.
+
+    #### Note 1: Length of the output and NaN values
+
     Rolling is designed to behaves like Pandas [DataFrame.rolling](
     https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rolling.html)
     which has some important implications:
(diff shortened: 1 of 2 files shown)


dascore/proc/rolling.py
```diff
@@ -281,6 +285,31 @@ def rolling(
...
+    ```python
+    def zcr_std(frame, axis=-1):
+        '''Compute standard deviation of zero crossing rate using rolling function.'''
+        zero_crossings = (frame[..., :-1] * frame[..., 1:]) < 0
+        return np.std(zero_crossings, axis=axis)
+
+    import dascore as dc
+
+    patch = dc.get_example_patch()
+    zcr_patch = patch.rolling(time=100, step=2, samples=True).apply(zcr_std)
```
d-chambers, line 310: Just for stylistic reason, it would be better to have the import before the function definition.

dascore/proc/rolling.py
```diff
@@ -281,6 +285,31 @@ def rolling(
...
+    Below is an example of applying a custom zero crossing rate function (zcr_std) with
+    rolling operation for a window size of 100 samples and skipping every other samples.
+    It applys the desired operation (which is multiplying every sample to its next
+    sample to determine the zero crossings) on the last axis of the sliced
+    frame (from rolling).
+
+
+    ```python
+    def zcr_std(frame, axis=-1):
+        '''Compute standard deviation of zero crossing rate using rolling function.'''
```
d-chambers, line 303: Double quotes are used for doc strings.

dascore/proc/rolling.py
```diff
@@ -281,6 +285,31 @@ def rolling(
     - if time = 3 * dt and step = 3 * dt
         [NaN, 2.0]
 
+
+    #### Note 2: Applying custom functions with rolling operation
+    When `apply` is the desired rolling operation and we are interested to use our
```
d-chambers, line 290: Probably want a return between the title and text.

d-chambers (approved):

dascore/proc/rolling.py
ahmadtourei: Since this doc string is itself in a doc, I have to use either single quotes or I can just use a hastag. Any alternatives?

d-chambers: Ah good point. Should be fine then.

PR #415, opened 2024-07-21
DFT padding and slope filter simplification

d-chambers:
## Description

This PR adds a `pad` argument to `Patch.dft` so that transformed dimensions can be padded to avoid slow lengths. It also simplifies some of the `Patch.slope_filter` implementation, as well as a few other improvements. 

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/proc/coords.py
+++ b/dascore/proc/coords.py
@@ -236,11 +236,11 @@ def drop_private_coords(self: PatchType) -> PatchType:
     >>> import numpy as np
     >>> import dascore as dc
     >>> pa = (
-    ...     dc.get_example_patch("random_patch")
-    ...     .update_coords(_private=np.array([1,2,3]))
+    ...     dc.get_example_patch("random_das")
+    ...     .update_coords(_private=(None, np.array([1,2,3])))
     ... )
-    >>> pa_no_priv = pa.drop_private_coords()
-    >>> assert "_private" not in pa.coords.coord_map
+    >>> pa_no_private = pa.drop_private_coords()
+    >>> assert "_private" not in pa_no_private.coords.coord_map
     """
     new_coord, data = self.coords.drop_private_coords(array=self.data)
     return self.new(coords=new_coord, dims=new_coord.dims, data=data)
--- a/dascore/proc/coords.py
+++ b/dascore/proc/coords.py
@@ -222,6 +222,31 @@ def drop_coords(self: PatchType, *coords: str | Collection[str]) -> PatchType:
 
 
 @patch_function()
+def drop_private_coords(self: PatchType) -> PatchType:
+    """
+    Drop any provate coords in the patch.
+
+    Parameters
+    ----------
+    self
+        Patch
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import dascore as dc
+    >>> pa = (
+    ...     dc.get_example_patch("random_patch")
+    ...     .update_coords(_private=np.array([1,2,3]))
+    ... )
+    >>> pa_no_priv = pa.drop_private_coords()
+    >>> assert "_private" not in pa.coords.coord_map
+    """
+    new_coord, data = self.coords.drop_private_coords(array=self.data)
+    return self.new(coords=new_coord, dims=new_coord.dims, data=data)
+
+
+@patch_function()
 def make_broadcastable_to(
     self: PatchType,
     shape: tuple[int, ...],
--- a/dascore/proc/coords.py
+++ b/dascore/proc/coords.py
@@ -2,18 +2,17 @@
 
 from __future__ import annotations
 
-from collections.abc import Collection, Sequence
+from collections.abc import Collection
 
 import numpy as np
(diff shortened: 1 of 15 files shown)


d-chambers:
I also added examples of slope filter in the FK page. 

@ahmadtourei or @Shihao-Yuan, do either of you have time to review this in the next few days? Once this merges I think we are good to make another release.

dascore/proc/coords.py
```diff
@@ -232,6 +221,31 @@ def drop_coords(self: PatchType, *coords: str | Collection[str]) -> PatchType:
     return self.new(coords=new_coord, dims=new_coord.dims, data=data)
 
 
+@patch_function()
+def drop_private_coords(self: PatchType) -> PatchType:
+    """
+    Drop any provate coords in the patch.
```
ahmadtourei, line 227: *private

dascore/proc/filter.py
```diff
@@ -406,12 +406,14 @@ def slope_filter(
...
-        and from 'vc' to 'vd'.
+        A length 4 array of the form [va, vb, vc, vd]. If notch is False,
+        the filter selects the apparent velocites between 'vb' and 'vc'
+        with tapering boundaries from 'va' to 'vb' and from 'vc' to 'vd'.
     dims
-        The dimensions used to determine slope.
+        The dimensions used to determine slope. The first dim is in the
+        numerator and the second in the denominator. (eg distance, time)
+        represents a velocity since distance/time has units of |L|/|T|
+        (commonly m/s).
```
ahmadtourei, line 416: Much clearer now.

ahmadtourei (approved):
Hey @d-chambers I just looked over this PR and it looks great. Feel free to merge and then I will publish the next release.

dascore/proc/filter.py
```diff
@@ -391,7 +391,7 @@ def gaussian_filter(
 def slope_filter(
     patch: PatchType,
     filt: Sequence[float],
-    dims: tuple[str, str] = ("time", "distance"),
+    dims: tuple[str, str] = ("distance", "time"),
```
Shihao-Yuan, line 394: I think we should avoid allowing users to assign this parameter. If we use a different order (dims=("time", "distance")), it will lead to incorrect results.

d-chambers: I think we want to leave this in case slopes need to be filtered with coordinates that have different names. This will allow the user to do that. However, I agree it can be a bit confusing, hence the default is what most users will need.

Shihao-Yuan: I agree we could keep it, but assigning it an incorrect order will lead to inaccurate results. If we choose to keep it, we must adjust the code accordingly. Specifically, when calculating apparent velocity, the numerator and denominator should not be determined by their natural order. Instead, their assignment should be based on the order specified in the `dims` variable.

d-chambers: Ya, good point. I think I handled that [here](https://github.com/DASDAE/dascore/blob/8cc1a6f3c5d8da7e18b0005a3750f70d6b142d32/dascore/proc/filter.py#L492-L507) but you are welcome to do additional testing if you like.

Shihao-Yuan: > Ya, good point. I think I handled that [here](https://github.com/DASDAE/dascore/blob/8cc1a6f3c5d8da7e18b0005a3750f70d6b142d32/dascore/proc/filter.py#L492-L507) but you are welcome to do additional testing if you like.

I've tested it (`dims= ("time","distance",)`) and found that wrong results before I shared the first comment. But you may double check that.

d-chambers: > I've tested it (dims= ("time","distance",)) and found that wrong result

Yes, in that case you would be filtering on slowness, as the docstring says the slope is interpreted as the first over the second value, so ("time", "distance") is time/distance. That is why the default  is distance/time (velocity), but I also added support for units so like the `pass_filter` you can indicate slowness or velocity explicitly with units. 

I did some testing in the [fk/slope filter recipe](https://dascore.netlify.app/recipes/fk) and it looked ok. Mind taking a look?

Shihao-Yuan: Those examples seem perfect. I think instead of letting users specify units, it might be more straightforward to choose between slowness or apparent velocity explicitly. Using the order of `time` and `distance` in the `dims` to determine whether it's a slowness filter or an apparent velocity filter is a bit unconventional. So I suggest that if we want to maintain flexibility while improving clarity, we could replace `dims` with `filt_type='slowness' or 'apparent velocity'`. What do you think?

d-chambers: I think support units is important for consistency (eg between `pass_filter` and `slope_filter`. I get the concern of ordering the dims not being the most intuitive way for a standard "time", "distance" patch, but wherever possible we are trying to ensure the patch methods are general enough to apply to any dimensions. 

If you want, you could add a separate method called `velocity_filter` which would just call `slope_filter` but might have a different interface that is more straight-forward for the "time" "distance" case.

PR #426, opened 2024-08-26
fix #346 scan robust to corrput file

### LINKED ISSUE #346: Scan not robust to corrupted file
## Description

@jinwar and his student collected a dataset which has a corrupted file. When attempting to read the time array in this file h5py raises an error stating the metadata checksum is bad. When trying to index this directory with `spool.update` the error bubbles to the surface and crashes the indexing.  This is because `dascore.scan` is not robust to files which have their file format correctly identified but then raise an error when attempting to read them. 

Although the presence of a corrupted file is certainly not something we can control, DASCore should be robust to this, issue a warning, then skip the problematic file and move on with indexing. This may be a bit tricky to test as the corrupted file is too large to include in the test suite. So, we could either:

1. Muck about with the bytes of a small test file until we can produce this same error, or
2. Create a test in which we monkey patch one of the formatter scan functions to raise an error if a specific file is given  

## Example
-- Can't currently reproduce with a self contained code snippet. 

## Versions
 - OS [e.g. Ubuntu 20.04]: CentOS
 - DasCore Version [e.g. 0.0.5]: 0.1.0
 - Python Version [e.g. 3.10]: 3.11

d-chambers:
## Description

This PR makes `dc.scan` robust to corrupted files whose format is determinable but when they are scanned they HDF5 raises an OSError. See #346.

closes #346.

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/io/core.py
+++ b/dascore/io/core.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import inspect
 import os.path
+import warnings
 from collections import defaultdict
 from collections.abc import Generator
 from functools import cache, cached_property, wraps
@@ -712,7 +713,7 @@ def _get_fiber_io_and_req_type(
 
     Raises
     ------
-    UnkownFileFormatError if no format is determinable from the
+    UnknownFileFormatError if no format is determinable from the
     patch_source
 
     """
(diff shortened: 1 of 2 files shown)


d-chambers:
Hey @ahmadtourei,

It looks like the issue is back. I guess we need to disable the cache for macos.

ahmadtourei:
> Hey @ahmadtourei,
> 
> It looks like the issue is back. I guess we need to disable the cache for macos.

Ah I see. Yes, I agree.

PR #430, opened 2024-09-04
add_distance_to

d-chambers:
## Description

This PR adds a `add_distance_to` patch method for calculating the distance to a point specified by a series. It can then be used to sort the channels of the patch. See #429. 

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/proc/coords.py
+++ b/dascore/proc/coords.py
@@ -615,9 +615,11 @@ def add_distance_to(
     """
     Calculate the distance to "origin" and create new coordinate.
 
-    A new coordinate called `origin_distance` is added to the output patch
-    to specify the exact distance, and the original location is added to
-    the patch attrs (e.g, patch.attrs.origin_x, patch.attrs.origin_y, ...)
+    A new coordinate called `origin_distance` (or another name controlled
+    by the pre-fix argument) is added to the output patch to specify the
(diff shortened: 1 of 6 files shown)


dascore/proc/coords.py
```diff
@@ -606,3 +606,68 @@ def squeeze(self: PatchType, dim=None) -> PatchType:
...
+
+    Examples
+    --------
+    >>> import pandas as pd
+    >>>
+    >>> import dascore as dc
+    >>>
+    >>> shot = pd.Series({"x": 10, "y": 10, "z": 0})
+    >>> patch = dc.get_example_patch("random_patch_with_xyz")
+    >>> patch_with_distance = patch.add_distance_to(shot)
```
ahmadtourei, line 645: Suggest calling it `patch_with_origin_dist`

ahmadtourei (approved):
Looks great!

d-chambers:
Thanks @ahmadtourei. I implemented your suggestion. I also got thinking and decided to store the origin location as coordinates (not associated with any dimension) rather than attributes. My thinking here is that we may want to support origin being a dataframe in the future, in which case the distance would be the shortest distance to any point in that dataframe. In such a  case we would need to store arrays rather than single floats for each coordinate of the origin so this change helps future proof this method.

PR #448, opened 2024-10-16
notch for Patch

ahmadtourei:
## Description

This PR adds notch filtering functionality for `Patch` as discussed in #443.

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/proc/filter.py
+++ b/dascore/proc/filter.py
@@ -296,11 +296,11 @@ def notch_filter(patch: PatchType, q, **kwargs) -> PatchType:
     >>> filtered = pa.notch_filter(time=60, distance=0.2, q=40)
 
     >>> # Optionally, units can be specified for a more expressive API.
-    >>> from dascore.units import m, ft, s, Hz
+    >>> #from dascore.units import m, ft, s, Hz
     >>> # Apply a notch filter along time axis to remove 60 Hz
-    >>> filtered = pa.notch_filter(time=60 * Hz, q=30)
+    >>> #filtered = pa.notch_filter(time=60 * Hz, q=30)
     >>> # Apply a notch filter along distance axis to remove 0.2 m wavelength
-    >>> filtered = pa.notch_filter(distance=0.2 * m, q=30)
+    >>> #filtered = pa.notch_filter(distance=0.2 * m, q=30)
     """
     size, axes = _create_size_fw_and_axes(patch, kwargs)
     data = patch.data
--- a/dascore/proc/filter.py
+++ b/dascore/proc/filter.py
@@ -15,7 +15,7 @@ import pandas as pd
 from scipy import ndimage
 from scipy.ndimage import gaussian_filter as np_gauss
 from scipy.ndimage import median_filter as nd_median_filter
-from scipy.signal import iirfilter, sosfilt, sosfiltfilt, zpk2sos
+from scipy.signal import filtfilt, iirfilter, iirnotch, sosfilt, sosfiltfilt, zpk2sos
 from scipy.signal import savgol_filter as np_savgol_filter
 
 import dascore as dc
@@ -94,7 +94,7 @@ def pass_filter(patch: PatchType, corners=4, zerophase=True, **kwargs) -> PatchT
     Parameters
     ----------
     corners
-        The number of corners for the filter.
+        The number of corners for the filter. Default is 4.
     zerophase
         If True, apply the filter twice.
     **kwargs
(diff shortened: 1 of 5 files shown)


ahmadtourei:
Hey @d-chambers, please review and let me know what you think. Also, I wonder how I can adjust the `_create_size_fw_and_axes(patch, kwargs)` function so it can accept specified units as well (please see two commented tests in tests/test_proc/test_filter.py) Thanks!

d-chambers (changes requested):
Looks ok, just a few things to tweak on the docstring stuff.

d-chambers:
> Also, I wonder how I can adjust the _create_size_fw_and_axes(patch, kwargs) function so it can accept specified units as well (please see two commented tests in tests/test_proc/test_filter.py) Thanks!

Probably the easiest thing to do for the units is to pull [`get_inverted_quant`](https://github.com/DASDAE/dascore/blob/3d635c1333a106ec5b48a65a7f463e8424cbf190/dascore/units.py#L253-L273) from being a nested function to being on the `dascore.units` level. Then you can do something like this (but I haven't tested it yet):

```python
from dascore.units import get_inverted_quant

size, axes = _create_size_fw_and_axes(patch, kwargs)
data = patch.data
for coord_name, info in get_multiple_dim_value_from_kwargs(patch, kwargs).items():
    dim, ax, value = info['dim'], info['axis'], info['value']
    coord = patch.get_coord(coord_name, required_evenly_sampled=True)  # will also raise if not evenly sampled.

    # Invert units if needed
    inverted_units = False
    if isinstance(value, dc.units.Quantity) and coord.units is not None:
        value, inverted_units = get_inverted_quant(value, coord.units)
    
    # Check valid parameters
    sr, value = to_float(coord), to_float(value)
    nyquist = 0.5 * sr 
    if w0 > nyquist:
        msg = f"possible filter values are in [0, {nyquist}] you passed {w0}"
        raise FilterValueError(msg)

    b, a = iirnotch(value, Q=q, fs=sr)
    data = filtfilt(b, a, data, axis=ax)

...
```


```

File: dascore/proc/filter.py [hunk 1]
```diff
@@ -265,6 +257,66 @@ def median_filter(
...
+    A notch filter is a band-stop filter with a narrow bandwidth (high quality factor).
+    It rejects a narrow frequency band and leaves the rest of the spectrum
+    little changed.
+
+    Parameters
+    ----------
+    patch
+        The patch to filter
+    samples
+        {sample_explination}
```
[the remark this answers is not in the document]
ahmadtourei: Since this is a filter, we actually do not need the `samples` argument.

dascore/proc/filter.py
```diff
@@ -265,6 +257,66 @@ def median_filter(
...
+    q
+        Quality factor (float). See [scipy.signal.iirnotch]
```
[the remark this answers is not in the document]
ahmadtourei: Done!

ahmadtourei:
> ```python
>         value, inverted_units = get_inverted_quant(value, coord.units)
> ```

Thanks for this explanation, Derrick! I'm unsure why the `inverted_units` is being reported False when the `get_inverted_quant` actually has inverted the unit! Do we need to fix this inside the `get_inverted_quant` function?

d-chambers:
> Do we need to fix this inside the get_inverted_quant function?

So the function returns the correctly inverted quantity but just the boolean is wrong? Yes, I suppose we should fix it, hopefully it hasn't caused issues elsewhere. 

Sorry about the linter failing. It has to do with newer versions of to pip I think. I will push a fix in a few minutes.

d-chambers:
> So the function returns the correctly inverted quantity but just the boolean is wrong? Yes, I suppose we should fix it, hopefully it hasn't caused issues elsewhere.

Actually, looking closer I think it is correct, just a bit confusing. With filtering we do expect the dimensionality of inputs to the be inverse of the coordinate units. For example, for the time dimension (seconds) we expect the filter parameters to be the multiplicative inverse dimensionality (1/seconds, Hz). So the second parameter returned from this function just indicates if the dimensionality of the filter argument were already the inverse dimensionality. Its probably safe to just ignore the second output parameter.

dascore/proc/filter.py
```diff
@@ -265,6 +249,80 @@ def median_filter(
...
+        inverted_units = True
+        # Invert units if needed
+        if isinstance(value, dc.units.Quantity) and coord.units is not None:
+            value, inverted_units = get_inverted_quant(value, coord.units)
+
+        # Check valid parameters
+        if inverted_units:
+            w0 = to_float(value)
+        else:
+            w0 = to_float(1 / value)
```
d-chambers, line 315: hmmm... its a bit odd this is required. Doesn't `get_inverted_quant` already do this?

d-chambers: Specifically [here](https://github.com/DASDAE/dascore/blob/3d635c1333a106ec5b48a65a7f463e8424cbf190/dascore/units.py#L266-L267).

d-chambers (approved):
Looks good @ahmadtourei, just double check the unit inversion stuff is right then merge when you are happy with it.

dascore/proc/filter.py
ahmadtourei: Thanks for looking into this more closely. I did not understand the behavior of the `get_inverted_quant` function well. We actually don't need this. Please see my last commit (changed examples and a test) for further details.

ahmadtourei:
closes #443.

PR #456, opened 2024-11-03
add dim_reduce to aggregates, refactor patch utils

d-chambers:
## Description

This PR does 3 sorta related things:

1. Adds a `dim_reduce` keyword to `Patch.aggregate` and friends (eg mean, min, max) which controls what happens to the aggregated dimension. Options now are to apply another numpy-like reducer, squeeze out the dimension, or empty it. The default behavior is unchanged, but the additional options are useful when calculating stats over patches in a spool.

2. Unified getting dimension name, axis, and values from args and kwargs into one new function `dascore.utils.patch.get_dim_axis_value` and replaced usages of two overlapping functions that did  this before. 

3. Added a `validate_call` option to `patch_function` which uses pydantic's [validate_call decorator](https://docs.pydantic.dev/latest/concepts/validation_decorator/#using-field-to-describe-function-arguments) to do type checking on patch functions when enabled. This makes it much easier to enforce certain types and ranges on patch function parameters. 

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/proc/aggregate.py
+++ b/dascore/proc/aggregate.py
@@ -8,11 +8,13 @@ from typing import Literal
 
 import numpy as np
 
+import dascore as dc
 from dascore.constants import PatchType
 from dascore.exceptions import ParameterError
 from dascore.utils.docs import compose_docstring
 from dascore.utils.misc import iterate
 from dascore.utils.patch import get_dim_axis_value, patch_function
+from dascore.utils.time import dtype_time_like, is_datetime64, is_timedelta64
 
 _AGG_FUNCS = {
     "mean": np.nanmean,
@@ -55,6 +57,9 @@ coord_mode
         squeeze - remove the aggregated dimension.
         min - keep the min value of the aggregated dimension. 
         mean - keep the mean value of the aggregated dimension.
+    If the operation fails and the coords are a time-type it will be tried
+    again converting to, then back, from floats. This can cause a small 
+    loss of precision.
 """
 
 _COORD_MODE_TYPE_HINT = Literal[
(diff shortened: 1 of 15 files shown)


dascore/proc/aggregate.py
```diff
@@ -31,6 +39,14 @@
     The dimension along which aggregations are to be performed.
     If None, apply aggregation to all dimensions sequentially.
     If a sequence, apply sequentially in order provided.
+dim_reduce
+    How to reduce the dimensional coordinate associated with the 
+    aggregated axis. Can be the name of any valid aggregator, a callable,
+    "empty" - which returns and empty coord, or "squeeze" which drops
```
ahmadtourei, line 45: It would be helpful to note that "empty" is the default behavior.

dascore/proc/aggregate.py
```diff
@@ -41,12 +57,44 @@
...
 @patch_function()
 @compose_docstring(params=AGG_DOC_STR, options=list(_AGG_FUNCS))
 def aggregate(
     patch: PatchType,
     dim: str | Sequence[str] | None = None,
     method: str | Callable = "mean",
+    dim_reduce: str | Callable = "empty",
 ) -> PatchType:
     """
     Aggregate values along a specified dimension.
```
ahmadtourei, line 100: The notes section in the docs should be revised based on the new `dim_reduce` argument.

dascore/utils/patch.py
```diff
@@ -428,40 +458,90 @@ def _format_datetime64(dt):
...
+
+    Returns
+    -------
+    Returns a tuple of:
+        ((dim, axis, value), (dim, axis, value), ...)
+    To support retreiving multiple values from the same inputs. If dim name
+    is found in args, its corresponding values is `None`.
 
-    Return the name of the dimension, its axis position, and its value.
+    Examples
```
ahmadtourei, line 499: Thanks for including these examples in the docs. Super helpful as this function is being used in a lot of other functions.

ahmadtourei (approved):
Looks great! Please see my comments on a couple of suggestions.

PR #465, opened 2024-12-04
neubrex decimation check

d-chambers:
## Description

This PR modifies the time/distance decimation check for reading neubrex DAS files. Previously, we simply raise an exception if the `DistanceDecimationFilter` or the `TimeDecimationFilter` attributes are non-zero. This is because I was not sure if this is a binary flag indicating a low-pass anti-aliasing filter was applied (to either the time or space dimension) or if it is an integer value indicating the time/distance coordinate was decimated. For example, a `DistanceDecimationFilter`=2 could mean every other channel was discarded. 

 It appears that some files (eg the [forge neubrex das files](https://data.openei.org/s3_viewer?bucket=gdr-data-lake&prefix=FORGE%2FDAS%2FNeubrex)) have a value of 1, so it is probably a binary flag. In this case, I don't think there is anything else for DASCore to do except add this info to the attrs. 

This PR just allows `DistanceDecimationFilter` and `TimeDecimationFilter` to be either 1 or 0 and adds `distance_decimation_filter` and `time_decimation_filter` to the patch attrs. 

@arturguzik, if you can confirm my intrepretation of the `TimeDecimationFilter` and `DistanceDecimationFilter` parameters that would be useful. Thanks! 

## Checklist

I have (if applicable):

### DIFF
--- a/.github/workflows/run_min_dep_tests.yml
+++ b/.github/workflows/run_min_dep_tests.yml
@@ -51,15 +51,18 @@ jobs:
           git fetch --tags --force # Retrieve annotated tags.
 
       # Have to use conda to install hdf5 so mac (I guess m1/2) runners work.
-      - uses: mamba-org/setup-micromamba@v1
+      - uses: mamba-org/setup-micromamba@v2
         with:
-          micromamba-version: '1.5.8-0' # versions: https://github.com/mamba-org/micromamba-releases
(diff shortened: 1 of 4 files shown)


d-chambers:
I also updated the mamba setup to actually use the correct version of python (it appears the latest version was being used before).

PR #469, opened 2024-12-19
Optional segyio and segy write

d-chambers:
## Description

This PR removes segyio as a hard dependency. Segyio is great, but usually very slow to upgrade to newer python versions which, in tern, keeps DASCore from upgrading.  This PR fixes this. 

It also implements very simple support for writing segy files as requested in #428. This will probably require more work in the future to get right, but at least we have a start here.

@aaronjgirard, do you mind reviewing this PR? 

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/io/segy/core.py
+++ b/dascore/io/segy/core.py
@@ -7,7 +7,7 @@ import segyio
 import dascore as dc
 from dascore.io.core import FiberIO
 
-from .utils import _get_attrs, _get_coords, _get_filtered_data_and_coords
+from .utils import _get_attrs, _get_coords, _get_filtered_data_and_coords, _is_segy
 
 
 class SegyV2(FiberIO):
@@ -21,11 +21,13 @@ class SegyV2(FiberIO):
 
     def get_format(self, path, **kwargs) -> tuple[str, str] | bool:
         """Make sure input is segy."""
-        try:
-            with segyio.open(path, ignore_geometry=True):
-                return self.name, self.version
-        except Exception:
-            return False
+        with open(path, "rb") as fp:
+            return _is_segy(fp)
+        # try:
+        #     with segyio.open(path, ignore_geometry=True):
+        #         return self.name, self.version
+        # except Exception:
+        #     return False
 
     def read(self, path, time=None, channel=None, **kwargs):
         """
--- a/dascore/io/segy/core.py
+++ b/dascore/io/segy/core.py
@@ -2,32 +2,33 @@
 
 from __future__ import annotations
 
-import segyio
-
 import dascore as dc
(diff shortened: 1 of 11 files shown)


d-chambers:
I am particularly interested if you have ideas for improving the detection of segy files found [here](https://github.com/DASDAE/dascore/blob/2c253c6dd3aa67f9f38f2436f85d9c2737d28498/dascore/io/segy/utils.py#L39-L79).

dascore/io/segy/core.py
```diff
@@ -55,7 +61,75 @@ def scan(self, path, **kwargs) -> list[dc.PatchAttrs]:
...
+
+        spec = segyio.spec()
+        version = self.version
+
+        # spec.sorting = 2
+        spec.format = 1  # 1 means float32 TODO look into supporting more
+        spec.samples = np.ones(len(time)) * len(channel)
+        spec.ilines = range(len(channel))
+        spec.xlines = [1]
+        # breakpoint()
```
d-chambers, line 94: remove

dascore/io/segy/utils.py
```diff
@@ -3,14 +3,80 @@
...
+
+# Valid data format codes as specified in the SEGY rev1 manual.
+VALID_FORMATS = [1, 2, 3, 4, 5, 8]
+
+# This is the maximum possible interval between two samples due to the nature
+# of the SEG Y format.
+MAX_INTERVAL_IN_SECONDS = 0.065535
+
+# largest number possible with int16
+MAX_NUMBER_OF_SAMPLES = 32767
```
d-chambers, line 27: Do we need these?

d-chambers:
This is failing. @aaronjgirard and I discussed `spool(data_directory_with_segy).update()` should issue a warning while `spool(segy_file.segy)` should raise. In both cases the message should indicate segyio needs to be installed.

Also, for now we are just going to support converting patches with a dimension of "channel" to segy. In the future we will figure out how to handle 'distance' or 'x', 'y' coords.

d-chambers:
We also need to add a test for standard patch conversion

dascore/io/segy/utils.py
aaronjgirard: It looks like we don't since they are edge cases. But for the writer, we may want to ensure we don't make a trace with more than max number of samples. Since DAS can be very small dt and very large time, this could become an issue.

d-chambers:
Hey @aaronjgirard,

I am going to merge this. I think I implemented everything we planned on for the first pass at segy write, and we really need this fix to support python 3.13. No doubt, we will have improvements to make to segy in the future!

Issue #10, opened 2021-12-31
Inconsistencies in decimate behavior


d-chambers:
**Description**

Slack message from Jin: 

There seems to be some bugs with decimate
patch1:
["select(copy=False,time=(numpy.datetime64('2015-11-09T21:17:00'), numpy.datetime64('2015-11-09T21:18:10')))",
 "decimate(copy=True,dim='time',factor=4,lowpass=True)",
 "decimate(copy=True,dim='time',factor=5,lowpass=True)",
 "decimate(copy=True,dim='time',factor=10,lowpass=True)",
 "decimate(copy=True,dim='time',factor=10,lowpass=True)"]

patch2:
["select(copy=False,time=(numpy.datetime64('2015-11-09T21:17:00'), numpy.datetime64('2015-11-09T21:18:10')))",
 'pass_filter(corners=4,time=(None, 0.5),zerophase=True)',
 "decimate(copy=True,dim='time',factor=2000,lowpass=False)"]

patch3:
data = p[0].data.copy()
lpdata = gjsignal.lpfilter(data,0.0005,0.5,axis=1)[:,::2000]
plt.figure()
plt.plot(dsp.data[200,:],label='patch.decimate')
plt.plot(dsp2.data[200,:],label='patch.pass_filter.decimate(lowpass=False)')
plt.plot(lpdata[200,:],label='gjsignal.lpfilter')
plt.legend()

![image](https://user-images.githubusercontent.com/11671536/147839776-4643c3c4-91a6-42d9-a983-487d2380df1e.png)

d-chambers:
Hey @jinwar,

The current decimate implementation was adopted from obspy's decimate which uses a [Cheby2-Lowpass](https://github.com/obspy/obspy/blob/c4faef6b2df110e844c28ca2aa9515888c019cb8/obspy/signal/filter.py#L348-L395) rather than a butterworth lowpass (which `pass_filter` uses). 

In order to get something up and going quickly I just decided to replicate ObsPy's behavior. Honestly, I would need to look more into on why the approaches are different and which one is better for dascore, but this at least explains the difference.

d-chambers:
Scipy just fixed the numerical instability issue with it's decimate function [here](https://github.com/scipy/scipy/issues/15072#event-7852095689). It should be included in 1.10, so we can just use scipy decimate and pin the scipy version >= 1.10.

PR #470, opened 2024-12-19
Low-freq proc

ahmadtourei:
## Description

This PR adds a recipe for the low-frequency processing routine. It also adds a delta patch (with unit value at the center of time or distance dimensions and zeros elsewhere) as an example patch. 

This PR, in conjunction with #341, deprecates DASDAE's [low-freq-real-time](https://github.com/DASDAE/low-freq-real-time) package.

## Checklist

I have (if applicable):


dascore/examples.py
```diff
@@ -415,6 +415,98 @@ def _ricker(time, delay):
...
+            delta_patch = dc.Patch(data=data, coords=coords, dims=dims, attrs=attrs)
+            return delta_patch.select(time=0, samples=True)
+    else:
+        if dim == "time":
+            patch = patch.select(distance=0, samples=True)
+        else:
+            patch = patch.select(time=0, samples=True)
+        data = np.zeros(patch.data.size)
+        data[len(data) // 2] = 1
+        return patch.update(data=data.reshape(patch.shape))
```
d-chambers, line 507: Two things here. 

First, perhaps we just do the dim check for the case no patch is provided? It should be doable to support any dimension name.

Second, I am not sure this will work with a transposed patch. It seems like it will always add the ones along the first dimension, but, if I recall correct, select will return a 2D patch (with a length of one in one of the dimensions). 

Off the top of my head, I think something like this will fix it:

```python
patch = patch.select(**{dim: 0}, samples=True)
axis = patch.dims.index(dim)
data = np.zeros_like(patch.data)
np.put_along_axis(data, data.shape[axis]//2, 1, axis=axis)
return patch.update(data=data)
```
Using zeros_like will ensure the dtype is correct.

docs/recipes/low_freq_proc.qmd
```diff
@@ -0,0 +1,183 @@
+---
+title: "Low-Frequency Processing"
+execute:
+  eval: false
```
d-chambers, line 4: Is there any way we can remove this? It would be nice if the code is forced to run to ensure the example always works (or the docs won't build).

docs/recipes/low_freq_proc.qmd
```diff
@@ -0,0 +1,183 @@
...
+---
+
+This recipe demonstrates how DASCore can be used to apply low-frequency (LF) processing to a spool of DAS data.
```
d-chambers, line 7: Maybe we can state this is equivalent to downsampling an entire spool in case someone isn't familiar with what "low-frequency processing" means.

docs/recipes/low_freq_proc.qmd
```diff
@@ -0,0 +1,183 @@
...
+```{python}
+# Retrieve a patch of appropriate size for LF processing that fits into memory
+pa_chunked_sp = sp.chunk(time=chunk_size)[0] 
+# Create a delta patch based on new patch size
+delta_pa = dc.get_example_patch("delta_patch", dim="time", patch=pa_chunked_sp)
+
+# Apply the low-pass filter on the delta patch
+delta_pa_low_passed = delta_pa.pass_filter(time=(None, cutoff_freq * filter_safety_factor))
+# Resample the low-passed filter patch
+new_time_ax = np.arange(delta_pa.attrs["time_min"], delta_pa.attrs["time_max"], np.timedelta64(dt, "s"))
```
d-chambers, line 81: Can you just use `patch.resample` here?

docs/recipes/low_freq_proc.qmd
```diff
@@ -0,0 +1,183 @@
...
+    Args:
+        start_time (np.datetime64): The start time of the data range.
+        end_time (np.datetime64): The end time of the data range.
+
+    Returns:
+        str: A file name in the format 'LFDAS_<start_time>_<end_time>.h5'.
+    """
+    start_time_str = _format_time_as_string(start_time)
+    end_time_str = _format_time_as_string(end_time)
+    return f"LFDAS_{start_time_str}_{end_time_str}.h5"
```
d-chambers, line 142: Take a look at `dascore.utils.patch.get_default_patch_name` and see if it can replace this.

d-chambers (changes requested):
Hey @ahmadtourei, it's looking good so far. I just have two main suggestions.

1. It would be nice if the delta patch wasn't restricted to time/distance if you pass in a template patch. I provided the code that (probably) fixes that. 

3. Do you think we could make the recipe executable? That way, we will have much more assurance the code continues to work as DASCore evolves (or the docs wont build). Moreover, we aren't using any dependencies we don't want to include like we did in the other non-executable recipes (eg the MPI one) so I don't see why not as long as the example spool is small, but maybe you have a reason I didn't think of?

dascore/examples.py
ahmadtourei: I agree we can support any dimension names if a patch is provided. 

Regarding the second point, the code works for a transposed patch as I'm selecting first in an if statement as below:
```{python}
        if dim == "time":
            patch = patch.select(distance=0, samples=True)
        else:
            patch = patch.select(time=0, samples=True)
```

ahmadtourei: > ```python
> patch = patch.select(**{dim: 0}, samples=True)
> ```

The dim value here actually should be other_dim (for `dim="time"` it should be `patch.select(distance=0, samples=True)`. How can I do that when the provided patch can have any dimensions?

docs/recipes/low_freq_proc.qmd
ahmadtourei: Apparently, they do not provide the same results:
```python
import numpy as np
import dascore as dc

pa = dc.get_example_patch()
dt = 5 

resampled_pa = pa.resample(time=np.timedelta64(dt, 's')) # has one time sample at zero

new_time_ax = np.arange(pa.attrs["time_min"], pa.attrs["time_max"], np.timedelta64(dt, "s"))
interpolated_pa = pa.interpolate(time=new_time_ax) # has two time samples at 0 and 5 sec
```

docs/recipes/low_freq_proc.qmd
```diff
@@ -0,0 +1,183 @@
...
+# Identify the indices where the absolute value of the data exceeds the threshold
+data_abs = np.abs(delta_pa_lfp.data)
+threshold = np.max(data_abs) * tolerance
+ind = data_abs > threshold
+ind_1 = np.where(ind)[1][0]
+ind_2 = np.where(ind)[1][-1]
+
+# Get the total duration of the processed delta function patch in seconds
+delta_pa_lfp_seconds = (delta_pa_lfp.attrs["time_max"] - delta_pa_lfp.attrs["time_min"]) / np.timedelta64(1, "s")
+# Convert the new time axis to absolute seconds, relative to the first timestamp
```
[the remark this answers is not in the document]
ahmadtourei: Done!

docs/recipes/low_freq_proc.qmd
ahmadtourei: I believe it can.

docs/recipes/low_freq_proc.qmd
ahmadtourei: Agreed, but how can I exclude parts such as saving the processed data, creating a new spool, and visualizing other than commenting these lines (lines > 133)?

docs/recipes/low_freq_proc.qmd
d-chambers: Ah, that is odd. I think #10 might be related.

docs/recipes/low_freq_proc.qmd
d-chambers: Why not just do them and show the output in the recipe? 

You can use `dascore.example.spool_to_directory` to save the spool to a directory. If you don't pass a path it will use an OS-specific temporary directory. `dascore.examples.random_spool` might be the easiest spool to create contiguous patches (although the data are boring).

dascore/examples.py
d-chambers: Ah yes, I see. I guess this calls for a little set math. 

What do you think of this? Should work with any name/number of dimensions. 

```python
# from dascore.utils.misc import iterate

used_dims = tuple(iterate(dim))
unused_dims = set(patch.dims) - set(dims) 
patch = patch.select(**{x: 0 for x in unused_dims}, samples=True)

# get data with ones centered on selected dimensions.
shape = patch.shape
index = tuple(
    shape[patch.dims.index[dim]] // 2 if dim in used_dims else 0 
    for dim in patch.dims
)
data = np.zeros_like(patch.data)
data[index] = 1

return patch.new(data=data)
```

docs/recipes/low_freq_proc.qmd
ahmadtourei: I have a feeling that this might be a bit confusing, similar to the "fetch" concept in earlier tutorials. However, how about this?

```python
for patch in sp_chunked_overlap:
    # rest of the code

    # Save processed patch 
    pa_lf_name = get_patch_name(pa_lfp_edgeless)
    pa_lf_path = os.path.join("/path/to/output/dir", pa_lf_name)
    # pa_lfp_edgeless.io.write(pa_lf_path, "dasdae") # commented to make this doc executable 

    # Add the processed patch to a list (so later we can create a spool)
    lf_patches = lf_patches.append(pa_lfp_edgeless)


# Visualize the results
```{python}
# Create a spool of LF processed results
# sp_lf = dc.spool(output_data_dir) # commented to make this doc executable 
sp_lf = dc.spool(lf_patches)

# Merge the spool and create a single patch. May need to sub-select before merging to prevent exceeding the memory limit.
sp_lf_merged = sp_lf.chunk(time=None)
pa_lf_merged = sp_lf_merged[0]

# Visualize the results. Try different scale values for better visualization.
pa_lf_merged.viz.waterfall(scale=0.1)
```

d-chambers: Looks good

dascore/examples.py
ahmadtourei: Works! thanks

d-chambers (approved):

d-chambers:
Feel free to merge when you are happy with it. Do make sure the docs build before you do though.

docs/recipes/low_freq_proc.qmd
ahmadtourei: I haven't worked with `patch.resample()` much; do you think this behavior is expected?

ahmadtourei:
> Feel free to merge when you are happy with it. Do make sure the docs build before you do though.

Thanks for your input and review! Let's merge the get_name branch first and then I'll merge this.

ahmadtourei:
The `api/dascore/proc/filter/slope_filter.qmd` doc fails with this error:
```python
AttributeError: module 'dascore.io.segy.core' has no attribute 'SegyV2'
```
I've merged master branch (after having the [optional_segyio](https://github.com/DASDAE/dascore/tree/optional_segyio) branch merged into the master's). Any thoughts on how this can be fixed? @d-chambers

d-chambers:
> Any thoughts on how this can be fixed?

Yes, this is a quirk of python's entry points. Any time you change them (eg with the IO functions) you have to re-install dascore. Try re-installing. 

Merging `get_name` into this branch is probably not prudent. When we merge `get_name` into master it will squash all the commits, then when we merge master back into this branch there will probably be a few merge conflicts. You can either undo that commit with `git revert` or just wait and deal with the merge conflicts.

ahmadtourei:
@d-chambers The only thing left here is resolving #474. It will be ready to merge after addressing that issue.

docs/recipes/low_freq_proc.qmd
d-chambers: In both cases they are just calling underlying scipy functions so I don't think there is much we can do about it anyway.

PR #461, opened 2024-11-27
patch_name method for dir spools

ahmadtourei:
## Description

This PR adds a method named `patch_name()` that returns the name of a patch in a directory spool.
It is related to discussion #460.

## Checklist

I have (if applicable):

- [ ] referenced the GitHub issue this PR closes.
- [ ] documented the new feature with docstrings or appropriate doc page.
- [ ] included a test. See [testing guidelines](https://dascore.org/contributing/testing.html).
- [ ] your name has been added to the contributors page (docs/contributors.md).
- [ ] added the "ready_for_review" tag once the PR is ready to be reviewed.

PR #471, opened 2024-12-21
add get_name patch/spool method

d-chambers:
## Description

This PR implements a `get_name` method for the patch and the spool to return a series of names. The names are either generated by columns provided to the function (which has sensible defaults to maintain the old behavior) or by existing columns such as `name` or `path`. 

This PR supersede  #461.

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/examples.py
+++ b/dascore/examples.py
@@ -451,6 +451,24 @@ def random_spool(time_gap=0, length=3, time_min=np.datetime64("2020-01-03"), **k
     return dc.spool(out)
 
 
+@register_func(EXAMPLE_SPOOLS, key="radom_directory_das")
+def random_directory_spool(path=None, **kwargs):
+    """
+    Create a random spool, then save to specified path.
+
+    Parameters
+    ----------
+    path
+        If provided, the path to save the directory spool. If None, use
+        a temporary path.
+
+    kwargs are passed to [`random_spool`](`dascore.examples.random_spool`)
+    """
+    spool = random_spool(**kwargs)
+    path = spool_to_directory(spool, path)
+    return dc.spool(path)
+
+
 @register_func(EXAMPLE_SPOOLS, key="diverse_das")
 def diverse_spool():
     """
--- a/dascore/examples.py
+++ b/dascore/examples.py
@@ -18,7 +18,7 @@ from dascore.exceptions import UnknownExampleError
 from dascore.utils.docs import compose_docstring
 from dascore.utils.downloader import fetch
 from dascore.utils.misc import register_func
-from dascore.utils.patch import get_patch_name
+from dascore.utils.patch import get_patch_names
 from dascore.utils.time import to_timedelta64
 
 EXAMPLE_PATCHES = {}
@@ -506,7 +506,7 @@ def spool_to_directory(spool, path=None, file_format="DASDAE", extention="hdf5")
         path = Path(tempfile.mkdtemp())
         assert path.exists()
     for patch in spool:
-        name = get_patch_name(patch).iloc[0]
+        name = get_patch_names(patch).iloc[0]
         out_path = path / (f"{name}.{extention}")
         patch.io.write(out_path, file_format=file_format)
     return path
--- a/dascore/examples.py
+++ b/dascore/examples.py
@@ -451,7 +451,7 @@ def random_spool(time_gap=0, length=3, time_min=np.datetime64("2020-01-03"), **k
     return dc.spool(out)
 
 
-@register_func(EXAMPLE_SPOOLS, key="radom_directory_das")
+@register_func(EXAMPLE_SPOOLS, key="random_directory_das")
 def random_directory_spool(path=None, **kwargs):
     """
     Create a random spool, then save to specified path.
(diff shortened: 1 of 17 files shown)


d-chambers:
@ahmadtourei, mind taking this for a spin to see if it behaves how you wanted ?

dascore/utils/patch.py
```diff
@@ -446,21 +431,90 @@ def get_start_stop_step(patch: PatchType, dim):
...
+    output will be the file name with the extension removed. The path must
+    use / as a delinater.
 
-    def _format_datetime64(dt):
+    Examples
+    --------
+    >>> import dascore as dc
+    >>> from dascore.utils.patch import get_patch_names
+    >>> patch = dc.get_example_patch()
+    >>> name = get_patch_names(patch)
```
ahmadtourei, line 470: It would be helpful to add a few examples of using this function for directory and memory spools. Based on the current function's name, user may assume it can be only applied on patch.

ahmadtourei, line 470: `name` is:

```python
0    DAS_______random__2017_09_18__2017_09_18T00_00_07
Name: network, dtype: object
```

We get `DAS_______random__2017_09_18__2017_09_18T00_00_07` using `name[0]` because the function returns a `<class 'pandas.core.series.Series'>`. Is there a specific reason that it returns a pandas series instead of the name in a string?

dascore/utils/patch.py
```diff
@@ -446,21 +431,90 @@ def get_start_stop_step(patch: PatchType, dim):
...
 
 
-def get_default_patch_name(patch):
-    """Generates the name of the node."""
+def get_patch_names(
+    patch_data: pd.DataFrame | dc.Patch | dc.BaseSpool,
+    prefix="DAS",
+    attrs=("network", "station", "tag"),
+    coords=("time",),
+    sep="__",
```
ahmadtourei, line 439: need test(s) for optional inputs . For example, get_patch_names(patch, patch) does not raise an error.

tests/conftest.py
```diff
@@ -385,6 +385,15 @@ def one_file_dir(tmp_path_factory, random_patch):
     return ex.spool_to_directory(spool, path=out)
 
 
+@pytest.fixture(scope="session")
+def random_spool_directory(tmp_path_factory):
+    """A directory with a few patch files."""
+    out = Path(tmp_path_factory.mktemp("one_file_file_spool"))
+    spool = ex.get_example_spool("random_das")
+    out_path = ex.spool_to_directory(spool, path=out)
+    return dc.spool(out_path).update()
```
ahmadtourei, line 394: It would be nice to have this as an example in dascore.examples as all of the example spools are memory spools.

ahmadtourei (changes requested):
Looks good! I have posted my comments below. 

One general comment that I have is because this can be a very useful function, it would be great to create a shortcut for it (something like `patch.name` and `spool.name` so the user can quickly get name(s) in string format. Then we can include it in the Shortcuts section of `dascore/docs/tutorial/patch.qmd`. I assume users with limited DASCore experience cannot quickly locate it in `dascore.utils.patch`.

dascore/utils/patch.py
d-chambers: Yes, for the function to work with a spool it needs to return multiple names. I avoided returning multiple types (based on if the input is a Patch or not) from `get_patch_names` as that is a bit of an anti-pattern. 

I understand the confusion though. I will add `get_patch_name` as a method of `Patch` that will do this. I think with the different name (`get_patch_name` vs `get_patch_names`) it will be clear.

d-chambers:
>  I assume users with limited DASCore experience cannot quickly locate it in dascore.utils.patch.

I agree. I think it should be a patch/spool method. Because there are optional parameters though, probably best not to make it an attribute (because then the users are always stuck with the defaults).

d-chambers:
Thanks @ahmadtourei for the great review. I have addressed all of your suggestions/concerns. Anything else you want to see in here? I also included a fix for the doc build so I am keen to merge this soon to see if the fix works.

dascore/examples.py
```diff
@@ -451,6 +451,24 @@ def random_spool(time_gap=0, length=3, time_min=np.datetime64("2020-01-03"), **k
     return dc.spool(out)
 
 
+@register_func(EXAMPLE_SPOOLS, key="radom_directory_das")
```
ahmadtourei, line 454: *random

ahmadtourei: Also, currently getting the following error trying to getting the spool:
```python
import dascore as dc
sp  = dc.get_example_spool("radom_directory_das")

AttributeError: module 'dascore.io.segy.core' has no attribute 'SegyV2'. Did you mean: 'SegyV2_0'?
```
I see you've already merged master so I'm unsure what the problem is.

docs/tutorial/patch.qmd
```diff
@@ -227,6 +227,18 @@ print(patch.channel_count) # to get the number of channels in the patch.
...
+
+```python 
+import dascore as dc
+
+patch = dc.get_example_patch()
+spool = dc.get_example_spool()
+
+patch_name = patch.get_patch_name()
+patch_names = spool.get_patch_names()
+```
```
ahmadtourei, line 240: great!

ahmadtourei (approved):

ahmadtourei:
> Thanks @ahmadtourei for the great review. I have addressed all of your suggestions/concerns. Anything else you want to see in here? I also included a fix for the doc build so I am keen to merge this soon to see if the fix works.

Thanks, Derrick, for your work on this PR! It looks great. The only new comment I have is a typo I listed below.

Issue #512, opened 2025-08-18
Patch.viz.waterfall() renders different plots across Matplotlib versions


ahmadtourei:
## Description

`Patch.viz.waterfall()` produces noticeably different plots with apparent saturation/contrast of traces when using matplotlib==3.9.2 versus the newer Matplotlib that gets installed alongside DASCore. 

#### matplotlib==3.9.2 
`pa.viz.waterfall(scale=0.0001)`
<img width="619" height="448" alt="Image" src="https://github.com/user-attachments/assets/84c9faf6-7ed6-46d2-8d87-5b2f23a21524" />

#### matplotlib==3.10.5
`pa.viz.waterfall(scale=0.0001)`
<img width="619" height="448" alt="Image" src="https://github.com/user-attachments/assets/8b9bbd6f-635a-4f8d-bdef-f77d83ab5eb7" />

`pa.viz.waterfall(scale=0.01)`
<img width="623" height="448" alt="Image" src="https://github.com/user-attachments/assets/f63c92ae-4e0d-4481-a706-e54955fc7e07" />

## Versions
 - OS [e.g. Ubuntu 20.04]: MacOS 15.6 
 - DASCore Version [e.g. 0.0.5]: 0.1.9
 - Python Version [e.g. 3.10]: 3.12

d-chambers:
Hey @ahmadtourei, have you had any thoughts on how we should go about addressing this?

d-chambers:
Tracing the history here, because the fix went in and then partly came back out.

- #574 (waterfall refactor) hard-pinned \`interpolation_stage="data"\` in the \`imshow\` call, with the comment *"these parameters are so that matplotlib versions > 3.10 behave like matplotlib < 3.9, which tends to work better for visualizing big DAS data. See #512."* That was the fix.
- #707 then made it a user-facing parameter and changed the default to \`"auto"\`, and raised the minimum matplotlib to 3.10.

\`"auto"\` is matplotlib 3.10's own default, and downsampling under \`"auto"\` resolves to \`"rgba"\` — which is exactly the behavior that produced the second screenshot above. So on current \`dev\`, the default rendering is the 3.10 one, and the pre-3.9 look is available on request:

```python
patch.viz.waterfall(scale=0.0001, interpolation_stage="data")
```

[...]

PR #574, opened 2025-10-16
Waterfall refactor

d-chambers:
## Description

This PR refactors `viz.waterfall` function. It does a few things:

1. Fixes typos, clarifies the docs, and some minor refactoring of internal functions.

2. Changes default behavior when scale=None to perform the standard statistical fence to select ranges rather than using the whole data range. This will eliminate, or at least alleviate, the need for constant fiddling with the scale parameter when there are a few outliers in the data (very common). The old behavior can re-created using scale=1.0. 

3. Replaces the assertions with proper exception raises for malformed parameters.

4. Makes the automatic Y axis flip only occur when the Y axis of the patch is time-like. This will still match seismic convention but stop flipping with other types of coordinates. This can always be manipulated directly with the returned matplotlib axis. 

## Summary by CodeRabbit

* **New Features**
  * Waterfall visualization now defaults to an IQR-based (1.5×IQR) outlier-resistant color scale and supports relative/absolute scaling options.
  * Utility to convert percentage values to fractional form added.

* **Improvements**
  * Stronger input validation with clearer errors for invalid scales/scale types and non-2D data; improved axis/time labeling, colorbar handling, and explicit show/return behavior.

* **Documentation**
  * Examples updated to showcase default IQR scaling and explicit scaling modes.

* **Tests**
  * New tests for scale validation, non-2D inputs, constant-data handling, and percent conversion.

* **Chores**
  * Raised minimum matplotlib requirement.

### DIFF
--- a/dascore/proc/taper.py
+++ b/dascore/proc/taper.py
@@ -277,7 +277,6 @@ def taper_range(
     >>> # Apply two non-overlapping tapers
     >>> taper_range = ((25,50,100,125), (150,175,200,225))
     >>> patch_tapered_5 = patch.taper_range(distance=taper_range)
-
     """
     dim, ax, values = get_dim_axis_value(patch, kwargs=kwargs)[0]
     coord = patch.get_coord(dim, require_sorted=True)
--- a/dascore/units.py
+++ b/dascore/units.py
@@ -14,7 +14,7 @@ from pint import DimensionalityError, Quantity, UndefinedUnitError, Unit
 import dascore as dc
 from dascore.compat import is_array
 from dascore.exceptions import UnitError
-from dascore.utils.misc import unbyte
+from dascore.utils.misc import iterate, unbyte
 from dascore.utils.time import dtype_time_like, is_datetime64, is_timedelta64, to_float
 
 str_or_none = TypeVar("str_or_none", None, str)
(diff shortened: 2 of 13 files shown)


ahmadtourei:
@d-chambers 
I tested it on three DAS datasets, and they all look good. However, #512 caught my attention again as different matplotlib versions result in different saturation for all three datasets I tested. I also realized that the `waterfall()` function is much slower with `matplotlib==3.10.7` compared to `matplotlib==3.9.2` (~2.5 minutes on my Mac for a 2GB patch compared to less than 10 sec with the older version).

d-chambers:
> However, https://github.com/DASDAE/dascore/issues/512 caught my attention again

Ya, that is unfortunate, but I am still not sure how we can handle it besides, perhaps raising an issue with matplotlib. Have you had any other thoughts? For the record, I haven't run into it, but I have been looking at minutes of data with sample rates up to 500 Hz lately, so maybe to small to show the effect you document?

d-chambers:
I dug into this a bit more, it appears that there where some changes in data interpolation from 3.9 to 3.10 you can see [here](https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.10.0.html?utm_source=chatgpt.com#imshow-interpolation-stage-default-changed-to-auto). So, if we simply set these parameters to the old defaults in `Patch.waterfall`'s `imshow` call we *should* get back to the old behavior.

PR #674, opened 2026-05-18
new patch functions for demean and demedian

andreas-wuestefeld:
Two very simple patch-functions: demean and demedian. They remove statics.
detrend( "linear") also removed the mean, but having a dedicated function is perhaps more intuitive.

Note that we found that demedian often works better than demean to remove common-mode noise, i.e. applying the function to the distance axis

Checklist
I have (if applicable):

 referenced the GitHub issue this PR closes.
[ X] documented the new feature with docstrings and/or appropriate doc page.
[ X] included tests. See [testing guidelines](https://dascore.org/contributing/testing.html).
[ X] added the "ready_for_review" tag once the PR is ready to be reviewed.

### DIFF
--- a/dascore/core/patch.py
+++ b/dascore/core/patch.py
@@ -404,6 +404,8 @@ class Patch(NamespaceOwner):
     correlate = dascore.proc.correlate
     correlate_shift = dascore.proc.correlate_shift
     decimate = dascore.proc.decimate
+    demean = dascore.proc.demean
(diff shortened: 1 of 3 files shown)


d-chambers:
For the doctests you might need to suppress output from some lines you aren't expecting a value from.  Eg 

```python
ax.set_title('title');
# Or
_ = ax.set_title('title')
```

Otherwise these return a text object that can fail the doctest.

PR #681, opened 2026-05-21
new transforms: STALTA, FBE, Kurtosis

andreas-wuestefeld:
I propose here to add a few new data transforms.

1) Frequency-Band Energy is de-facto standard to visualise DAS data ("the waterfall plot"). This implementation is based purely on DAScore functionality, only wrapping it up in a dedicated function. Note that this is not tested for data-stream. If you have suggestions how to do that, I that would be appreciated.

2) Short-Term Average / Long-Term Average (STALTA) is a function commonly used in geophysics for onset detection. This implementation is based purely on DAScore functionality, only wrapping it up in a dedicated function.

3) Kurtosis is a powerful onset detector transform with only a single parameter. 

I hope these will be useful for the community

## Checklist

I have (if applicable):

## Summary by CodeRabbit

* **New Features**
  * Added Frequency‑Band Energy (FBE), Kurtosis, and STA/LTA signal transforms and exposed them as patch methods; rolling/window helpers now support overlap and forwarded arguments.

* **Bug Fixes**
  * Colorbar units now render with square brackets; improved error message spacing.

* **Documentation**
  * Added bibliographic reference and revamped release/publishing and contributor workflow docs.

* **Tests**
  * Comprehensive tests added/expanded for new transforms and rolling/window behavior.

* **Chores**
  * CI workflow trigger and .gitignore updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DASDAE/dascore/pull/681?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/dascore/transform/fbe.py
+++ b/dascore/transform/fbe.py
@@ -1,0 +1,80 @@
+"""
+Patch function for 'Frequency-Band Energy' transform
+"""
+
+from __future__ import annotations
+
+from dascore.constants import PatchType
+from dascore.utils.patch import patch_function
+from dascore.utils.time import to_float
+
+
+@patch_function()
+def fbe(
+    patch: PatchType,
+    corners: tuple[float, float] = (None, None),
+    time: float | None = None,
+    step: float | None = None,
+    db: bool = True,
+) -> PatchType:
+    """
+    Compute the rolling Root-Mean-Squared (RMS) of the Energy in a
+    Frequency Band (i.e  the Frequency-Band Energy, or FBE).
+    This is commonly called 'the waterfall plot' in DAS-processing.
+    This implementation is a wrapper to DAScore functionality:
+        1) Apply a 'pass_filter' to the patch
+        2) Apply rolling-funtction of window length "time"
+        3) Calculate RMS value for each channel
+
+
+    Parameters
+    ----------
+    patch
+        Input DASCore patch.
+    corners
+        Two-element tuple with frequencies to calculate energy in.
+    time
+        window length in which to caluclate engergy.
+    step
+        time-step for rolling window. Defaults to sampling rate. This can be
+        used for coarser sampling of the resulting patch
+    db
+        Return patch data in decibel [dB] instead of orginal units
+
+    Returns
+    -------
+    PatchType
+        A new patch containing FBE-RMS traces.
+
+    Example
+    --------
+    >>> import dascore as dc
+    >>>
+    >>> p = dc.examples.get_example_patch('example_event_2')
+    >>>
+    >>> fbe = p.fbe(corners = (100, 200), time = 0.002)
+    >>> ax = fbe.viz.waterfall(cmap = 'Spectral_r')
+    """
+    if step is None:
+        step = to_float(patch.get_coord("time").step)
+
+    if time is None:
+        time = 20 * step
+
+    if any(x is not None for x in corners):
+        if not (corners[1] > corners[0]):
+            raise ValueError(
+                "The second frequency corner must be larger than the first."
+            )
+        patch = patch.pass_filter(time=corners)
+
+    fbe = ((patch**2).rolling(time=time, step=step).mean() ** 0.5).update(
+        attrs={"data_type": "Frequency-Band Energy"}
+    )
+
+    if db:
+        fbe = (10 * fbe.log10()).update(
+            attrs={"data_type": "Frequency-Band Energy", "data_units": "dB"}
+        )
+
+    return fbe
--- a/dascore/transform/fbe.py
+++ b/dascore/transform/fbe.py
@@ -61,12 +61,20 @@ def fbe(
     if time is None:
         time = 20 * step
 
-    if any(x is not None for x in corners):
-        if not (corners[1] > corners[0]):
+    if len(corners) != 2:
+        raise ValueError(
+            "Corners must be a two-element tuple, with (low, high) frequency corners"
+        )
+
+    if corners != (None, None):
+        low, high = corners
+
+        if low is not None and high is not None and high <= low:
             raise ValueError(
                 "The second frequency corner must be larger than the first."
             )
-        patch = patch.pass_filter(time=corners)
+
+        patch = patch.pass_filter(time=(low, high))
 
     fbe = ((patch**2).rolling(time=time, step=step).mean() ** 0.5).update(
         attrs={"data_type": "Frequency-Band Energy"}
--- a/dascore/transform/fbe.py
+++ b/dascore/transform/fbe.py
@@ -12,7 +12,8 @@ from dascore.utils.time import to_float
 @patch_function()
 def fbe(
     patch: PatchType,
-    corners: tuple[float, float] = (None, None),
+    freqmin: float | None = None,
+    freqmax: float | None = None,
     time: float | None = None,
     step: float | None = None,
     db: bool = True,
@@ -31,8 +32,12 @@ def fbe(
     ----------
     patch
         Input DASCore patch.
-    corners
-        Two-element tuple with frequencies to calculate energy in.
+    freqmin
+        Lower-end of the frequency band to calculate energy in.
+        If set to None a low-pass filter is applied
+    freqmax
+        Upper-end of the frequency band to calculate energy in.
+        If set to None a high-pass filter is applied
     time
         window length in which to caluclate engergy.
     step
@@ -61,20 +66,12 @@ def fbe(
     if time is None:
         time = 20 * step
 
-    if len(corners) != 2:
-        raise ValueError(
-            "Corners must be a two-element tuple, with (low, high) frequency corners"
-        )
-
+    corners = (freqmin, freqmax)
     if corners != (None, None):
-        low, high = corners
+        if freqmin is not None and freqmax is not None and freqmax <= freqmin:
+            raise ValueError("freqmax must be larger than freqmin.")
 
-        if low is not None and high is not None and high <= low:
-            raise ValueError(
-                "The second frequency corner must be larger than the first."
-            )
-
-        patch = patch.pass_filter(time=(low, high))
+        patch = patch.pass_filter(time=corners)
 
     fbe = ((patch**2).rolling(time=time, step=step).mean() ** 0.5).update(
         attrs={"data_type": "Frequency-Band Energy"}
--- a/dascore/transform/fbe.py
+++ b/dascore/transform/fbe.py
@@ -20,13 +20,13 @@ def fbe(
     **kwargs,
 ) -> PatchType:
     """
-    Compute the rolling Root-Mean-Squared (RMS) of the Energy in a
-    Frequency Band (i.e  the Frequency-Band Energy, or FBE).
-    This is commonly called 'the waterfall plot' in DAS-processing.
+    Compute the rolling Frequency Band Energy in a window.
+    This is the Root-Mean-Squared (RMS) of the Energy in a Frequency Band (the FBE),
+    and commonly called 'the waterfall plot' in DAS-processing.
     This implementation is a wrapper to DAScore functionality:
         1) Apply a 'pass_filter' to the patch
-        2) Apply rolling-funtction of window length "time"
-        3) Calculate RMS value for each channel
+        2) Apply rolling-funtction of a window along a coordinate
+        3) Calculate RMS value
 
 
     Parameters
--- a/dascore/transform/fbe.py
+++ b/dascore/transform/fbe.py
@@ -57,7 +57,7 @@ def fbe(
     >>>
     >>> p = dc.examples.get_example_patch('example_event_2')
     >>>
-    >>> fbe = p.fbe(corners = (100, 200), time = 0.002)
+    >>> fbe = p.fbe(freqmin=100, freqmax=200, time = 0.002)
     >>> ax = fbe.viz.waterfall(cmap = 'Spectral_r')
     """
     if step is None:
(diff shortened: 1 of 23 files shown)


dascore/transform/fbe.py
```diff
@@ -0,0 +1,85 @@
...
+
+@patch_function()
+def fbe(
+    patch: PatchType,
+    freqmin: float | None = None,
+    freqmax: float | None = None,
+    time: float | None = None,
+    step: float | None = None,
+    db: bool = True,
+) -> PatchType:
```
d-chambers, line 21: I would suggest a slightly different API to align better with DASCore's ethos, mainly of being dimension-agnostic.

 Consider how `Patch.pass_filter` uses kwargs to indicate the dimension and filter bounds `Patch.pass_filter(time=(freqmin, freqmax))`. Here we might just do the same, then we can pass the kwargs directly to `Patch.pass_filter` without having to do any validations (we just let that function handle it). 

Then, rather than `time` we could use `window`. Also, would `step` or `overlap` be more natural? I have been meaning to add an `overlap` parameter to `rolling` for awhile, which is mutually exclusive to `step`.  I added that (#686) so you can use it if you think it makes sense here. I would think a default value of `overlap=50*dc.units.percentage` would make sense for what people would normally expect from an FBE no? 

So, something like:

```python
from dascore import get_example_patch
from dascore.units import s, percent

patch.fbe(time=(30, 100), window=5, overlap=2)
# or
patch.fbe(time=(100, ...), window=10*s, overlap=50*percent)
```

`dascore.utils.patch.get_dim_axis_value` will be helpful here to extract the dimension name from the kwargs, so you can pass it to rolling as **{dim: window, overlap=overlap}.

dascore/transform/stalta.py
```diff
@@ -0,0 +1,64 @@
...
+
+from __future__ import annotations
+
+from dascore.constants import PatchType
+from dascore.utils.patch import patch_function
+from dascore.utils.time import to_float
+
+
+@patch_function()
+def stalta(
```
d-chambers, line 13: Same comment here about DASCore's use of kwargs. Maybe a better API would be:

`Patch.stalta(time=(sta, lta)` with no default values because these can be really dataset dependent. 

Also, I have mixed feeling about the `absolute` parameter. In general, most signals are either zero-meaned (eg after de-trending), in which case absolute is a must, or have large offsets in which case meaningful differences in onset signal may not be captured anyway if the offset is large. 

So, I wonder if we ought to remove this parameter and either 1) automatically do abs (but this might be a wasted computation if it was already done), or 2) dont, and just state it is usually needed before calling this function in the docstring and examples.

dascore/transform/fbe.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    Compute the rolling Root-Mean-Squared (RMS) of the Energy in a
+    Frequency Band (i.e  the Frequency-Band Energy, or FBE).
+    This is commonly called 'the waterfall plot' in DAS-processing.
+    This implementation is a wrapper to DAScore functionality:
+        1) Apply a 'pass_filter' to the patch
+        2) Apply rolling-funtction of window length "time"
+        3) Calculate RMS value for each channel
```
d-chambers: The first line of the docstring needs to be exactly one line (<85 chars) because it is used in a table to summarize this function. Additional details can go in the next paragraph. See [here](https://dascore.org/contributing/documentation.html#doc-strings) for more details.

dascore/transform/kurtosis.py
```diff
@@ -0,0 +1,234 @@
...
+
+@patch_function()
+def kurtosis(
+    patch: PatchType,
+    winlen: float,
+    dim: str = "time",
+    recursive: bool = True,
+) -> PatchType:
+    """
+    Compute kurtosis along a patch dimension. Note that for best results
```
d-chambers: Same comment here, first line needs to be a single line followed by a space.

dascore/transform/kurtosis.py
```diff
@@ -0,0 +1,234 @@
...
+        out = np.empty_like(data, dtype=float)
+
+        # Per-channel initialization
+        varx = np.std(data, axis=0) ** 2  # the version on Maggi uses the std not var
+        mean_value = np.zeros(data.shape[1:], dtype=float)
+        var_value = np.zeros(data.shape[1:], dtype=float)
+        kurt_value = np.zeros(data.shape[1:], dtype=float)
+
+        varx2 = varx**2
+        for i in range(npts):
```
d-chambers: Performance implications for this loop? Seems it might get quite slow over a large time axis no? you could move this outside of the kurtosis function, then wrap in a `dascore.utils.jit.maybe_numba_jit` to automatically compile if numba is installed. Same for any of the other sub-functions that might benefit.

dascore/transform/kurtosis.py
```diff
@@ -0,0 +1,234 @@
...
+        Window length in seconds.
+    dim
+        Dimension along which to compute kurtosis. Defaults to ``"time"``.
+    recursive
+        If True, use recursive pseudo-kurtosis: Instead of computing kurtosis
+        in a sliding window (computationally expensive for continuous data),
+        the @langet2014 propose a recursive formulation. This acts like an
+        exponentially weighted moving estimator, so the algorithm updates
+        continuously without storing long windows of data.
+        If False, the common kurtosis calculation is used
```
d-chambers: Perhaps point out how the decay exponent is still based on the selected window length? Otherwise it is not clear how these parameters relate when `recursive=True`.

d-chambers: Also, perhaps useful to give some guidance here on when to use recursive vs not?

d-chambers (changes requested):
Some suggestions to tweak the API and a few docstrings.

dascore/transform/fbe.py
andreas-wuestefeld: Thanks for the suggestions. It makes sense to keep the DAScore style, and your suggestions are indeed helpful there. I have not seen any FBE on other axis then time, so that could be interesting.
Regarding overlap, I am reluctant. Could you explain more why that is a beneficial/superior parameter? Step is an established parameter, and everyone will understand that this is the desired sampling rate. Overlap seems a bit ambiguous and prone to trial and error to get what you want.

dascore/transform/stalta.py
andreas-wuestefeld: agreed on the time=(sta,lta). It seemed more natural to have it that way, but your suggestion in more DAScore-consistent. It should also be possible to do it along other axis
agreed on the absolute parameter. Will remove it and opt for suggestion 2)

dascore/transform/kurtosis.py
andreas-wuestefeld: I have no experience with numba. 
I have used the kurtosis for largish arrays, but no feeling if it could be improved.
I tried vectorizing the loop, but that was indeed slower.

dascore/transform/fbe.py
d-chambers: Hmm... yes good point about step. I mostly had thought about it in terms of window overlap because it qualitatively relates to smoothing. But, if you are doing more than just visualizing the result right away i can see why you would want more control over the output sampling rate. 

So, I suggest you either just keep step or support step and overlap. Passing both arguments to patch.rolling should just work.

d-chambers: You will have to pull/rebase to current master first though.

andreas-wuestefeld:
I implemented the suggested changes to FBE, but not the overlap yet.
One can indeed to an FBE along distance axis. Note sure what it tells us though :-)
I guess energy in a wavenumber range...

```
 import dascore as dc

p = dc.examples.get_example_patch('example_event_2')

fbe_patch = p.fbe(distance=(.01,.05), window = 5)
ax = fbe_patch.viz.waterfall(cmap = 'Spectral_r')
_ = ax.set_title('FBE along distance-axis')
```
image

d-chambers:
> One can indeed to an FBE along distance axis. Note sure what it tells us though :-)

Wow, it looks quite cool whatever it is. Nice cmap selection.

andreas-wuestefeld:
@d-chambers the numba-jit functions in kurtosis return fails during test by CODECOV
Locally the tests seem to work.

Is there a special decorator to be used in tests to tell him to not try and use numba?

d-chambers:
Hey @andreas-wuestefeld,

Nice work on the numba stuff. I should have warned you about the test coverage issue when the python code gets compiled (codecov doesn't see that it gets run, as you noticed). The solution is usually to use the `.func` attr (which preserves the original function) to run a few smoke tests. So, if you update your test cases to something like this: 

```python
class TestKurtosisHelpers:
    """Tests for kurtosis helper functions."""

    def test_validate_window_returns_sample_count(self):
        """Ensure window length converts to sample count."""
        assert _validate_window(winlen=0.02, step=0.01) == 2

    def test_validate_window_rejects_non_positive_winlen(self):
        """Ensure non-positive window lengths raise."""
        with pytest.raises(ValueError, match="winlen must be positive"):
            _validate_window(winlen=0, step=0.01)

    def test_validate_window_rejects_too_short_window(self):
        """Ensure windows shorter than two samples raise."""
        with pytest.raises(ValueError, match="too small"):
            _validate_window(winlen=0.005, step=0.01)

    def test_moving_sum_values(self):
        """Ensure moving sum uses clipped centered windows."""
        data = np.arange(5, dtype=float).reshape(5, 1)

        out, counts = _moving_sum.func(data, nwin=3)

        assert np.allclose(out[:, 0], [1, 3, 6, 9, 7])
        assert np.allclose(counts, [2, 3, 3, 3, 2])

    def test_windowed_kurtosis_constant_data_returns_nan(self):
        """Ensure zero-variance data returns NaNs."""
        data = np.ones((10, 3), dtype=float)

        out = _windowed_kurtosis.func(data, nwin=3)

        assert np.isnan(out).all()

    def test_windowed_kurtosis_non_constant_data(self):
        """Ensure positive-variance windows return finite values."""
        rng = np.random.default_rng(42)
        data = rng.normal(size=(10, 3))

        out = _windowed_kurtosis.func(data, nwin=3)

        assert out.shape == data.shape
        assert np.isfinite(out).any()

    def test_recursive_kurtosis_preserves_shape(self):
        """Ensure recursive helper preserves input shape."""
        rng = np.random.default_rng(42)
        data = rng.normal(size=(20, 3))
        varx = np.var(data, axis=0)

        out = _recursive_kurtosis.func(data, step=0.01, winlen=0.05, varx=varx)

        assert out.shape == data.shape

    def test_recursive_kurtosis_constant_data_returns_nan(self):
        """Ensure recursive helper handles zero-variance data."""
        data = np.zeros((20, 3), dtype=float)

        out = _recursive_kurtosis.func(
            data,
            step=0.01,
            winlen=0.05,
            varx=np.zeros(data.shape[1]),
        )

        assert np.isnan(out).all()
```
 
It hits all the lines.  I had codex whip up a [quick profile scipt](https://github.com/user-attachments/files/28176084/time_kurtosis_numba.py) and the results look quite promising so far. 

| Data shape | Method | Python | Numba | Speedup |
|---:|---|---:|---:|---:|
| 100 x 20 | `patch.kurtosis(recursive=True)` | 3.734 ms | 1.314 ms | 2.84x |
| 100 x 20 | `patch.kurtosis(recursive=False)` | 18.510 ms | 1.377 ms | 13.45x |
| 300 x 50 | `patch.kurtosis(recursive=True)` | 19.610 ms | 1.310 ms | 14.97x |
| 300 x 50 | `patch.kurtosis(recursive=False)` | 123.475 ms | 1.766 ms | 69.93x |
| 1000 x 100 | `patch.kurtosis(recursive=True)` | 126.057 ms | 1.469 ms | 85.83x |
| 1000 x 100 | `patch.kurtosis(recursive=False)` | 824.250 ms | 5.798 ms | 142.16x |

Run config: `winlen=0.1`, `step=0.004`, `window_samples=25`, `repeats=5`.


In general, whenever you have a function with loops that takes pure numpy arrays and basic numbers `maybe_numba_jit` is reasonable to look at.

d-chambers (approved):
Looks good after API adjustments, just make sure the coverage is back up before merge. 

You can add the `overlap` parameter to `fbe`, or if you prefer not to, we can leave it for consideration later. 

Currently ['stft'](https://github.com/DASDAE/dascore/blob/689826aed9b7447fdb33360d50ab0cbceac8c046/dascore/transform/fourier.py#L376-L383)  only has `overlap` and not `step`. You are certainly right that `step` is useful, so I think we should consider adding it later.  It would be reasonable to unify all the windowing functions in a later PR by carefully considering their APIs and recording the guidance in the developer docs.

andreas-wuestefeld:
Wow, that is indeed a massive speedup!
I can't take all the credit for implementing it. ChatGPT was doing most of the hard work :-)

d-chambers:
Also, I am not exactly sure why github thinks you changed so many files wrt master. I think it might be because you pulled dasdae:master into your branch (as opposed to rebase) but there are also some duplicated looking commits in here too, so I am not exactly sure what happened with the git history. Either way, just make sure to squash and merge and it should clean it up fine.

andreas-wuestefeld:
@d-chambers Yes, sorry for the mess-up. I will move my git-origin to the main repository once all current branches are merged. I don't want to mess things up even more, but all new activities will be then without my fork (hopefully)

andreas-wuestefeld:
I will also implement the overlap to the FBE and STALTA
I feel it's better to start a new PR and feature upgrade and not mess this one up even more

PR #685, opened 2026-05-21
Feature/dominant frequency

andreas-wuestefeld:
Some more transformations that have been useful for me in the past:
mean_frequency and median_frequency apply a rolling window and calculate a measure of the "dominant frequency"

Sometimes mean_frequency shows better what you expect from your data, sometimes median_frequency. Here are options for both.

Note that these functions make use of the "partial" function on the rolling_function. 
this can be updated once the rolling_function allows for arguments directly

## Checklist

I have (if applicable):

## Summary by CodeRabbit

* **New Features**
  * Patch objects gain mean_frequency and median_frequency transforms for rolling-window spectral analysis with configurable window/step, optional fmin/fmax, and selectable PSD method; outputs include updated data type and units (Hz).

* **Tests**
  * Added tests for PSD/FFT behavior, helper routines, frequency-bounds validation, output attributes, numerical stability, and axis/shape handling for the new transforms.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DASDAE/dascore/pull/685?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/dascore/transform/mean_median_frequency.py
+++ b/dascore/transform/mean_median_frequency.py
@@ -257,7 +257,7 @@ def mean_frequency(
     >>>
     >>> patch = dc.get_example_patch('example_event_2')
     >>>
-    >>> fig, axs = plt.subplots(2,1, layout='constrained', figsize=(8,10))
+    >>> fig, axs = plt.subplots(1,2, layout='constrained', figsize=(12,4))
     >>> ax = patch.viz.waterfall(cmap='seismic', ax=axs[0])
     >>>
     >>> para = {'cmap':'turbo', 'scale':(50,300), 'scale_type':'absolute' }
@@ -285,9 +285,15 @@ def mean_frequency(
         frq, pxx = _get_psd_in_window(
             data, dt=dt, method=method, fmin=fmin, fmax=fmax, nperseg=nperseg
         )
-
         # calculate mean frequency
+        """
         mfr = np.sum(frq * pxx, axis=-1) / np.sum(pxx, axis=-1)
+        """
+        numerator = np.sum(frq * pxx, axis=-1)
+        denominator = np.nansum(pxx, axis=-1)
+        mfr = np.divide(
+            numerator, denominator, out=np.zeros_like(numerator), where=denominator != 0
+        )
 
         return mfr
 
@@ -377,7 +383,7 @@ def median_frequency(
     >>>
     >>> patch = dc.get_example_patch('example_event_2')
     >>>
-    >>> fig, axs = plt.subplots(2,1, layout='constrained', figsize=(8,10))
+    >>> fig, axs = plt.subplots(1,2, layout='constrained', figsize=(12,4))
     >>> ax = patch.viz.waterfall(cmap='seismic', ax=axs[0])
     >>>
     >>> para = {'cmap':'turbo', 'scale':(50,300), 'scale_type':'absolute' }
--- a/dascore/transform/mean_median_frequency.py
+++ b/dascore/transform/mean_median_frequency.py
@@ -374,7 +374,7 @@ def median_frequency(
     >>> ax = patch.viz.waterfall(cmap='seismic', ax=axs[0])
     >>>
     >>> para = {'cmap':'turbo', 'scale':(50,300), 'scale_type':'absolute' }
-    >>> med = patch.mean_frequency(winlen=.01, step=.001, fmin=50, fmax=300)
+    >>> med = patch.median_frequency(winlen=.01, step=.001, fmin=50, fmax=300)
     >>> ax = med.viz.waterfall(**para, ax=axs[1] )
     """
     dt = patch.get_coord("time").step
--- a/dascore/transform/mean_median_frequency.py
+++ b/dascore/transform/mean_median_frequency.py
@@ -14,6 +14,7 @@ from scipy.signal import get_window
 
 import dascore as dc
 from dascore.constants import PatchType
+from dascore.utils.patch import patch_function
 
 
 def _welch(xx, win, fs=1.0, nperseg=256):
@@ -204,7 +205,7 @@ def _get_psd_in_window(data, method="WELCH", dt=1, fmin=None, fmax=None, nperseg
 
 
 # %%
-@dc.patch_function(required_dims=("time",), history="full")
+@patch_function(required_dims=("time",), history="full")
 def mean_frequency(
     patch: PatchType,
     winlen: float | None = None,
@@ -244,11 +245,16 @@ def mean_frequency(
     method
         Method for calculating spectrum.
         Can be any of "welch" (default), "fft", or "dft"
-    Returns
 
+    Returns
+    -------
+        The Patch instance with the mean-frequency as data.
 
     Example
-    ----------
+    -------
+    >>> import dascore as dc
+    >>> import matplotlib.pyplot as plt
+    >>>
     >>> patch = dc.get_example_patch('example_event_2')
     >>>
     >>> fig, axs = plt.subplots(2,1, layout='constrained', figsize=(8,10))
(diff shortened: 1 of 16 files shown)


dascore/transform/mean_median_frequency.py
```diff
@@ -0,0 +1,418 @@
+"""Rolling mean and median frequency transforms for DASCore patches"""
```
d-chambers, line 1: Maybe some `maybe_numba_jit` would be appropriate here too?

dascore/transform/mean_median_frequency.py
```diff
@@ -0,0 +1,418 @@
...
+
+    # now apply the rolling function; make sure time axis is last dimension
+    dim = "time"
+    mfr_patch = (
+        patch.transpose(..., dim)
+        .rolling(time=winlen, step=step, samples=False)
+        .apply(fun)
+    )
+    if patch.get_axis("distance") != mfr_patch.get_axis("distance"):
+        mfr_patch = mfr_patch.transpose()
```
d-chambers, line 309: This is problematic because you didn't declare distance was required. You might have `time` and `depth` for example, then this would fail.

A better approach is to not depend on `distance` at all. 

```python  
original_dims = patch.dims
... # do stuff here
  out = out.transpose(*original_dims)
```

dascore/transform/mean_median_frequency.py
```diff
@@ -0,0 +1,418 @@
...
+    frq = frq[inf1:inf2]
+    return (
+        frq,
+        pxx,
+    )
+
+
+# %%
+@patch_function(required_dims=("time",), history="full")
+def mean_frequency(
```
d-chambers, line 206: You probably already know what I am going to say here, but...

I really don't like to restrict patch functions to a particular dimension when there isn't really a very good technical reason to do it. It is not just because there might be some theoretical (but bizarre) application to the distance axis. 

Consider, for example, that in a higher-dimensional patch, there might be multiple time dimensions (eg absolte_time, shot_time, relative_time, lag_time, etc.).  It could become vary awkward to use this function in such a case if the desired time axis isn't called exactly "time". 

So, my suggestion here is like with the fbe function: use kwargs to indicate window length. Then we know the window length and the dimension from one parameter call.

dascore/transform/mean_median_frequency.py
```diff
@@ -0,0 +1,418 @@
...
+    patch: PatchType,
+    winlen: float,
+    step: float | None = None,
+    fmin: float | None = None,
+    fmax: float | None = None,
+    method: str = "welch",
+    nperseg: int = 256,
```
d-chambers, line 213: This parameter is not documented in your docstring. Also, it has a very opaque name which probably corresponds to the scipy/numpy name, but I find it a bit unhelpful. I suggest naming int something like 
segment_samples, welsh_samples or spectral_window_samples.

dascore/transform/mean_median_frequency.py
```diff
@@ -0,0 +1,418 @@
...
+
+import numpy as np
+from scipy import fftpack
+from scipy.signal import get_window
+
+from dascore.constants import PatchType
+from dascore.utils.patch import patch_function
+
+
+def _welch(xx, win, fs=1.0, nperseg=256):
```
d-chambers, line 15: Two suggestions here. 

First, I know this is mainly adapted from scipy's function, but the variable naming in scipy is often less than clear. Perhaps we could give the public facing parameters some names that are a bit longer but more specific? 

Second, this function assumes a 2d array as `xx`.  We probably want to just refactor a bit to allow an array of any number of dimension (eg for higher order patches common in active source workflows). 

Here are codex's suggestions to make this work. If this is too onerous let's add an explicit check before the code gets here to indicate only 2D patches are accepted. 

 1. Stop indexing xx.shape[2].

 n_samples = xx.shape[-1]

  2. Flatten all leading dimensions before the segment loop:

  lead_shape = xx.shape[:-1]
  x_flat = xx.reshape(-1, n_samples)

  3. Allocate PSD using flattened trace count:

  n_traces = x_flat.shape[0]
  nfreq = nfft // 2 + 1
  psd = np.zeros((n_traces, nfreq), dtype=float)

  4. Remove the loop over xx.shape[1].

  Current code loops over the middle “rolling window” axis:

  for i in range(xx.shape[1]):
      x = xx[:, i, :]

  That is the 3D assumption. Instead, operate on all flattened traces together:

  x = x_flat

  5. Compute each Welch segment for all traces at once:

  for k, ind in enumerate(indices):
      segment = x[:, ind : ind + nperseg]
      segment = segment - np.mean(segment, axis=-1, keepdims=True)
      xft = fftpack.rfft(segment * win, nfft)
      ...

  6. Accumulate/average PSD across segments.

  You can keep the existing incremental average logic, but it should update pxx shaped (n_traces, nfreq) instead of (distance, nfreq).

  7. Keep separate even/odd FFT unpacking branches, but make them use x_flat.

  Even branch:

  pxx[:, (0, -1)] = xft[:, (0, -1)] ** 2
  pxx[:, 1:-1] = xft[:, 1:-1:2] ** 2 + xft[:, 2::2] ** 2

  Odd branch:

  pxx[:, 0] = xft[:, 0] ** 2
  pxx[:, 1:] = xft[:, 1::2] ** 2 + xft[:, 2::2] ** 2

  8. Apply scaling to pxx once after averaging.
  9. Reshape back to the original leading dimensions:

  psd = pxx.reshape(lead_shape + (pxx.shape[-1],))

  10. Build frequency axis from nfft:

  f = np.arange(pxx.shape[-1]) * (fs / nfft)

  So structurally _welch should go from (only works for (a, b, samples):

  psd = np.zeros((xx.shape[0], xx.shape[1], nfreq))
  for i in range(xx.shape[1]):
      x = xx[:, i, :]
      ...
      psd[:, i, :] = pxx

  to (works for (..., samples)):

  lead_shape = xx.shape[:-1]
  x = xx.reshape(-1, xx.shape[-1])
  pxx = compute_welch_for_2d_array(x)
  psd = pxx.reshape(lead_shape + (pxx.shape[-1],))

dascore/transform/mean_median_frequency.py
```diff
@@ -0,0 +1,418 @@
...
+    dim = "time"
+    mfr_patch = (
+        patch.transpose(..., dim)
+        .rolling(time=winlen, step=step, samples=False)
+        .apply(fun)
+    )
+    if patch.get_axis("distance") != mfr_patch.get_axis("distance"):
+        mfr_patch = mfr_patch.transpose()
+
+    return mfr_patch.update(attrs={"data_type": "Mean Frequency", "data_units": "Hz"})
```
d-chambers, line 311: If we no longer assume time index, data units should become inverse of current axis units. 

```python
coord = patch.get_coordinate(dim)
units = 1/dc.get_quantity(coord.units) if coord.units else None
```

dascore/transform/mean_median_frequency.py
```diff
@@ -0,0 +1,418 @@
...
+
+    if method.upper() == "WELCH":
+        frq, pxx = _welch(data, win, (1 / dt), nperseg=nperseg)
+
+    elif method.upper() == "FFT":
+        frq, pxx = _fft_psd(data, fs=(1 / dt))
+
+    # use only desired part of spectrum
+    inf1 = 0 if fmin is None else np.searchsorted(frq, fmin)
+    inf2 = -1 if fmax is None else np.searchsorted(frq, fmax)
```
d-chambers, line 194: As code rabbit caught, this drops the final frequency bin when fmax is None because it uses inf2 = -1. That omits the Nyquist bin from default calculations, biasing both mean and median frequency. You probably just want `None` as the slice stop when no upper bound is provided.

d-chambers (changes requested):
Main suggestions here:

1) Consider broadening the API to permit any dimension name to work, eg for higher order patches with multiple time-like dimensions. 
 
2) Decide if patches with 1 or 3+ dimensions are allowed. If so, rework the helper functions.  If not, raise a error early on.
 
3) Consider renaming some of the opaque numpy/scipy style variables to something more informative unless it becomes overly verbose. Particularly, the function parameters would benefit from clear names, even if internal variables keep the abbreviated form.

4) See if  `maybe_numba_jit` can come in handy here, especially for the `_welch` function.

andreas-wuestefeld:
I will refactor this completely, building on the stft-function. That will make everything more DAScore compatible, and easier to maintain

Issue #390, opened 2024-06-06
patch.dft() does not change data_type 


ahmadtourei:
## Description

`patch.dft()` does not change data_type attribute so when make waterfall plot from a patch with strain_rate data type (e.g., `patch.dft("time").abs().viz.waterfall()`), the colorbar has `strain_rate ` label.

## Versions
 - OS [e.g. Ubuntu 20.04]:
 - DasCore Version [e.g. 0.0.5]:
 - Python Version [e.g. 3.10]:

d-chambers:
hmmm.... should `Patch.dft` just remove the data_type tag? Or maybe it is still technically correct since it is derived from strain rate, its just not in strain rate units?

ahmadtourei:
Yeah, I'm not sure either. Keeping data_type as strain rate can be a bit misleading. Also, do we need it for `patch.idft()`? If not, removing data_type sounds okay to me. Or, maybe updating data_type to `amplitude` if `real=True` and `complex_number` if `real=False`? Then, we may make the `patch.abs()` update patch's data_type to `amplitude` if the original `data_type` is `complex`. What do you think?

PR #691, opened 2026-05-24
Feature/spectra

andreas-wuestefeld:
This PR include several feature additions and sligth behavioral changes. None is backward braking.
- Added patch-transform "patch.spectra". it builds on patch.dft to calculate the spectra, but allows to output 
  * amplitude spectra, power spectra, or power psectral density
  * optionally data can be converted to decibel
  * This is a separate function because in the future we might add an implementation to Welch's method, which can then be an option to choose from...
  
  @d-chambers : I would apprecieate if you had a look at the the conversions, if they are in line with the DAScore convention of dft. I checked and tested, but good if a second pair eyes have a look

- Added "patch.viz.specplot" to visualize spectra publication ready. 
  This is a wrapper around "patch.viz.waterfall". Feature are:
  * Option to show the ft_axis in log-scale (auto-detected whihc axis)
  * Default interpolation = bilinear
  * Default scale = (.5,1), since the lower energy end is ofter not of interest
  * Publication-readiness by automatically labelling axis "frequency" or "wavenumber", instead "ft_time" or "ft_distance"
  * format log-scale ticks in human-readable numbers (.1, 1, 10 ,100 instead of 10*-1, 10^0, 10^1, 10^2)
	 
  
- Modified dascore/viz/waterfall.py 
  * colorbar now automatically shows "extend" triangles if the color limits do not encompass the full data range
  * default colormaps depending on patch.datatype
	- Note: this may be put in a user-customizable .ini file, if there are needs for more such customisations
  * new boolean option "cbar" to indicate if a colorbar is to be shown or not (earlier behavior was to toggle with cmap=None)
  * added eps number to avoid division-by-zero warning when taking the log
  
- dascore/utils/plotting.py‎: _get_dim_label
  * original ``maybe_units = attrs.get(f"{dim}_units")```was not working for ft-axis. The more modern ``patch.get_coord(dim).units`` works
  * code converts units of 1/s to units of Hz, just for the labelling
  
‎- dascore/transform/fourier.py‎: _get_dft_attrs
  * issue #390 suggested to remove data_type of dft-transformed data. However this is useful for the automatic colormap selection. Instead of deleting, it is now set to "fourier transformed"
  
@d-chambers : Should we extent the _VALID_DATA_TYPES in /dascore/dascore/constants.py, and also include FBE, STALTA, KURTOSIS, FourierTransform

NOTES: 
I noticed there are no tests for dascore/utils/plotting.py. Is that by design?
example_event_2 has not datetime64 time-coordinate. Everything is float64. is it easy to update this in file on the server? or should it be done on the  dascore/examples.py script?

RESULT example:
Figure 2026-05-24 135447

## Checklist

I have (if applicable):

## Release Notes

* **New Features**
  * Added `spectra` patch transformation for computing amplitude, power, or power spectral density spectra with optional dB scaling along any dimension
  * Added `specplot` visualization for displaying spectral data with publication-ready axis labels
  * Enhanced waterfall plotting with intelligent colormap selection and new colorbar control

* **Bug Fixes**
  * Fixed Fourier transform metadata handling to preserve data type information
  * Improved dimension unit label rendering for frequency coordinates

* **Tests**
  * Added comprehensive test suites for spectra and specplot functionality

[Review Change Stack](https://app.coderabbit.ai/change-stack/DASDAE/dascore/pull/691?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/dascore/transform/spectra.py
+++ b/dascore/transform/spectra.py
@@ -63,7 +63,7 @@ def spectra(
     >>>
     >>> for a, dim in enumerate(["time", "distance"]):
     ...     spec = patch.spectra(dim=dim, kind="PSD", db=True)
-    ...     ax = spec.viz.spectraplot(
+    ...     ax = spec.viz.specplot(
     ...         log=True,
     ...         ax=axs[a + 1],
     ...         show=False,
@@ -87,7 +87,7 @@ def spectra(
     >>>
     >>> for a, dim in enumerate(["time", "distance"]):
     ...     spec = patch.spectra(dim=dim, kind="PSD", db=True)
-    ...     ax = spec.viz.spectraplot(
+    ...     ax = spec.viz.specplot(
     ...         log=True,
     ...         ax=axs[a + 1],
     ...         show=False,
--- a/dascore/transform/spectra.py
+++ b/dascore/transform/spectra.py
@@ -9,7 +9,7 @@ from dascore.constants import PatchType
 from dascore.utils.patch import _get_dx_or_spacing_and_axes, patch_function
 
 
-@patch_function(required_dims=("time",), history="full")
+@patch_function()
 def spectra(
     patch: PatchType,
     dim: str = "time",
@@ -132,8 +132,7 @@ def spectra(
         out = out.update(attrs={"data_type": "Power Spectral Density"})
 
     else:
-        raise ValueError("ERROR: Unknown option: kind=", kind)
-
+        raise ValueError(f"Unknown kind={kind!r}. Expected one of: 'AS', 'PS', 'PSD'.")
     if db:
         out = out.set_units(units.dB)
 
--- a/dascore/transform/spectra.py
+++ b/dascore/transform/spectra.py
@@ -51,39 +51,49 @@ def spectra(
 
     Example
     -------
+    Example 1: dispersion event
 
-    Example 2: dispersion event
     >>> import dascore as dc
     >>> import matplotlib.pyplot as plt
     >>>
-    >>> patch = dc.examples.dispersion_event().T.set_units(distance='m')
+    >>> patch = dc.examples.dispersion_event().T.set_units(distance="m")
     >>>
-    >>> fig, axs = plt.subplots(1,3,figsize=(16,4), layout='constrained')
-    >>> patch.viz.waterfall(cmap='RdBu_r', ax=axs[0], show=False);
+    >>> fig, axs = plt.subplots(1, 3, figsize=(16, 4), layout="constrained")
+    >>> ax = patch.viz.waterfall(cmap="RdBu_r", ax=axs[0], show=False)
     >>>
-    >>> for a, dim in enumerate(['time', 'distance']):
-    >>>     spec  = patch.spectra(dim=dim, kind='PSD',  db=True)
-    >>>     ax = spec.viz.spectraplot(log=True, ax=axs[a+1], show=False, scale=[0.7, 1])
-    >>>     ax.set_title('Fourier-Transform along ' + dim.capitalize())
-    >>>     if dim=='time':
-    >>>         ax.set_ylim((5,200))
+    >>> for a, dim in enumerate(["time", "distance"]):
+    ...     spec = patch.spectra(dim=dim, kind="PSD", db=True)
+    ...     ax = spec.viz.spectraplot(
+    ...         log=True,
+    ...         ax=axs[a + 1],
+    ...         show=False,
+    ...         scale=[0.7, 1],
+    ...     )
+    ...     _ = ax.set_title("Fourier-Transform along " + dim.capitalize())
+    ...     if dim == "time":
+    ...         _ = ax.set_ylim((5, 200))
 
+    Example 2: seismic event
 
-    Example 2: seimic event
     >>> import dascore as dc
     >>> import matplotlib.pyplot as plt
     >>>
     >>> patch = dc.examples.example_event_2().T.decimate(time=10)
-    >>> new = dc.to_datetime64(patch.coords.get_array('time') + 1704400020)
+    >>> new = dc.to_datetime64(patch.coords.get_array("time") + 1704400020)
     >>> patch = patch.update_coords(time=new)
     >>>
-    >>> fig, axs = plt.subplots(1,3,figsize=(18,5), layout='constrained')
-    >>> patch.viz.waterfall(cmap='RdBu_r', ax=axs[0]);
+    >>> fig, axs = plt.subplots(1, 3, figsize=(18, 5), layout="constrained")
+    >>> ax = patch.viz.waterfall(cmap="RdBu_r", ax=axs[0])
     >>>
-    >>> for a, dim in enumerate(['time', 'distance']):
-    >>>     spec  = patch.spectra(dim=dim, kind='PSD',  db=True)
-    >>>     ax = spec.viz.spectraplot(log=True, ax=axs[a+1], show=False, scale=[0.7, 1])
-    >>>     ax.set_title('Fourier-Transform along ' + dim.capitalize())
+    >>> for a, dim in enumerate(["time", "distance"]):
+    ...     spec = patch.spectra(dim=dim, kind="PSD", db=True)
+    ...     ax = spec.viz.spectraplot(
+    ...         log=True,
+    ...         ax=axs[a + 1],
+    ...         show=False,
+    ...         scale=[0.7, 1],
+    ...     )
+    ...     _ = ax.set_title("Fourier-Transform along " + dim.capitalize())
     """
     spec = patch.dft(dim=dim, real=True, pad=True).abs()
     if db:
--- a/dascore/transform/spectra.py
+++ b/dascore/transform/spectra.py
@@ -5,17 +5,15 @@ from __future__ import annotations
 import numpy as np
 
 from dascore import units
-
 from dascore.constants import PatchType
 from dascore.utils.patch import _get_dx_or_spacing_and_axes, patch_function
-
 
 
 @patch_function(required_dims=("time",), history="full")
 def spectra(
     patch: PatchType,
     dim: str = "time",
-    kind: str = 'PSD',
+    kind: str = "PSD",
     db: bool = False,
 ) -> PatchType:
     """
@@ -87,43 +85,46 @@ def spectra(
     >>>     ax = spec.viz.spectraplot(log=True, ax=axs[a+1], show=False, scale=[0.7, 1])
     >>>     ax.set_title('Fourier-Transform along ' + dim.capitalize())
     """
-    spec = patch.dft(dim=dim, real=True,  pad=True).abs()
+    spec = patch.dft(dim=dim, real=True, pad=True).abs()
     if db:
         # add a small number to prevent division-by zero when taking the log10
         spec += np.finfo(spec.data.dtype).eps
 
-
     _, axis = _get_dx_or_spacing_and_axes(patch, dim)
     n = patch.data.shape[axis[0]]
 
-    if kind.upper() == 'AS':
+    if kind.upper() == "AS":
         out = spec
         if db:
             out = 20 * out.log10()
 
-        out = out.update(attrs={'data_type':"Amplitude Spectrum"})
+        out = out.update(attrs={"data_type": "Amplitude Spectrum"})
 
-    elif kind.upper() == 'PS':
-        out = spec*spec
+    elif kind.upper() == "PS":
+        out = spec * spec
         if db:
             out = 10 * out.log10()
 
-        out = out.update(attrs={'data_type':"Power Spectrum",} )
+        out = out.update(
+            attrs={
+                "data_type": "Power Spectrum",
+            }
+        )
 
-    elif kind.upper() == 'PSD':
-        fsamp = 1 / (spec.get_coord('ft_'+dim).step * spec.get_coord('ft_'+dim).units)
-        out = spec*spec  / (n * fsamp)
+    elif kind.upper() == "PSD":
+        fsamp = 1 / (
+            spec.get_coord("ft_" + dim).step * spec.get_coord("ft_" + dim).units
+        )
+        out = spec * spec / (n * fsamp)
         if db:
             out = 10 * out.log10()
 
-        out = out.update(attrs={'data_type':"Power Spectral Density"})
+        out = out.update(attrs={"data_type": "Power Spectral Density"})
 
     else:
-       raise ValueError('ERROR: Unknown option: kind=', kind)
-
+        raise ValueError("ERROR: Unknown option: kind=", kind)
 
     if db:
         out = out.set_units(units.dB)
 
     return out
-
--- a/dascore/transform/spectra.py
+++ b/dascore/transform/spectra.py
@@ -1,0 +1,129 @@
+"""Patch function for calculating spectra as a matrix"""
+
+from __future__ import annotations
+
+import numpy as np
+
+from dascore import units
+
+from dascore.constants import PatchType
+from dascore.utils.patch import _get_dx_or_spacing_and_axes, patch_function
+
+
+
+@patch_function(required_dims=("time",), history="full")
+def spectra(
+    patch: PatchType,
+    dim: str = "time",
+    kind: str = 'PSD',
+    db: bool = False,
+) -> PatchType:
+    """
+    Compute the amplitude spectrum, power spectrum, or power spectral density.
+
+    This function applies a real-valued discrete Fourier transform along the
+    specified dimension and returns the requested spectral representation as a
+    DASCore Patch. The transformed dimension is replaced by its corresponding
+    frequency-domain coordinate.
+
+    Parameters
+    ----------
+    patch
+        Input patch containing the data to transform.
+    dim
+        Dimension along which to compute the spectrum. Defaults to ``"time"``.
+    kind
+        Spectral quantity to return. Options are:
+
+        - ``"AS"``: amplitude spectrum, ``abs(DFT(x))``.
+        - ``"PS"``: power spectrum, ``abs(DFT(x)) ** 2``.
+        - ``"PSD"``: power spectral density, normalized by
+          ``n_samples * sampling_rate``.
+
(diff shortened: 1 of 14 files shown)


d-chambers:
Hey @andreas-wuestefeld,

I will take a look. Overall, I think these are  good features/changes, but there might be a few tweaks we can make to to improve user experience. 

Also, there is quite a lot here. No worries for this one, but in the future it will be easier/faster to review PRs with smaller changes (eg this one could be broken into 3 probably).

dascore/transform/spectra.py
```diff
@@ -0,0 +1,139 @@
...
+from dascore.utils.patch import _get_dx_or_spacing_and_axes, patch_function
+
+
+@patch_function()
+def spectra(
+    patch: PatchType,
+    dim: str = "time",
+    kind: str = "PSD",
+    db: bool = False,
+) -> PatchType:
```
d-chambers, line 18: I think `spectrum` is probably better here. I know there are multiple spectra if the patch has more than one channel, but we have other functions like `mean` and `median` where this applies as well (they arent called `means` and `medians`).

dascore/transform/spectra.py
```diff
@@ -0,0 +1,139 @@
...
+
+    _, axis = _get_dx_or_spacing_and_axes(patch, dim)
+    n = patch.data.shape[axis[0]]
+
+    if kind.upper() == "AS":
+        out = spec
+        if db:
+            out = 20 * out.log10()
+
+        out = out.update(attrs={"data_type": "Amplitude Spectrum"})
```
d-chambers, line 111: This should all be `Patch.update_attrs`, the difference is that `Patch.update` will blow away old attrs that are not part of the default. 

Please write a test or two for this. 

1. A patch with attrs.tag  = "something" survives the transform. This tests that known attr values survive the transform. 
2. A patch with a non-standard attr survive the transform (patch.attrs.custom_attr = 2)

dascore/transform/spectra.py
```diff
@@ -0,0 +1,139 @@
...
+    kind: str = "PSD",
+    db: bool = False,
+) -> PatchType:
+    """
+    Compute the amplitude spectrum, power spectrum, or power spectral density.
+
+    This function applies a real-valued discrete Fourier transform along the
+    specified dimension and returns the requested spectral representation as a
+    DASCore Patch. The transformed dimension is replaced by its corresponding
+    frequency-domain coordinate.
```
d-chambers, line 25: Also explain the relationship between this and  `dft`. IE `dft` is for reversible Fourier Transform only, this function is to get derived products that aren't reversible. Add a `See Also` section to link to `dft`. See DASCore's [cross referencing](https://dascore.org/contributing/documentation.html#cross-references) for how to do this. Also, worth adding a see also lint to this function from dft.

dascore/transform/spectra.py
```diff
@@ -0,0 +1,139 @@
...
+    ...     spec = patch.spectra(dim=dim, kind="PSD", db=True)
+    ...     ax = spec.viz.specplot(
+    ...         log=True,
+    ...         ax=axs[a + 1],
+    ...         show=False,
+    ...         scale=[0.7, 1],
+    ...     )
+    ...     _ = ax.set_title("Fourier-Transform along " + dim.capitalize())
+    """
+    spec = patch.dft(dim=dim, real=True, pad=True).abs()
```
d-chambers, line 98: DFT used like this should be idempotent, so one can do:

```python
p1 = patch.dft("time").spectra()
# And
p2 = patch.spectra()
# then
p1 == p2
```
but would you mind writing a test for exactly this case just in case? I could see why one would like to use this function in either case.

dascore/utils/plotting.py
```diff
@@ -26,6 +28,8 @@ def _get_cmap(cmap):
     """Return a color map from a colormap or string."""
     if isinstance(cmap, str):  # get color map if a string was passed
         cmap = plt.get_cmap(cmap)
+        cmap.set_over(cmap(1.0))
+        cmap.set_under(cmap(0.0))
```
d-chambers, line 32: Apparently the cmap can be shared in some MPL, best to make a copy:   

```python
cmap = plt.get_cmap(cmap).copy()
cmap.set_over(cmap(1.0))
cmap.set_under(cmap(0.0))
```

dascore/viz/specplot.py
```diff
@@ -0,0 +1,78 @@
...
+    scale: float | Sequence[float] | None = (0.5, 1),
+    scale_type: Literal["relative", "absolute"] = "relative",
+    interpolation: str | None = "bilinear",
+    log: bool = False,
+    show: bool = False,
+    **kwargs,
+) -> plt.Axes:
+    """
+    Plot spectra of a patch
+    """
```
d-chambers, line 30: This is not a correct docstring. It needs a fully numpy sytle docstring. See [here](https://dascore.org/contributing/documentation.html#doc-strings)

dascore/viz/specplot.py
```diff
@@ -0,0 +1,78 @@
...
+        show=show,
+        scale=scale,
+        scale_type=scale_type,
+        interpolation=interpolation,
+        log=False,
+    )
+
+    """
+    Customisations
+    """
```
d-chambers, line 43: Not wrong I guess, but these actually create string objects and are not used like this in any other DASCore code. Just use a code comment and not a multi-line string.  Also, "customization" doesn't really tell us anything.

dascore/viz/specplot.py
```diff
@@ -0,0 +1,78 @@
...
+    Format axis labels.
+    """
+    for dim, x in zip(dims, ["y", "x"], strict=True):
+        if dim != fft_dim:
+            continue
+
+        # replace labels with publication-ready descriptors
+        label = _get_dim_label(patch, dim)
+        label = label.replace("Ft_time", "Frequency")
+        label = label.replace("Ft_distance", "Wavenumber")
```
d-chambers, line 62: Probably better to have a dict of replacements you want to make at the top of the module.

dascore/viz/specplot.py
```diff
@@ -0,0 +1,78 @@
...
+    is_fft_dim = [d.startswith("ft_") for d in dims]
+    if all(x is False for x in is_fft_dim):
+        raise Exception("Patch does not contain a Fourier-transformed coordinate")
+
+    fft_axis = is_fft_dim.index(True)
+    fft_dim = dims[fft_axis]
+
+    """
+    Format axis labels.
+    """
```
d-chambers, line 54: Same here; dont use a mutli-line string unless it is as docstring.

dascore/viz/waterfall.py
```diff
@@ -131,15 +158,47 @@ def _add_colorbar(ax, im, patch, log):
     cb.set_label(label)
 
 
+def _default_colormap_per_datatype(patch):
+    """
+    Select a default colormap based on datatype
+    """
+    default_maps = {
```
d-chambers, line 165: Probably better to put this in dascore.constants, or  at least at the top of the module.

dascore/viz/waterfall.py
```diff
@@ -131,15 +158,47 @@ def _add_colorbar(ax, im, patch, log):
...
+        "phase_rate": "seismic",
+        "temperature": "coolwarm",
+        "temperature_gradient": "RdYlBu_r",
+    }
+
+    this_type = str(patch.attrs.get("data_type", "")).lower()
+    if this_type in default_maps.keys():
+        cmap = default_maps[this_type]
+    else:
+        cmap = "bwr"  # default
```
d-chambers, line 187: better: 
```python
this_type = str(patch.attrs.get("data_type", "")).lower()
cmap = default_map.get(this_type, "bwr")
```

tests/test_viz/test_waterfall.py
```diff
@@ -258,3 +258,9 @@ def test_scale_zero_raises(self, random_patch):
 
     def test_percent_scale(self, random_patch):
         """Ensure the percent unit works with scale."""
```
d-chambers, line 260: Empty test. I guess this was here before? You can just remove it, or try to make it work. It looks like scale should work with percentages, probably not that hard to test/fix.

dascore/transform/spectra.py
```diff
@@ -0,0 +1,139 @@
...
+    ...         _ = ax.set_ylim((5, 200))
+
+    Example 2: seismic event
+
+    >>> import dascore as dc
+    >>> import matplotlib.pyplot as plt
+    >>>
+    >>> patch = dc.examples.example_event_2().T.decimate(time=10)
+    >>> new = dc.to_datetime64(patch.coords.get_array("time") + 1704400020)
+    >>> patch = patch.update_coords(time=new)
```
d-chambers, line 83: These magic numbers seem a bit odd, also there is a lot of code here. I recommend just showing the first example, and deleting this one as it doesn't really add much to teaching users how to use this function. It just shows it works on other patches too, which one would assume anyway no?

dascore/transform/spectra.py
```diff
@@ -0,0 +1,139 @@
...
+
+import numpy as np
+
+from dascore import units
+from dascore.constants import PatchType
+from dascore.utils.patch import _get_dx_or_spacing_and_axes, patch_function
+
+
+@patch_function()
+def spectra(
```
d-chambers, line 13: This function body is a bit ugly with all the branches. One might simply consider:

```python
kind = kind.upper()
data_types = {
    "AS": "Amplitude Spectrum",
    "PS": "Power Spectrum",
    "PSD": "Power Spectral Density",
}
spec = patch.dft(dim=dim, real=True, pad=True).abs()
spec = spec + np.finfo(spec.data.dtype).eps if db else spec
# Scale spectral quantity from dft.
if kind == "AS":
    out = spec
elif kind == "PS":
    out = spec * spec
elif kind == "PSD":
    n = len(patch.get_coord(dim))
    ft_coord = spec.get_coord("ft_" + dim)
    fsamp = 1 / (ft_coord.step * ft_coord.units)
    out = spec * spec / (n * fsamp)
else:
    msg = f"Unknown kind={kind!r}. Expected one of: {tuple(data_types)}."
    raise ParameterError(msg)
# Stabilize db and scale DB
if db:
    db_scale = 20 if kind == "AS" else 10
    out = db_scale * out.log10()
    out = out.set_units(units.dB)

return out.update_attrs(data_type=data_types[kind])
```

dascore/transform/spectra.py
```diff
@@ -0,0 +1,139 @@
...
+    patch: PatchType,
+    dim: str = "time",
+    kind: str = "PSD",
```
d-chambers, line 16: Would be good to make this a `kind: Literal['PSD', 'PS', 'AS'] = 'PSD'` to help out static analysis.

dascore/viz/specplot.py
```diff
@@ -0,0 +1,78 @@
...
+        log=False,
+    )
+
+    """
+    Customisations
+    """
+    dims = patch.coords.dims
+    is_fft_dim = [d.startswith("ft_") for d in dims]
+    if all(x is False for x in is_fft_dim):
+        raise Exception("Patch does not contain a Fourier-transformed coordinate")
```
d-chambers, line 47: Bare `Exception` is discouraged. Probably want `CoordError` from dascore.exceptions

andreas-wuestefeld:
@d-chambers
Yes it got bigger than anticipated, and then difficult for me to untangle.

Take your time to review. I will probably not have much time next week to work much on it.

dascore/viz/waterfall.py
```diff
@@ -117,11 +117,38 @@ def _format_axis_labels(ax, patch, dims_r):
             ax.invert_yaxis()
 
 
-def _add_colorbar(ax, im, patch, log):
+def _add_colorbar(ax, im, patch, log, scale):
```
d-chambers, line 120: Better, I think, to pass in data to calc min/maxes so we dont have to make another array copy for abs, then just undo log10 op.

Could also clean up the if branching for extents. 

Something like:
```python
def _add_colorbar(ax, im, data, patch, log, scale):
    """
    Add a colorbar with appropriate labels to the plot.
    """
    mi = 10**data.min()
    mx = 10**data.max()
    
    above, below = False, False
    if scale is not None and len(scale) == 2:
        above = (mx > scale[1]) and not np.isclose(scale[1], mx)
        below = (mi < scale[0]) and not np.isclose(scale[0], mi)
    extend_map = {
        (True, True): "both", 
        (True, False): "max",
        (False, True): "min",
    }
    extend = extend_map.get((above, below), "neither")
    cb = ax.get_figure().colorbar(
        im, ax=ax, fraction=0.05, pad=0.025, extend=extend, extendfrac=0.025
    )
    data_type = str(patch.attrs.get("data_type", ""))
    data_units = get_quantity_str(patch.attrs.data_units) or ""
    dunits = f" [{data_units}]" if (data_type and data_units) else f"{data_units}"
    if log:
        dunits = f"{dunits} - log_10"
    label = f"{data_type}{dunits}"
    cb.set_label(label)
```

dascore/viz/waterfall.py
```diff
@@ -117,11 +117,38 @@ def _format_axis_labels(ax, patch, dims_r):
...
+    else:
+        extend = "neither"
+
+    cb = ax.get_figure().colorbar(
+        im, ax=ax, fraction=0.05, pad=0.025, extend=extend, extendfrac=0.025
+    )
+
     data_type = str(patch.attrs.get("data_type", ""))
     data_units = get_quantity_str(patch.attrs.data_units) or ""
     dunits = f" [{data_units}]" if (data_type and data_units) else f"{data_units}"
```
d-chambers, line 154: Why only square brackets when data_type is defined?

dascore/viz/waterfall.py
```diff
@@ -235,14 +296,24 @@ def waterfall(
     patch = _validate_patch_dims(patch)
     # Setup axes and data
     ax = _get_ax(ax)
-    cmap = _get_cmap(cmap)
-    data = np.log10(np.absolute(patch.data)) if log else patch.data
+    float_dtype = np.result_type(patch.data.dtype, np.float64)
+    eps = np.finfo(float_dtype).eps
```
d-chambers, line 300: Code rabbit has a good point here. It would be better to abs the patch first, then calc `eps` so `eps` cant end up a complex number.

d-chambers (changes requested):
A few high-level things. 

1) in the PR description you say (regarding the default color bar selection): "Note: this may be put in a user-customizable .ini file, if there are needs for more such customisations". However, this isnt true (nor should it be yet, working on a wider DASCore config story on dev branch). Please edit the post so that it correctly reflects features/changes, agents will use it for changelog generation later on. 

2)  `spectra` is probably better named `spectrum` (see comment for justification) and could be simplified. 

3) For spectra (or spectrum) I think the units are handled correctly, but could you ensure there are a few simple tests on a happy path case? 

4) It would be nice if this patch function was smart enough to handle time domain patches or frequency domain patches. For example, you might imagine a case where the patch has already been DFT'ed, then the user wouldn't have to redo that, but could just get the appropriate scaling. So, 

```python
dft_patch = patch.dft("time", real=True)

spec1 = patch.spectrum("time")
spec2 = dft_patch.spectrum("time")

# and
assert spec1 == spec2
```

I can help you with this is you would like. 

5) Regarding the proper scaling for PSD,  I will think on it a bit more, but right now it seems straight-forward. If we consider a signal with units of V transformed along the time axis, naively `dft` returns units of V/Hz. It is already a amplitude density, so I think AS should be abs(DFT) * f_s (to get units in V), PS = AS**2, and PSD = PS / f_s or (abs(DFT)**2 * f_s).  The other issue is accounting for the one-sided nature of the real DFT. You probably want to double the non-zero frequencies (and maybe non-nyquist too? I dont remember).  Consider something like this:

```python
patch = patch + patch.select(ft_time=(0 +eps, ...))
```
Could also be worthwhile to ensure there are no negative frequencies, or just fold them into the positive ones if there are. In fact, this should probably be it's own function in this module, maybe `fold_ft` or something like that. 

Sorry if this is a lot, let me know what questions you have and if you need a hand with any of it.

dascore/transform/spectra.py
andreas-wuestefeld: Naively I would assume that patch.dft("time").spectra() take the spectrum of a spectrum (i,e, the [cepstrum](https://en.wikipedia.org/wiki/Cepstrum))

Or we could make that the default behaviour: the function just converts the dft into various amplitude conversions. We could then actually just make functions called ".psd()" and ".powerspectrum()"

d-chambers: Good point, I have never used a cepstrum myself and so I didnt think of that ;). so `power_spectrum`, `psd`, and `amplitude_spectrum`  simply re-scale the dft and handle the one-sided frequency issue? That seems like a clean approach.  

What would calling these functions on a time series do? Just raise a error if the data are not complex, or they dont look like they have undergone a `dft`?

andreas-wuestefeld: The more I think about it, the more I am inclined to handle the output directly in "dft()", simply adding a optional parameter to dft
Scale = ['fft', 'as', 'ps', 'psd']

d-chambers: Good idea,  then we just need to reject some scales in idft. Maybe just a check that the input array is complex (if we don't do that already) else raise

andreas-wuestefeld: Cleanest is probably to withdraw the pull request, and submit several smaller ones, which also include your suggestions.

d-chambers: Yes,  that's probably the cleanest way forward.

dascore/transform/spectra.py
andreas-wuestefeld: This "magic number" was used to trick DAScore into understanding the time axis is datetime64
example_event_2() has actually a numpy array:
➤ Coordinates (distance: 601, time: 1001)
    *distance: CoordRange( min: 400 max: 1000 step: 1 shape: (601,) dtype: int64 units: m )
    *time: CoordRange( min: 0.0 max: 0.1 step: 0.0001 shape: (1001,) dtype: float64 units: s )

the conversion was mainly for more visually pleasing output. This perhaps should be done by default in the examples.py?

andreas-wuestefeld:
Closed since this is was separated into several separate smaller PRs

PR #693, opened 2026-05-31
added output type to dft; data_type is now 'fourier transform'

andreas-wuestefeld:
## Description
This PR adds option to output various amplitude outputs types to DFT (amplitude spectra, power spectra, or power spectral density). Optionally data can be converted to decibel
The default output is "FFT", so that it remains backward compatible

## Checklist

I have (if applicable):

## Summary by CodeRabbit

* **New Features**
  * dft adds selectable outputs: FFT, amplitude spectrum (AS), power spectrum (PS), and power spectral density (PSD); case-insensitive selection and optional dB conversion for spectral outputs.

* **Bug Fixes**
  * DFT preserves prior data-type metadata, sets transformed data type to "fourier transform", rejects invalid output or dB-for-FFT, and prevents inversion for non-FFT spectral outputs.

* **Style/UX**
  * Dimension labels now read units from coordinate metadata for plotting.

* **Tests**
  * Expanded tests cover all output kinds, scaling, dB behavior, validation, and inversion rules.

* **Chores**
  * Lint workflow push trigger restricted to master branch.

### DIFF
--- a/dascore/transform/fourier.py
+++ b/dascore/transform/fourier.py
@@ -277,7 +277,7 @@ def dft(
     """
     output = output.upper()
     if output not in DFT_OUTPUT_TYPES:
-        msg = f"Unknown kind={output!r}. Expected one of: {DFT_OUTPUT_TYPES}."
+        msg = f"Unknown output={output!r}. Expected one of: {DFT_OUTPUT_TYPES}."
         raise ValueError(msg)
     if output == "FFT" and db:
         msg = "db=True is only supported for output='AS', 'PS', or 'PSD'."
--- a/dascore/transform/fourier.py
+++ b/dascore/transform/fourier.py
@@ -160,46 +160,22 @@ def _get_transformed_domain_extent(patch, dims):
     return extent
 
 
-def _get_one_sided_spectral_factor(patch, real):
-    """Get the factor for converting rfft spectra to one-sided spectra."""
-    if real is None:
-        return 1
-
-    # rfft drops negative frequencies only along the real axis. The returned
-    # factor is shaped for broadcasting over the full transformed array.
-    ft_dim = f"ft_{real}"
-    axis = patch.get_axis(ft_dim)
-    factor = np.ones(patch.coord_shapes[ft_dim])
-    n_samples = len(patch.get_coord(real))
-
-    # Even-length real inputs include a Nyquist bin; odd-length inputs don't.
-    stop = -1 if n_samples % 2 == 0 else None
-    factor[slice(1, stop)] = 2
-
-    shape = [1] * patch.ndim
-    shape[axis] = len(factor)
-    return factor.reshape(shape)
-
-
 def _convert_dft_spectral_amplitudes(patch, output, dims, real, db):
     """Convert the FFT output to spectral amplitude representations."""
     amp = patch.abs()
     extent = _get_transformed_domain_extent(amp, dims)
-    one_sided_factor = _get_one_sided_spectral_factor(patch, real)
     data_units = _get_dft_data_units(patch, dims, output)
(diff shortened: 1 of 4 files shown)


d-chambers:
Hey @andreas-wuestefeld,
I am still thinking about the scaling, but I think the changes I just pushed are right. I will check back tomorrow, but if you want to take a look go for it. 

Also, I tried to untangle the changes from the plotting from this PR, may need to do a bit more there, not sure. Will check tomorrow.

andreas-wuestefeld:
@d-chambers This looks excellent. Thanks for the effort in figuring out scaling. And verifying for idft...
Maybe its worth adding an example line for the new features?

One additional idea would be to automatically convert 1/sec to Hz in dascore.units.invert_quantity
This had the effect that some units would not be reduced by PINT. For example the PSD of strainrate data is now reduced to ε^2/s, but more in line with convention would be a unit of (ε/s)^2 / Hz

the lint-errors seem straight forward.

andreas-wuestefeld:
@d-chambers Something seems to be off with decibels for 'PS' and 'PSD' calculations
Please hold merging until I can have a closer look!

d-chambers:
sure, no problem.

d-chambers:
And of course, feel free to push more changes (it was me that hijacked your pr 😉 )

dascore/transform/fourier.py
```diff
@@ -110,13 +145,83 @@ def _get_untransformed_dims(patch, dims):
     return out
 
 
+def _get_transformed_domain_extent(patch, dims):
+    """Get the transformed-domain extent from the DFT bin spacing."""
+    extent = 1
```
andreas-wuestefeld, line 150: if `real = True`, all of AS, PS, and PSD show values that are double compared to `real = false`

My understanding is that usually  these methods calculate the complex FFT, and then only preserve the positive frequencies .

I am not sure about the "extent" parameter, but if we add something like
`extent = 2 if real else 1`
then the amplitude values match again between `real = True` and `real = False`

We could also force `real = True` for non-FFT outputs, but FK plots need negative frequencies (wavenumbers)

This is above my theoretical understanding, so I am just pointing out my observations :-)

d-chambers: If you set `extent = 2 if real else 1` you would essentially be undoing `_get_one_sided_spectral_factor` wouldn't you?  So you don't want the non-fft real output to be doubled? 

If you calculate real-valued inputs, for example, you need to account for the energy in the negative frequency bins, otherwise the values you derive wont preserve energy (Parseval's identity) nor will the values represent the amplitude of the base harmonics (depending on your scaling).  I spent some time a few years ago convincing myself this was right with simple harmonics. You can see my thinking in [DASCore's dft note](https://dascore.org/notes/dft_notes.html). 

I agree it can be a bit confusing though, because the doubling happens only from some combination of `output` and `real`. Maybe the best approach is just to document clearly `dft` doesnt do this and leave it as an exercise for the user to account for the missing energy?

andreas-wuestefeld: Yes, that may be the best way. I also did "convince myself" about Fourier amplitudes a while a go while working on a different software package.
As long as it is properly documented, the user can apply conversions to their desired output or scaling

d-chambers:
Ok, I removed the one-sided scaling and (hopefully) documented properly. Mind taking a quick look that this produces what you expect? Once this merges I think we can properly finish and test #691. It would also be good to wrap up #685 and do a release in the next couple of weeks.

andreas-wuestefeld:
This is now good to merge from my perspective!

d-chambers (approved):

d-chambers:
Merge away 🥳

PR #694, opened 2026-05-31
Waterfall with default colormaps per datatype

andreas-wuestefeld:
## Description
Waterfall had the default colormap "bwr". This PR makes use of the data_type attribute to set defaul colormaps depending on patch.datatype. Maps are defined in constants.py
Also, added a boolean option "cbar" to indicate if a colorbar is to be shown or not (earlier behavior was to toggle with cmap=None)

Fixed with adding eps number to data to avoid division-by-zero warning when taking the log

Colorbar now automatically shows "extend" triangles if the color limits do not encompass the full data range

## Checklist

I have (if applicable):

## Summary by CodeRabbit

* **New Features**
  * Automatic default colormap selection for waterfall plots based on data type.
  * Option to omit the colorbar by passing no colormap (cmap=None).

* **Bug Fixes**
  * Safer log scaling that avoids -inf on zeros and handles all-zero data.
  * Improved colorbar extension behavior to reflect actual data limits.

* **Tests**
  * Added tests for colorbar visibility, log handling, scaling modes, and default colormap selection.

### DIFF
--- a/dascore/constants.py
+++ b/dascore/constants.py
@@ -147,12 +147,12 @@ for each attribute.
 
 select_values_description = """
 Any dimension name can be passed as key, and the values can be:
-    - a Slice or a tuple of (min, max) for that dimension. 
+    - a Slice or a tuple of (min, max) for that dimension.
       `None` and ... both indicate open intervals.
-    - an array of values to select, which must be a subset of the 
+    - an array of values to select, which must be a subset of the
       coordinate array.
     - an array of booleans of the same length as the coordinate where
-      `True` indicates values to keep. 
+      `True` indicates values to keep.
 """
 
 check_behavior_description = """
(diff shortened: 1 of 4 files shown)


d-chambers:
Hey @andreas-wuestefeld,

I just made two small tweaks.

1) In order to avoid adding a new parameter (cbar) and breaking the existing meaning of `None` for cmap, I set the default to "default" which means, use a default for a given data type (your selection logic). Sorry, I probably misled you for setting `cmap`'s default to `None` without remembering that already had a meaning.  I could also see this being named "auto" if you think that is better. 

2) Re-worked the log path so that we don't have to take abs of the array twice (which can be expensive for large arrays). 

Otherwise, lgtm, merge when you are ready.

d-chambers (approved):

d-chambers:
Actually, it looks like the spec-plot depends on this behavior. So I will merge for now,  so I can re-base that branch to master. If you want to tweak anything else we can open another PR.

PR #696, opened 2026-06-01
Coerce numeric relative offsets for timedelta64 coordinates (fixes #604)

### LINKED ISSUE #604: stft fails unit coercion with a timedelta64 coordinate
When the time dimension is `timedelta64` type rather than `datetime64` the value specified in `stft` must be a timedelta64, but it should also support a float/int value which gets  converted into a `timedelta64`. When the time coordinate is a `datetime64` this works as intended.

## Example

```python
import dascore as dc

time = patch.get_coord("time")

new = patch.update_coords(time=time.values - time.values[-1])

new.stft(time=.1)  # Raises an error

new.stft(time=dc.to_timedelta64(.1))  # works

```

Also, the time coordinate seems to be wrong on the transform in this case.

## Versions
 - OS [e.g. Ubuntu 20.04]:
 - DASCore Version [e.g. 0.0.5]: Master branch
 - Python Version [e.g. 3.10]:

gaoflow:
Fixes #604.

## Problem

With a `timedelta64` time coordinate, passing a numeric window length to `stft` raised a numpy ufunc type error, even though the equivalent explicit-duration call worked:

```python
import dascore as dc

time = patch.get_coord("time")
new = patch.update_coords(time=time.values - time.values[-1])  # timedelta64 coord

new.stft(time=0.1)                      # numpy._core._exceptions._UFuncBinaryResolutionError
new.stft(time=dc.to_timedelta64(0.1))   # works
```

## Root cause

`BaseCoord._get_compatible_value` (`dascore/core/coords.py`) only coerced a *relative* offset to `timedelta64` when the coordinate was `datetime64`:

```python
if relative:
    if np.issubdtype(self.dtype, np.datetime64):
        value = dc.to_timedelta64(value)
    value = self._get_relative_values(value)
```

For a `timedelta64` coordinate the numeric offset stayed a `float`, so `_get_relative_values` evaluated `coord.min() + value` as `timedelta64 + float` and raised. A relative offset into a time-like coordinate is a duration in both cases.

## Fix

Use the existing `dtype_time_like` helper (already imported) so a numeric relative offset is coerced to `timedelta64` for both `datetime64` and `timedelta64` coordinates.

## Verification

- `new.stft(time=0.1)` now succeeds and its output is **identical** (`Patch.equals`) to `new.stft(time=dc.to_timedelta64(0.1))`, which also resolves the reporter's secondary note that "the time coordinate seems to be wrong on the transform in this case".
- The `datetime64` path is unchanged; relative `select` on a `timedelta64` coordinate with a numeric offset now works too.
- Two regression tests added (both **fail on `master`** with the ufunc error, pass with this change):
  - `tests/test_core/test_coords.py::TestSelect::test_select_relative_numeric_offset_timedelta` (root cause, coord level)
  - `tests/test_transform/test_fourier.py::TestSTFT::test_numeric_window_with_timedelta_coord` (end-to-end stft)
- Full suite: **5936 passed, 80 skipped, 4 xfailed**. `pre-commit run --files ...` (ruff, ruff-format, typos) clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Relative time-based coordinate selection now behaves consistently for all time-like coordinate types when using numeric or duration offsets.

* **Tests**
  * Added tests verifying relative selection with numeric offsets on duration-style time coordinates and a regression test for spectral transforms using alternative time coordinate formats.

### DIFF
--- a/dascore/core/coords.py
+++ b/dascore/core/coords.py
@@ -625,9 +625,13 @@ class BaseCoord(DascoreBaseModel, abc.ABC):
         # if null or ... just return None
         if not is_array(value) and (pd.isnull(value) or value is Ellipsis):
             return None
-        # special case for datetime and relative
+        # special case for datetime/timedelta and relative
         if relative:
-            if np.issubdtype(self.dtype, np.datetime64):
+            # A relative offset into a time-like coordinate is a duration, so
+            # coerce it to timedelta64 for both datetime64 and timedelta64
+            # coords (previously only datetime64 was handled, so a numeric
+            # offset into a timedelta64 coord raised a ufunc type error).
+            if dtype_time_like(self.dtype):
                 value = dc.to_timedelta64(value)
             value = self._get_relative_values(value)
         # apply validators. These can, eg, coerce to correct dtype.
--- a/dascore/core/coords.py
+++ b/dascore/core/coords.py
@@ -627,10 +627,7 @@ class BaseCoord(DascoreBaseModel, abc.ABC):
             return None
         # special case for datetime/timedelta and relative
         if relative:
-            # A relative offset into a time-like coordinate is a duration, so
-            # coerce it to timedelta64 for both datetime64 and timedelta64
-            # coords (previously only datetime64 was handled, so a numeric
-            # offset into a timedelta64 coord raised a ufunc type error).
+            # A relative offset into any time-like coord is a duration.
             if dtype_time_like(self.dtype):
                 value = dc.to_timedelta64(value)
             value = self._get_relative_values(value)
--- a/tests/test_core/test_coords.py
+++ b/tests/test_core/test_coords.py
@@ -633,6 +633,24 @@ class TestSelect:
         else:
             assert ind_1 == ind_2
 
+    def test_select_relative_numeric_offset_timedelta(self):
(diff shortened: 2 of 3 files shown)


dascore/core/coords.py
```diff
@@ -625,9 +625,13 @@ def _get_compatible_value(self, value, relative=False):
...
             return None
-        # special case for datetime and relative
+        # special case for datetime/timedelta and relative
         if relative:
-            if np.issubdtype(self.dtype, np.datetime64):
+            # A relative offset into a time-like coordinate is a duration, so
+            # coerce it to timedelta64 for both datetime64 and timedelta64
+            # coords (previously only datetime64 was handled, so a numeric
+            # offset into a timedelta64 coord raised a ufunc type error).
+            if dtype_time_like(self.dtype):
```
d-chambers: We really dont need this much explanation, could you please shorten it to one comment line?

d-chambers (changes requested):

dascore/core/coords.py
binggao1230: Done — collapsed it to a single line.

d-chambers (approved):

PR #698, opened 2026-06-01
Retain step when slicing an evenly sampled coordinate (fixes #567)

### LINKED ISSUE #567: Evenly sampled coords don't retain step when squished
## Description
Evenly sampled coordinates should retain step even with length of 1, for example from a patch.select.

## Example
```python
import dascore as dc

sub_patch = patch.select(time=1, samples=1)

time = sub_patch.get_coord("time")

assert time.step is not None
```

## Expected behavior
The above test should pass, meaning the time coordinate, although of length 1, should retain its step. 

## Versions
 - OS [e.g. Ubuntu 20.04]:
 - DASCore Version [e.g. 0.0.5]:
 - Python Version [e.g. 3.10]:

gaoflow:
## Description

Closes #567.

Evenly sampled coordinates should keep their `step` even when reduced to length 1, e.g. via `patch.select(time=1, samples=1)`. Currently the step is lost:

```python
import dascore as dc

sub_patch = patch.select(time=1, samples=1)
time = sub_patch.get_coord("time")
assert time.step is not None  # fails on master
```

## Root cause

`CoordRange.__getitem__` handled slice indexers by materializing the values (`self.values[item]`) and rebuilding via `get_coord(data=...)`. For a length-1 result that produces a `CoordMonotonicArray`, which has no `step` — so an evenly sampled coordinate silently becomes non-evenly-sampled.

## Fix

A (possibly strided) slice of an evenly sampled range is still evenly sampled, so `CoordRange.__getitem__` now computes the resulting `start`/`stop`/`step` directly for slice indexers and returns a `CoordRange`. Non-slice indexers (boolean / fancy arrays) are unchanged. Strided, reverse, and empty slices were verified to behave as before.

## Verification

- New regression tests at the coord level (`TestSelect::test_single_sample_retains_step`, plus a datetime variant) and at the patch level (the issue's exact example). All fail on master / pass with the fix.
- Full test suite: 5937 passed, 80 skipped, 4 xfailed.
- pre-commit clean.

### DIFF
--- a/dascore/core/coords.py
+++ b/dascore/core/coords.py
@@ -1169,8 +1169,16 @@ class CoordRange(BaseCoord):
         if isinstance(item, slice):
             start = None if item.start is ... else item.start
             end = None if item.stop is ... else item.stop
-            # Todo we can probably add more intelligent logic for slices.
             item = slice(start, end, item.step)
+            # A (possibly strided) slice of an evenly sampled range is still
+            # evenly sampled, so preserve the step rather than collapsing to a
+            # CoordMonotonicArray (which loses step for len==1). See #567.
+            indices = range(len(self))[item]
+            if len(indices):
+                new_step = self.step * indices.step
+                new_start = self.start + indices.start * self.step
+                new_stop = new_start + new_step * len(indices)
+                return self.new(start=new_start, stop=new_stop, step=new_step)
         out = self.values[item]
         return get_coord(data=out, units=self.units)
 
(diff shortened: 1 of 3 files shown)


d-chambers:
@gaoflow, this looks pretty good, will run the CI. What do you think of DASCore's coordinate system? Do you have ideas for simplifying the architecture (for a future version/PR of course).

d-chambers (approved):

PR #702, opened 2026-06-04
add option to set data_type in patch_function decorator

d-chambers:
## Description

  Adds a `data_type` argument to `patch_function` so patch methods can declare a constant output `data_type` in the decorator. The decorator batches this update with history so it does not need an extra `update_attrs`
  call when both are applied.

  This also moves constant `data_type` assignments for strain-rate, FBE, STA/LTA, and kurtosis transforms into their decorators, and registers the new snake_case transform data types in `VALID_DATA_TYPES`.

## Summary by CodeRabbit

* **New Features**
  * Added support for three new data types: `frequency_band_energy`, `kurtosis`, and `stalta`.

* **Improvements**
  * Standardized how transforms declare and report their output data types, ensuring consistent data type naming across all transforms.
  * Frequency Band Energy, Kurtosis, and STA/LTA transforms now properly register their output data types.

### DIFF
--- a/dascore/constants.py
+++ b/dascore/constants.py
@@ -71,6 +71,9 @@ VALID_DATA_TYPES = (
     "strain",
     "temperature",
     "temperature_gradient",
+    "frequency_band_energy",
+    "kurtosis",
+    "stalta",
 )
 
 # Valid categories (of instruments)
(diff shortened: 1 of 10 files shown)


d-chambers:
replaced by #709

PR #721, opened 2026-06-29
Merge master into dev

d-chambers:
This merges master into dev.

## Summary by CodeRabbit

* **New Features**
  * Added new transforms for frequency-band energy, STA/LTA, and kurtosis.
  * Added support for additional Febus, OptoDAS, and SR-4731 file formats.
  * Introduced a new spectrum plotting view and improved waterfall plot defaults.

* **Bug Fixes**
  * Improved handling of time-like values, chunking, rolling windows, and coordinate selection.
  * Fixed array handling so non-NumPy array-like inputs are preserved more reliably.
  * Expanded Fourier output modes and improved plot labeling/units.

## Changelog

none


d-chambers:
Triaged the CodeRabbit review on this PR. Since this is a `master -> dev` merge, the review covered the full delta, so most findings are about pre-existing code rather than changes introduced here. To keep this PR focused on the merge, the legitimate bugs have been split out into their own issues rather than fixed inline:

- #722 — specplot drops `**kwargs` instead of forwarding them to waterfall
- #723 — specplot raises a bare `ValueError` for non-2D patches
- #724 — `saturate_add`/`saturate_subtract` convert NaT into a real time-like value
- #725 — kurtosis fails on descending or unevenly-sampled coordinates
- #726 — stalta does not validate that the long-term window exceeds the short-term window

A few CodeRabbit findings were reviewed and are **not** being actioned (noted here so they aren't re-raised):

- `dascore/compat.py` (`_make_immutable`) — preserving wrapper objects without forcing a NumPy copy is intentional; the docstring explicitly states DASCore does not guarantee physical immutability for non-NumPy array-likes.
- `dascore/io/sr4731/core.py` (`scan`) — returning `list[PatchAttrs]` is the correct scan contract and matches the other readers; the suggested `payload["attrs"]` wrapping is a false positive.

The remaining items (e.g. fourier chained-transform metadata, aggregate bool units, `quantity_to_samps`, `get_intervals` time-like normalization, febus lazy reads, `_reduce_time_like` precision) are pre-existing and better handled as separate follow-ups, not in this merge.

Issue #583, opened 2025-11-20
Spool iteration not robust to files changing distance


d-chambers:
## Description

Suppose a spool has two types of patches. In the first, distance ranges from 0 too 100. In the second type of file distance ranges from 0 to 10. Now, if the spool is selected with `distance=(20, 3)` the spool wont automatically throw out type 2 files because it doesn't know anything about distance. However, when trying to read the index of a type 2 file an `IndexError` is raised. 

## Example

A bit complicated to set up but described above.

## Expected behavior

Ideally, when we redo the indexing, the spool select operation will remove the type two files. Until then, it should just be skipped when iterating, but still raise when accessed directly.

## Versions
 - OS [e.g. Ubuntu 20.04]:
 - DASCore Version [e.g. 0.0.5]:
 - Python Version [e.g. 3.10]:

PR #738, opened 2026-07-03
Speed up reading and merging many files

d-chambers:
## Description

Reading and merging spools with many (hundreds of) files was much slower than the underlying file IO: profiling a 300-file merge showed only ~30% of the time going to actual HDF5 reads, the rest being fixed per-file python overhead (~12 ms/file). Merging also held every source patch plus the concatenated copy in memory at once. This PR removes the biggest overheads in the load path and streams merges:

- `DirectorySpool._load_patch` reads directly through the file's `FiberIO`, skipping `dc.read` format detection and the spool indexing machinery applied to its throwaway output. Range selection is only applied when the patch is actually trimmed (tracked by the instruction dataframe's `_modified` flag). The fast path requires a concrete format/version and falls back to `dc.read` otherwise (including for multi-patch files without selection, so each row still maps to its own patch).
- `MemorySpool` is lazy: the managing dataframes are only built when an operation needs them, and `len`, integer access, and iteration are served directly from an immutable patch tuple for spools created from patches. This makes the spool created (and immediately unwrapped) by every `FiberIO.read` call nearly free. Equality is normalized so it doesn't depend on whether the lazy dataframes were built, derived spools don't retain their parents, and `new_from_df` no longer builds dummy dataframes it discards.
- **Merges stream into a pre-allocated array** sized from the instruction dataframe instead of loading all sources and concatenating: peak memory for a merge drops ~45% (to output + one file) and a full-data copy is removed. Mixed dtypes upcast as `np.concatenate` would; inexact estimates grow or trim the buffer; inestimable cases (unknown steps, no varying indexed dim) use the previous materialized merge. Both paths produce identical results, and gap/continuity checks are unchanged (they happen on the dataframes at chunk time, before any data is read).
- `_get_patches_from_index` looks up instruction rows via a cached `groupby` mapping instead of scanning the whole instruction dataframe per patch, removing O(sources × outputs) scaling when iterating chunked spools.
- Patches trimmed to nothing (see #583) raise a dedicated `MissingPatchError` (an `IndexError` subclass for backwards compatibility) which iteration catches by type; the previous "out of bounds" message matching is gone and unexpected `IndexError`s always propagate. `FileSpool` had the same latent crash and gets the same treatment.
- `get_quantity_str` (called by the pydantic serializers of attrs/coord models on every dump) caches string validation and unit stringification. Caches are only keyed on types with exact equality (`str` and `pint.Unit`); a `Quantity` cannot key a cache because equal quantities with different units (`1 m == 100 cm`) would collide. A test guards this.
- New CodSpeed benchmarks cover indexing, merging, and chunked iteration over a many-file directory, plus in-memory spool creation/access and many-patch merges (~1 s total wall time).

Benchmarks (DASDAE format, warm cache, local disk):

| Scenario | Before | After |
|---|---|---|
| 300 files (1.4 GB), `chunk(time=None)` merge | ~4.6 s | ~1.9 s |
| — peak memory for the above | ~3.0 GB | ~1.6 GB |
| 1000 small files, merge all | ~12 s | ~3.4 s |
| 1000 small files, chunk into 250 patches | ~15 s | ~4.7 s |

For reference, raw pytables reads + `np.concatenate` of the same 300-file dataset take ~1.5 s. A likely follow-up (separate PR) is optional executor-parallel patch loading for cold/network storage.

## Checklist

I have (if applicable):

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Faster patch loading for directory-based workflows with improved handling of multi-patch files.
  * Enhanced lazy in-memory spooling for more efficient access patterns.
* **Bug Fixes**
  * Missing/trimmed entries now raise `MissingPatchError` (still catchable as `IndexError`), improving patch disambiguation.
  * Chunk/merge behavior is more consistent: dimension order and coordinate metadata are preserved; streaming merge matches materialized output.
  * Quantity-to-string formatting avoids over-normalizing equal quantities.
* **Tests**
  * Added coverage for multi-patch directory behavior, lazy spool semantics, streaming merge edge cases, and unit formatting.

### DIFF
--- a/dascore/clients/dirspool.py
+++ b/dascore/clients/dirspool.py
@@ -15,12 +15,7 @@ from typing_extensions import Self
 
 import dascore as dc
 from dascore.constants import PROGRESS_LEVELS
-from dascore.core.spool import (
-    BaseSpool,
-    DataFrameSpool,
-    MemorySpool,
-    _patches_from_read,
-)
+from dascore.core.spool import BaseSpool, DataFrameSpool, MemorySpool
 from dascore.exceptions import MissingPatchError
 from dascore.io.indexer import AbstractIndexer, DirectoryIndexer
 from dascore.utils.docs import compose_docstring
@@ -134,7 +129,7 @@ class DirectorySpool(DataFrameSpool):
         final_kwargs.update(self._select_kwargs)
         patches = self._read_patches(final_kwargs)
         if patches is None:  # fast path doesn't apply, use generic read.
-            patches = _patches_from_read(dc.read(**final_kwargs))
+            patches = list(dc.read(**final_kwargs))
         if not patches:
             # Iteration skips these with a warning, see #583.
             msg = (
@@ -172,7 +167,7 @@ class DirectorySpool(DataFrameSpool):
         spool = fiber_io.read(kwargs["path"], **select)
         if not isinstance(spool, MemorySpool):
             return None
-        patches = spool._unprocessed_patches()
+        patches = list(spool)
         # Without selection, a multi-patch file is ambiguous: the row refers
         # to one specific patch. Let the generic path pick the right one.
         if len(patches) > 1 and not select:
--- a/dascore/clients/dirspool.py
+++ b/dascore/clients/dirspool.py
@@ -15,7 +15,7 @@ from typing_extensions import Self
 
 import dascore as dc
 from dascore.constants import PROGRESS_LEVELS
-from dascore.core.spool import BaseSpool, DataFrameSpool
+from dascore.core.spool import BaseSpool, DataFrameSpool, MemorySpool
 from dascore.io.indexer import AbstractIndexer, DirectoryIndexer
 from dascore.utils.docs import compose_docstring
 from dascore.utils.pd import adjust_segments
@@ -126,5 +126,46 @@ class DirectorySpool(DataFrameSpool):
         """Given a row from the managed dataframe, return a patch."""
         final_kwargs = dict(kwargs)
         final_kwargs.update(self._select_kwargs)
-        patch = dc.read(**final_kwargs)[0]
-        return patch
+        patches = self._read_patches(final_kwargs)
+        if patches is None:  # fast path doesn't apply, use generic read.
+            patches = [dc.read(**final_kwargs)[0]]
+        if not len(patches):
+            # Note: DataFrameSpool.__iter__ matches on "out of bounds" to
+            # skip patches trimmed to nothing (see #583).
+            msg = "index of [0] is out of bounds for spool."
+            raise IndexError(msg)
+        return patches[0]
+
+    def _read_patches(self, kwargs) -> list | None:
+        """
+        Read patches directly through the file's FiberIO.
+
+        This skips the format detection of dc.read and the spool indexing
+        machinery applied to its output, which add up when loading many
+        files. Returns None when the fast path can't be safely used.
+        """
+        fmt, version = kwargs.get("file_format"), kwargs.get("file_version")
+        if not fmt:
+            return None
+        fiber_io = dc.io.FiberIO.manager.get_fiberio(format=fmt, version=version)
+        # Only apply select kwargs when the source patch is trimmed by the
+        # instruction df or the spool itself; otherwise the whole file is
+        # wanted and selection is wasted work.
+        if kwargs.get("_modified") or self._select_kwargs:
+            select = {
+                k: v
+                for k, v in kwargs.items()
+                if k not in ("path", "file_format", "file_version")
+                and not k.startswith("_")
+            }
+        else:
+            select = {}
+        spool = fiber_io.read(kwargs["path"], **select)
+        if not isinstance(spool, MemorySpool):
+            return None
+        patches = spool._unprocessed_patches()
+        # Without selection, a multi-patch file is ambiguous: the row refers
+        # to one specific patch. Let the generic path pick the right one.
+        if len(patches) > 1 and not select:
+            return None
+        return patches
--- a/dascore/clients/dirspool.py
+++ b/dascore/clients/dirspool.py
@@ -129,14 +129,14 @@ class DirectorySpool(DataFrameSpool):
         patches = self._read_patches(final_kwargs)
         if patches is None:  # fast path doesn't apply, use generic read.
             patches = [dc.read(**final_kwargs)[0]]
-        if not len(patches):
+        if not patches:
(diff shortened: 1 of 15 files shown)


dascore/clients/dirspool.py
```diff
@@ -126,5 +126,46 @@ def _load_patch(self, kwargs) -> Self:
...
+            select = {}
+        spool = fiber_io.read(kwargs["path"], **select)
+        if not isinstance(spool, MemorySpool):
+            return None
+        patches = spool._unprocessed_patches()
+        # Without selection, a multi-patch file is ambiguous: the row refers
+        # to one specific patch. Let the generic path pick the right one.
+        if len(patches) > 1 and not select:
+            return None
+        return patches
```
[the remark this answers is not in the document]
d-chambers: Good catch — the fast path now requires a concrete format and version and otherwise defers to dc.read, which detects the actual version from the file. Fixed in e0c8916d, with a regression test.

dascore/units.py
```diff
@@ -220,22 +231,34 @@ def get_quantity_str(quant_value: str | Quantity | None) -> str | None:
...
 
 
+@cache
+def _validate_quantity_str(quant_str: str) -> None:
+    """Raise a UnitError if the string doesn't specify a valid quantity."""
+    try:
+        get_quantity(quant_str)
+    except UndefinedUnitError:
+        msg = f"DASCore failed to parse the following unit/quantity: {quant_str}"
+        raise UnitError(msg)
```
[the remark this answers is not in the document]
d-chambers: Fixed in e0c8916d (raise ... from e).

dascore/core/spool.py
```diff
@@ -665,13 +667,101 @@ def get_contents(self) -> pd.DataFrame:
...
+        if (
+            patches is not None
+            and not self._select_kwargs
+            and isinstance(item, int | np.integer)
+        ):
+            try:
+                return patches[item]
+            except IndexError:
+                msg = f"index of [{item}] is out of bounds for spool."
+                raise IndexError(msg)
```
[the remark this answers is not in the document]
d-chambers: Fixed in e0c8916d (raise ... from None).

dascore/core/spool.py
```diff
@@ -53,6 +62,62 @@
 T = TypeVar("T")
 
 
+def _dim_slice(ndim: int, axis: int, start: int, stop: int) -> tuple:
```
d-chambers: There is already a utility for this, `broadcast_for_index`, just use that.

dascore/core/spool.py
```diff
@@ -53,6 +62,62 @@
...
+    """Get an nd slice tuple selecting start:stop along axis."""
+    out = [slice(None)] * ndim
+    out[axis] = slice(start, stop)
+    return tuple(out)
+
+
+def _patches_from_read(spool: BaseSpool) -> list[PatchType]:
+    """Get the raw patch list from a spool returned by dc.read."""
+    if isinstance(spool, MemorySpool):
```
d-chambers: Do we really need this helper? Shouldn't the list transformation do this already then we can just remove this code?

dascore/core/spool.py
```diff
@@ -372,6 +441,15 @@ def _get_source_df(self):
     def _get_instruction_df(self):
         """Function to get the current df."""
 
+    def _get_instruction_indices(self):
```
d-chambers: Can we just inline this?

dascore/core/spool.py
```diff
@@ -663,15 +825,106 @@ def get_contents(self) -> pd.DataFrame:
...
         if data is not None:
-            dfs = self._get_dummy_dataframes(patches_to_df(data))
-            self._df, self._source_df, self._instruction_df = dfs
+            if isinstance(data, dc.Patch):
+                self._patches = (data,)
+            elif isinstance(data, Sequence) and all(
+                isinstance(x, dc.Patch) for x in data
+            ):
+                # Copy to an immutable tuple so neither mutating the input
+                # sequence nor the stored one can change the spool contents.
```
d-chambers: remove this comment

dascore/core/spool.py
```diff
@@ -692,6 +945,26 @@ def _load_patch(self, kwargs) -> Self:
         """Load the patch into memory."""
         return kwargs["patch"]
 
+    def _unprocessed_patches(self) -> list[PatchType]:
```
d-chambers: This has a bed smell to it. SHouldn't Spool.__iter__ already handle this such that `list(spool)` would do the same thing as this?

dascore/utils/patch.py
```diff
@@ -445,14 +454,40 @@ def _get_new_coord(df, merge_dim, coords, drop_conflicting=False):
...
-    new_coord = _get_new_coord(df, merge_dim, coords, drop_conf_coords)
+    new_coord = _get_merged_coord(df, merge_dim, coords, drop_conf_coords)
     coord = new_coord.coord_map[merge_dim] if merge_dim in dims else None
     new_attrs = combine_patch_attrs(attrs, merge_dim, coord=coord, **merge_kwargs)
     patch = dc.Patch(data=new_data, coords=new_coord, attrs=new_attrs, dims=dims)
     new_dict = {"patch": patch}
     return [new_dict]
 
 
+def _patch_dim_summary(patch: PatchType) -> dict:
```
d-chambers: This should probably be a method on the `CoordManager`

dascore/core/spool.py
d-chambers: Done in a43da4ca — replaced with broadcast_for_index.

dascore/core/spool.py
d-chambers: Right — with the lazy iteration fast path, list(spool) already returns the raw patches, so both this helper and _unprocessed_patches are removed in a43da4ca.

dascore/core/spool.py
d-chambers: Inlined into _get_patches_from_index with a manual cache lookup in a43da4ca.

dascore/core/spool.py
d-chambers: Removed in a43da4ca.

dascore/core/spool.py
d-chambers: Agreed — the lazy __iter__ fast path makes list(spool) equivalent, so this method is removed and all callers use list(spool) (a43da4ca).

dascore/utils/patch.py
d-chambers: Moved to CoordManager._get_dim_summary in a43da4ca.

PR #749, opened 2026-07-10
Add CoordSegmented: compositional segmented coordinates

d-chambers:
## Description

Adds `CoordSegmented`, a coordinate composed of an ordered sequence of monotonic segments (`CoordRange` or `CoordMonotonicArray`), plus a `concat_coords` function and two new `BaseCoord` methods (`simplify`, `get_discontinuities`). This PR only adds the class — nothing constructs it in production code yet; wiring patch merge/chunk to produce it is a follow-up PR.

### Motivation

Merging nearly-contiguous patches currently forces a binary coordinate choice: snap to a uniform `CoordRange` (endpoints exact, but interior values can be arbitrarily wrong when a tolerated gap is absorbed) or materialize a full `CoordArray` (exact but heavy, structure lost). `CoordSegmented` is the middle ground: it holds the original segments, its values are exactly their concatenation, and segment boundaries record data gaps without altering any value. Memory is O(segments), min/max are exact without materialization, and gap introspection becomes a coordinate primitive.

### What's included

- **`CoordSegmented`**: full `BaseCoord` contract. Select/getitem preserve segment structure and match `CoordMonotonicArray` semantics exactly (tested by parity); a window inside a gap selects nothing; a window inside one block degrades to that block's plain coordinate. `step` is None; reverse-sorted coords supported; ns-precision datetimes never pass through floats.
- **`concat_coords(*coords, units=None)`**: truth-preserving concatenation. Inputs are ordered by envelope; overlapping, mixed-dtype, mixed-units, or mixed-direction inputs raise. Returns a plain coordinate when inputs fuse to a single segment.
- **Normal form**: exactly uniform array segments promote to ranges (bit-exact check only), exactly continuing ranges fuse, adjacent irregular arrays fuse, single segments degrade to the bare coordinate — so equal values compare and fingerprint equal regardless of assembly order.
- **`BaseCoord.simplify(tolerance)`**: bounded-error alternative to `snap()` — re-fits segments as evenly sampled ranges, never moving any value more than `tolerance`. Other coords return self.
- **`BaseCoord.get_discontinuities(kind, tolerance)`**: dataframe of segment boundaries and gaps; the primitive behind a future `spool.get_gaps()`.
- **`CoordSegmented.from_array(array, tolerance=None)`**: segment a monotonic array into uniform runs so gaps *inside* an array coordinate become queryable boundaries; optional tolerance re-fits with bounded error.
- **`patch.split_gaps(dim=None)`** and a `split=` kwarg on `dc.write`: a written patch must be contiguous, so saving a patch with gapped dimensional coords raises by default; `split=True` writes each contiguous section as its own patch (multi-patch formats only — the check lives once in the generic write dispatcher, individual FiberIOs are untouched apart from a declarative `multi_patch_write` attribute). This removes any need for native segment serialization: split pieces are plain patches, and gaps persist *between* patches where spools and indices already see them. Includes a DASDAE fix uniquifying group names within a write batch (pre-existing silent overwrite of same-named patches).
- **`get_coord(segments=...)`** dispatches to `concat_coords`; `CoordSegmented` and `concat_coords` are exported from `dascore.core`.

### Notes for review

- Direct `CoordSegmented(...)` construction with segments that fuse to one coordinate raises (a pydantic constructor cannot return a different class); `concat_coords` is the intended API.
- The `segments` field is typed `tuple[BaseCoord, ...]` instead of a concrete union: pydantic union dispatch runs member before-validators against foreign instances. Concrete types are enforced in the model validator, and a `field_serializer` keeps round-trips lossless.
- 98 new tests, including parity against equivalent monotonic coords, error-bound checks for `simplify`, and an ns bit-exactness regression test.

## Changelog

- added: `CoordSegmented` and `concat_coords(...)` for exact piecewise-monotonic coordinates with queryable sampling changes and gaps.
- added: `patch.split_gaps()` and `dc.write(..., split=True)` split segmented dimensions into contiguous patches; `split=True` needs a format that can hold multiple patches and raises `ParameterError` otherwise.

## Checklist

I have (if applicable):

## Summary by CodeRabbit

* **New Features**
  * Added segmented (piecewise) coordinate support with normalization and `concat_coords`, plus `CoordSegmented` and `get_coord(segments=...)`.
  * Exposed additional coordinate-related symbols and added a new display style for segmented coordinates.
  * Added `Patch.split_gaps` and updated `write(..., split=...)` to split gapped dimensional coordinates when supported (enabled for DASDAE and Pickle output).
* **Bug Fixes**
  * Fixed descending monotonic indexing/selection to preserve exactness for time-like and large integer values.
  * Made `simplify` and discontinuity reporting consistently return concrete results.
* **Tests**
  * Added extensive segmented-coordinate and gap-handling coverage, including serialization and indexing/simplify/snap behavior.

### DIFF
--- a/dascore/core/coords.py
+++ b/dascore/core/coords.py
@@ -2021,9 +2021,14 @@ class CoordSegmented(BaseCoord):
         out = np.concatenate([x.values for x in self.segments])
         return array(out.astype(self.dtype, copy=False))
 
-    @cached_method
     def _as_monotonic(self) -> CoordMonotonicArray:
-        """Return an equivalent (materialized) monotonic array coord."""
+        """
+        Return an equivalent (materialized) monotonic array coord.
+
+        Deliberately not cached: transient materialization for rare
+        operations (snap, get_next_index) must not permanently defeat the
+        O(segments) memory model.
+        """
         return CoordMonotonicArray(values=self.values, units=self.units)
 
     def _min(self):
@@ -2129,14 +2134,35 @@ class CoordSegmented(BaseCoord):
             return self._select_by_array(args, relative=relative, samples=samples)
         if samples:
             return self._select_by_samples(args)
-        # Index math is delegated to an equivalent monotonic array (the values
-        # are identical by construction); the resulting slice is re-applied
-        # segment-wise so structure is preserved. Array args were handled
-        # above, so monotonic select always returns a slice here.
-        _, indexer = self._as_monotonic().select(args, relative=relative)
-        assert isinstance(indexer, slice)
-        start_i, stop_i, _ = indexer.indices(len(self))
-        return self._slice_segments(start_i, stop_i), indexer
+        # Delegate to each segment and compose the global slice from segment
+        # offsets. The window over a monotonic coordinate keeps a contiguous
+        # run of samples, and range segments answer in O(1), so selection
+        # stays O(segments) and never materializes the concatenated values.
+        v1, v2 = self.get_slice_tuple(args, relative=relative)
+        kept, lo, hi = [], None, None
+        for seg, off in zip(self.segments, self._segment_offsets()):
+            seg_min, seg_max = seg.min(), seg.max()
+            if (v2 is not None and seg_min > v2) or (v1 is not None and seg_max < v1):
+                continue  # entirely outside the window
+            inside_lo = v1 is None or v1 <= seg_min
+            inside_hi = v2 is None or v2 >= seg_max
+            if inside_lo and inside_hi:  # entirely inside; keep untouched
+                sub, seg_lo, seg_hi = seg, 0, len(seg)
+            else:  # boundary segment; delegate the exact trim
+                sub, indexer = seg.select((v1, v2))
+                seg_lo, seg_hi, _ = indexer.indices(len(seg))
+                if seg_hi <= seg_lo:
+                    continue
+            if lo is None:
+                lo = int(off) + seg_lo
+            hi = int(off) + seg_hi
+            kept.append(sub)
+        if not kept:
+            return self.empty(), slice(0, 0)
+        new = self._rebuild(kept)
+        start = None if lo == 0 else lo
+        stop = None if hi >= len(self) else hi
+        return new, slice(start, stop)
 
     def _get_index(self, value, forward=True):
         """Get the index corresponding to a value."""
--- a/dascore/core/coords.py
+++ b/dascore/core/coords.py
@@ -2049,10 +2049,8 @@ class CoordSegmented(BaseCoord):
         return self.__class__(segments=segments)
 
     def _rebuild(self, segments) -> BaseCoord:
-        """Build the simplest coordinate from a list of segments."""
+        """Build the simplest coordinate from a (non-empty) list of segments."""
         segments = tuple(segments)
-        if not segments:
-            return self.empty()
         if len(segments) == 1:
             return segments[0]
         try:
@@ -2108,12 +2106,12 @@ class CoordSegmented(BaseCoord):
             return self._select_by_samples(args)
         # Index math is delegated to an equivalent monotonic array (the values
         # are identical by construction); the resulting slice is re-applied
-        # segment-wise so structure is preserved.
+        # segment-wise so structure is preserved. Array args were handled
+        # above, so monotonic select always returns a slice here.
         _, indexer = self._as_monotonic().select(args, relative=relative)
-        if isinstance(indexer, slice):
-            start_i, stop_i, _ = indexer.indices(len(self))
-            return self._slice_segments(start_i, stop_i), indexer
-        return get_coord(data=self.values[indexer], units=self.units), indexer
+        assert isinstance(indexer, slice)
+        start_i, stop_i, _ = indexer.indices(len(self))
+        return self._slice_segments(start_i, stop_i), indexer
 
     def _get_index(self, value, forward=True):
         """Get the index corresponding to a value."""
@@ -2236,8 +2234,9 @@ class CoordSegmented(BaseCoord):
         else:
             step = span / (n - 1)
             zero = 0
-        if (step > zero) != ascending or step == zero:
-            return None
+        # Strictly monotonic segments guarantee a nonzero step matching the
+        # sort direction.
+        assert step != zero and (step > zero) == ascending
         candidate = CoordRange(
             start=first, stop=last + step, step=step, units=self.units
         ).change_length(n)
@@ -2349,7 +2348,8 @@ def concat_coords(*coords, units=None) -> BaseCoord:
     _validate_segment_compat(tuple(flat))
     multi = [x for x in flat if len(x) > 1]
     ascending = multi[0].sorted if multi else True
-    flat.sort(key=lambda x: to_float(x.min()), reverse=not ascending)
+    # Sort on native values; float conversion would collapse ns datetimes.
+    flat.sort(key=lambda x: x.min(), reverse=not ascending)
     if len(flat) == 1:
         return _maybe_promote_segment(flat[0])
     _validate_segment_chain(tuple(flat))
--- a/dascore/core/coords.py
+++ b/dascore/core/coords.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import abc
 import fnmatch
 import hashlib
+import itertools
 import json
 import re
 from collections.abc import Sized
@@ -21,6 +22,7 @@ import numpy as np
 import pandas as pd
 from pydantic import (
     ValidationError,
+    field_serializer,
     field_validator,
     model_validator,
 )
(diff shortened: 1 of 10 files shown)


dascore/core/coords.py
```diff
@@ -1714,6 +1772,593 @@ def reverse_sorted(self):
...
+            raise CoordError(msg)
+    if units is not None:
+        flat = [x.set_units(units) for x in flat]
+    if not flat:
+        msg = "concat_coords requires at least one non-empty coordinate."
+        raise CoordError(msg)
+    _validate_segment_compat(tuple(flat))
+    multi = [x for x in flat if len(x) > 1]
+    ascending = multi[0].sorted if multi else True
+    flat.sort(key=lambda x: to_float(x.min()), reverse=not ascending)
```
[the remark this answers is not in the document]
d-chambers: Good catch — fixed in 0ff3f5ee. Sorting now uses native values (`key=lambda x: x.min()`), and a regression test (`test_ns_precision_sort_order`) constructs two segments 35 ns apart whose float-converted mins collide, asserting correct ordering.

dascore/core/coords.py
```diff
@@ -1714,6 +1772,593 @@ def reverse_sorted(self):
...
+        if not isinstance(seg, CoordRange | CoordMonotonicArray):
+            msg = (
+                "Segments must be CoordRange or CoordMonotonicArray, "
+                f"got {type(seg)}."
+            )
+            raise CoordError(msg)
+        if not len(seg):
+            msg = "Segments must not be empty."
+            raise CoordError(msg)
+    dtypes = {np.dtype(s.dtype) for s in segments}
```
d-chambers: [P1] Preserve values when numeric dtypes differ

This compatibility check allows `int64` and `float64` segments because their dtype kinds are both numeric, but `values` then uses `np.result_type`/`concatenate` and can silently alter integers above 2^53. For example, concatenating float `[0., 1.]` with `int64([2**53 + 1, 2**53 + 3])` produces float values which cast back to `[2**53, 2**53 + 4]`. That violates `concat_coords`' truth-preserving guarantee. Please reject differing dtypes or explicitly prove that conversion to the common dtype is lossless for every segment.

dascore/core/coords.py
```diff
@@ -1714,6 +1772,593 @@ def reverse_sorted(self):
...
+        """Apply select, return selected coords and index for selecting data."""
+        if is_array(args):
+            return self._select_by_array(args, relative=relative, samples=samples)
+        if samples:
+            return self._select_by_samples(args)
+        # Index math is delegated to an equivalent monotonic array (the values
+        # are identical by construction); the resulting slice is re-applied
+        # segment-wise so structure is preserved. Array args were handled
+        # above, so monotonic select always returns a slice here.
+        _, indexer = self._as_monotonic().select(args, relative=relative)
```
d-chambers: [P1] Avoid float-based lookup for reverse nanosecond datetimes

Delegating here reaches `CoordMonotonicArray._get_index`, which negates reverse coordinates through `to_float`. At modern epochs that collapses sub-128 ns differences. A reverse segmented datetime with offsets `[95, 85, 75] ns` and `[60, 50, 40] ns` from `2024-01-01` returns the entire coordinate for `select((t0 + 75 ns, t0 + 75 ns))` instead of the single 75 ns sample. The new class therefore loses the ns-exact behavior for reverse selection even though construction is exact. Please use native-value search for reverse time coordinates (ideally fixing the shared monotonic lookup) and add this regression.

dascore/core/coords.py
```diff
@@ -1714,6 +1772,593 @@ def reverse_sorted(self):
...
+                result.append(run_fit if run_fit is not None else run[0])
+                run, run_fit = [seg], self._fit_run([seg], tol)
+        result.append(run_fit if run_fit is not None else run[0])
+        return self._rebuild(result)
+
+    def _get_tolerance(self, tolerance):
+        """Coerce the tolerance to the dtype expected for value deviations."""
+        if tolerance is None:
+            tolerance = 0
+        if dtype_time_like(self.dtype):
```
d-chambers, line 2263: [P2] Normalize quantity-valued tolerances

Unit-bearing tolerances currently fail before fitting: `coord.set_units('m').simplify(1 * get_quantity('m'))` raises a Pint `DimensionalityError` at the comparison with dimensionless zero, and a time coordinate with `1 * get_quantity('s')` raises `NotImplementedError` in `to_timedelta64`. `get_discontinuities(kind='gaps', tolerance=...)` shares this path. Please convert quantity tolerances to `self.units` (or seconds for time-like coordinates) before validating and comparing them.

dascore/core/coords.py
```diff
@@ -2044,6 +2694,13 @@ def _maybe_get_start_stop_step(data):
                 return _min, _max + _step, _step, is_monotonic
         return None, None, None, is_monotonic
 
+    if segments is not None:
+        others = (data, values, start, min, stop, max, step, shape, dtype)
```
d-chambers: [P1] Allow the standard CoordManager round trip

This rejects the representation used by `CoordManager`: `_get_coord_dim_map` calls `coord.model_dump(exclude_defaults=True)` and then `get_coord(**coord)`, so a `CoordSegmented` arrives with `segments` plus derived `shape`, `dtype`, and `step`. Consequently `patch.update_coords(distance=segmented_coord)` raises `CoordError: segments cannot be combined with other get_coord inputs`, making the new coordinate impossible to attach to a Patch. Please either preserve BaseCoord instances in CoordManager or allow/validate these derived fields when `segments` is present; add a Patch/CoordManager integration test.

d-chambers (commented):
I found four correctness gaps in targeted adversarial checks. The focused coordinate suites pass (3,486 tests), but these cases are outside the current coverage and affect the advertised BaseCoord/truth-preserving contracts.

dascore/core/coords.py
d-chambers: [P1] Keep value selection O(segments), not O(samples)

This delegation materializes and caches the entire concatenated coordinate on the first ordinary `select`. With 1,000,000 samples in 100 range segments, a cold selection took ~122 ms and retained ~7.6 MiB, versus ~0.10 ms for an already-materialized `CoordMonotonicArray`; at 5,000,000 samples it took ~615 ms. Warm calls are faster only because `_as_monotonic` permanently caches the full array, forfeiting the advertised O(segments) memory model. Since Patch.select is the common operation this coordinate is meant to accelerate, please locate bounds from segment envelopes and delegate only within the boundary segments, producing the same contiguous slice without materializing all values. A scaling benchmark should lock this in.

d-chambers (commented):
Second-pass review focused on scaling and integration. The four prior correctness findings remain open because the PR head is unchanged. I found one additional performance issue that cuts against the primary O(segments) motivation.

dascore/core/coords.py
d-chambers: Fixed in b5548b40 — segments must now share one dtype kind (mixing int64 + float64 raises `CoordError`); width promotion within a kind (i4+i8, f4+f8, M8[s]+M8[ns]) remains allowed since it is lossless. Test includes the 2**53+1 preservation case with array segments.

dascore/core/coords.py
d-chambers: Fixed in b5548b40 at the shared lookup as suggested: `CoordMonotonicArray._get_index` now negates time-like values on their int ns view and signed numerics natively (`_negate_for_search`); only unsigned ints, which would wrap, still fall back to float. Added your exact reverse-segmented scenario plus a plain reverse monotonic ns regression.

dascore/core/coords.py
d-chambers: Fixed in b5548b40 — `_get_tolerance` converts pint quantities to the coordinate's units (seconds for time-like coords) before validation, covering both `simplify` and `get_discontinuities`. Tests cover m/cm equivalence, ms on time coords, and dimensionality mismatch raising.

dascore/core/coords.py
d-chambers: Fixed in b5548b40 — `get_coord(segments=...)` now tolerates the derived `shape`/`dtype` fields and dict-form segments from `model_dump`, so the CoordManager round trip works. Integration test attaches a segmented coord to a Patch, exercises `update_coords`, cross-gap `select`, and patch equality.

dascore/core/coords.py
d-chambers: Valid — fixed in 88ab9781. Selection now composes the global slice from per-segment selects: fully-inside segments are kept untouched, only the (at most two) boundary segments trim, so it's O(segments) with no materialization. Measured on your scenario scaled up (5M samples, 100 range segments): ~1.8 ms cold vs the ~615 ms reviewed, with zero retained value arrays — `_as_monotonic` is no longer cached (transient, for the rare snap/get_next_index paths only). The lock-in is structural rather than a timing benchmark: `test_select_never_materializes` patches `CoordSegmented.values` to raise and runs window/empty/full/relative selects on a 20M-sample coord, so any future regression to a materializing path fails deterministically rather than flakily.

Note on the review summary: the four first-round findings were already fixed in b5548b40 — this review was submitted against 0ff3f5ee and raced that push. They're addressed on the current head.

dascore/io/core.py
```diff
@@ -1294,11 +1297,50 @@ def get_format(
...
+    for patch, has_gaps in zip(spool, gapped):
+        patches.extend(patch.split_gaps() if has_gaps else [patch])
+    if len(patches) > 1 and not fiber_io.multi_patch_write:
+        msg = (
+            f"Format {fiber_io.name} writes a single patch per file, so "
+            "gapped patches cannot be split into it. Use patch.split_gaps() "
+            "and write each patch to its own file."
+        )
+        raise ParameterError(msg)
+    return dc.spool(patches)
```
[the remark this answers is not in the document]
d-chambers: Applied in 66c4d466 — the helper now materializes the in-memory spool once and both passes use the same list, with a strict zip. (Cheap by construction: this path only runs for MemorySpool, whose patches are already in RAM.)

dascore/io/dasdae/core.py
```diff
@@ -73,10 +74,16 @@ def write(
...
+        # batch so same-named patches (e.g. gap-split siblings that differ
+        # only along a non-named dimension) don't overwrite each other.
         patch_names = get_patch_names(patches).values
+        counts: dict[str, int] = {}
         for patch, name in zip(patches, patch_names):
-            _save_patch(patch, waveforms, name)
+            num = counts.get(name, 0)
+            counts[name] = num + 1
+            unique_name = name if num == 0 else f"{name}__{num}"
+            _save_patch(patch, waveforms, unique_name)
```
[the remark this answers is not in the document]
d-chambers: Partially applied in 66c4d466: the zip is now strict=True, so any drift between the name pass and the patch pass fails loudly instead of writing wrong group names. I deliberately did not materialize the spool with list(): the DASDAE writer streams one patch at a time, and materializing would regress memory for large spool-to-file conversions to fix a skew that strict zip now detects. For spool types with consistent iteration (all current ones), behavior is unchanged.

d-chambers:
Failing job is related to timeout on macos with remote reading. Will open different PR to fix this.

PR #751, opened 2026-07-11
Spool index backend

d-chambers:
Replaces DASCore's PyTables/HDF5 spool index with a relational SQLite metadata engine and unifies every spool type (in-memory, directory, and their unions) on top of it. Selection, candidate identification, chunk planning, and union now go through one lazy `PatchCatalog`, so the same pushdown-capable semantics apply whether patches come from files or memory. The old `ChunkManager` and PyTables index are removed.

## Motivation

The old index stored a flat, per-directory PyTables table and only served directory spools; in-memory and union spools each had their own ad-hoc metadata paths, and chunking went through a separate `ChunkManager`. Behavior diverged across spool types, queries kept everything in memory, and scaling to large archives was awkward. A relational index with a real query planner lets one code path answer counts, selections, and chunk plans lazily and push work down to the database.

## What changed

- **Relational backend (`io/index/`).** A normalized seven-table schema (sources, patches, attrs, coordinate definitions, links) with keys, value-kind checks, and cascade foreign keys, validated before any mutation. Coordinate definitions separate an exact fingerprint from a summary-only dedup key. Nullable nanosecond columns are fetched without a float64 round trip (ns epochs exceed 2**53). Ingest normalizes unit-bearing values to canonical base SI, and the query builder compiles the selector spec to SQL (attrs exact, regex as candidate + residual, coordinate predicates candidacy-only with exact re-trim at load).
- **One engine for every spool.** `PatchCatalog` owns the backend, a resolver, and the directory synchronizer; directory and memory spools share it, so lazy selection, pushdown, and chunk planning are identical across types. `spool + spool` unions catalogs table-to-table. In-memory patches carry an eager instance identity, giving spools set semantics by patch identity.
- **Chunk planning.** `ChunkManager` is replaced by an inspectable `ChunkPlan`, exposed publicly as `Spool.chunk_plan(...)` with the same arguments as `chunk`.
- **Unit-aware selection.** Numeric coordinate summaries are stored in canonical SI, so a bare range means SI regardless of native units and the exact per-patch trim defers its representation until each patch is known — correct across mixed unitful/unitless archives. Coordinate selection accepts a `(start, stop)` range or a boolean mask; scalar/membership selectors are rejected eagerly.
- **Relocations/removals.** PyTables dropped from `pyproject.toml`/`environment.yml`; the chunk planner, path classifiers, and relative-selection helpers moved to `utils/`.

## Summary by CodeRabbit

* **New Features**
  * Added SQLite-backed directory indexing with incremental updates and a catalog-based patch selection workflow.
  * Expanded spool functionality: richer `select` options (`_attrs`/`_coords`, `samples`, `relative`), `chunk_plan`, and `chunk(..., group, missing_dim)`, plus spool-union support (`spool + spool`).
  * Added config controls for chunk grouping (`sampling_group_tolerance`, `groupby_attrs`).

* **Bug Fixes**
  * Improved patch identity/serialization behavior and corrected selection/trim exactness, including unit-aware and regex-chaining query handling.
  * Refined chunk/merge validation and edge-case behavior.

* **Documentation**
  * Updated spool index/selection/chunking docs and directory indexer tutorial for the SQLite model.

* **Chores**
  * Migrated HDF5 handling/tests from PyTables to h5py; updated related configs and fixtures.

## Changelog

- changed **breaking**: the spool hierarchy collapses to one concrete `dc.Spool`; `MemorySpool`, `DirectorySpool`, `FileSpool`, the `dascore.clients` package, and the `select_kwargs` parameter are removed — use `dc.spool(...)` and `spool.has_live_patches`.
- changed **breaking**: `Spool.update()` is allowed only on a root spool; any derived spool (select, slice, sort, chunk, `+`) raises.
- changed **breaking**: `Spool.select` no longer accepts coordinate boolean masks — use `spool.map(lambda p: p.select(dim=mask))`; `spool[bool_array]` still selects membership and `samples=True` ranges still apply per patch.
- changed **breaking**: directory indexes use a seven-table SQLite schema in `.dascore_index.sqlite3`; the PyTables index is gone, so existing indexes must be deleted and rebuilt.
- changed **breaking**: SQLite supports concurrent readers and one serialized writer on local filesystems; network filesystems with unreliable locking are not supported.
- removed **breaking**: `PyTablesReader`, `PyTablesWriter`, and the `HDF5Reader`/`HDF5Writer` aliases from `dascore.utils.hdf5`; use `H5Reader`/`H5Writer`.
- changed: spools present patches in a defined order — construction order for patch lists, time order for directories, and rows without a value for the ordering key sort last — and equality compares effective contents rather than backing.
- changed: `a + b` bakes each operand's pending trims and sort order into the union instead of dropping them, including per-patch time order for interleaved multi-patch files.
- changed: `Spool.sort(...)` accepts any coordinate of the spool, not just `time`/`distance`, and contiguous descending-coordinate patches now merge under `chunk(dim=None)`.
- changed: chunking is defined on dimensions only and is unit-aware — compatible spellings such as metres and feet merge, incompatible dimensionalities can never plan together — and chunking a restructured spool along a *different* dimension plans over its current patches, while re-chunking the same dimension still re-plans from the original members.
- changed: `Spool.select` validates unknown names and quantity dimensionality, composes chained regexes with AND, supports explicit `_attrs`/`_coords` namespaces, and skips (with a warning) attribute values whose units are dimensionally incompatible across files instead of failing the index update.


d-chambers:
@coderabbitai review

d-chambers:
@coderabbitai review

d-chambers:
@coderabbitai review

d-chambers:
@coderabbitai review

PR #753, opened 2026-07-16
implemented gap-sensistive waterfall plot

andreas-wuestefeld:
The waterfall plot is very fast by using the imshow functionality. This however assumes evenly sampled axis. DAScore however also support monotonic, but uneven, axis. 

One example of such data is when files of a spool are processed as continuous data and a transformed data-product is amalgamated into a compressed axis. A simple example would be the absolute maximum of each channel within each file. Data gaps due to missing files would not plot properly, as imhow stretches dataover the gaps. this also results in the axis ticks not aligning correctly with the cell position.

This PR fixes the issue. Imshow is still used for regular arrays. However, if one or more axis have gaps, pcolormesh is used. The user can choose to fill gaps with a selectable color, or to correctly stretch the last sample before the gap.

## Checklist

I have (if applicable):

- [x ] documented the new feature with docstrings and/or appropriate doc page.
- [ x] included tests. See [testing guidelines](https://dascore.org/contributing/testing.html).

## Summary by CodeRabbit

* **New Features**
  * Waterfall plots now render irregular, monotonic coordinate grids using mesh-based rendering.
  * Added gap-aware visualization with optional `gap_color` and tunable `gap_factor`.
* **Bug Fixes**
  * Preserved colorbars, scaling, and axis formatting when gaps are present.
  * Improved behavior for datetime/duration axes, including consistent time tick placement across gaps.
* **Tests**
  * Added coverage for gap detection, gap band masking, renderer selection, and invalid `gap_factor` handling.

### DIFF
--- a/dascore/viz/waterfall.py
+++ b/dascore/viz/waterfall.py
@@ -329,6 +329,9 @@ def waterfall(
         'auto' (default) selects a suitable interpolation stage automatically.
         See matplotlib's imshow documentation for more details. This option
         does not apply when ``pcolormesh`` is used.
+    gap_factor
+        Coordinate intervals larger than this factor times the median interval
+        are treated as gaps. Must be greater than 1.
     gap_color
         Matplotlib color used to display gaps in irregular dimension
         coordinates. When a color is provided, a masked row or column is
@@ -337,9 +340,6 @@ def waterfall(
         without expanding the data matrix. This option only applies when
         ``pcolormesh`` is used. Existing masked or NaN data receive the same
         color as coordinate gaps.
-    gap_factor
-        Coordinate intervals larger than this factor times the median interval
-        are treated as gaps. Must be greater than 1.
     log
         If True, visualize the common logarithm of the absolute values of patch data.
         To avoid log(0), the abs(array) is cast to float64 and a small value
--- a/dascore/viz/waterfall.py
+++ b/dascore/viz/waterfall.py
@@ -6,6 +6,7 @@ from collections.abc import Sequence
 from typing import Literal
 
 import matplotlib as mpl
+import matplotlib.dates as mdates
 import matplotlib.pyplot as plt
 import numpy as np
 
@@ -28,7 +29,14 @@ def _validate_scale_type(scale_type):
     """Validate that scale_type is either 'relative' or 'absolute'."""
     valid_types = {"absolute", "relative"}
     if scale_type not in valid_types:
-        msg = f"scale_type must be one of {valid_types}, " f"but got '{scale_type}'"
+        msg = f"scale_type must be one of {valid_types}, but got '{scale_type}'"
+        raise ParameterError(msg)
+
+
+def _validate_gap_factor(gap_factor):
+    """Validate the factor used to identify coordinate gaps."""
+    if not np.isfinite(gap_factor) or gap_factor <= 1:
+        msg = "gap_factor must be a finite number greater than 1"
         raise ParameterError(msg)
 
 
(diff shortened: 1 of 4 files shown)


dascore/viz/waterfall.py
```diff
@@ -157,6 +165,111 @@ def _get_waterfall_colormap(patch, cmap=None):
...
 
 
+def _get_plot_coord(coord):
+    """Return coordinate values in units understood by matplotlib meshes."""
+    values = np.asarray(coord)
+    if np.issubdtype(values.dtype, np.datetime64):
+        return mdates.date2num(values)
+    if np.issubdtype(values.dtype, np.timedelta64):
+        return values / np.timedelta64(1, "s")
+    return values
```
d-chambers: The current _get_plot_coord handles three cases:

  1. Numeric coordinates: returns them unchanged.
  2. datetime64: calls mdates.date2num.
  3. timedelta64: converts to seconds manually.

  With Matplotlib ≥3.10 (DASCore's lowest supported version)

  - pcolormesh already converts datetime64 coordinates into Matplotlib date numbers.
  - DASCore’s existing to_float already converts timedelta64 values to seconds.
  - Numeric values require no conversion.

So we really just need

```x = to_float(x) if is_timedelta64(x) else x```

where this function is called.

File: dascore/viz/waterfall.py [hunk 2]
```diff
@@ -157,6 +165,111 @@ def _get_waterfall_colormap(patch, cmap=None):
...
+
+
+def _get_gap_edges(values, gap_factor):
+    """Return cell edges and locations of gaps in monotonic center values."""
+    values = np.asarray(values)
+    if len(values) == 1:
+        return np.asarray([values[0] - 0.5, values[0] + 0.5]), np.array([])
+    diffs = np.diff(values)
+    abs_diffs = np.abs(diffs)
+    representative_step = np.median(abs_diffs)
```
d-chambers: So this will assume a single sampling rate correct? Maybe worth noting in the doc string because it is possible to have a concatenation of several RangeCoords with different sampling rates. This would not handle that case very well.

dascore/viz/waterfall.py
```diff
@@ -157,6 +165,111 @@ def _get_waterfall_colormap(patch, cmap=None):
...
+            dim_edges = (
+                np.concatenate(
+                    (
+                        [values[0] - (values[1] - values[0]) / 2],
+                        values[:-1] + np.diff(values) / 2,
+                        [values[-1] + (values[-1] - values[-2]) / 2],
+                    )
+                )
+                if len(values) > 1
+                else dim_edges
```
d-chambers: This could use a well-named helper function.

d-chambers (changes requested):
A few suggestions.

My main concern with this is that some of the gap detection/handling logic doesnt really belong in waterfall.py because it has nothing to do with plotting. Could you please move it utils/coords.py, or perhaps make a new utils/gap.py? I would keep its functions private, as the dev branch will eventually supercede how gaps are dealt with. Maybe add a few dedicated tests as well for that module.

andreas-wuestefeld:
Thanks for the excellent suggestions! 
I am actually worried that the introduction of pcolormesh will make the rather clean waterfall function a bit crowded, but I can't think of a better way to handle data gaps

d-chambers (approved):
Looks good to me. We might come back later and tweak the gap code, but this does the job for now. 

I agree waterfall is getting a bit too crowded, but it's probably better to have an extra param or two rather than a `waterfall_with_gaps` function.

PR #754, opened 2026-07-17
Expose exact coordinates through scan payloads

d-chambers:
## Description

- add `dc.scan_payloads(...)` as a sibling to `dc.scan(...)`, returning raw `ScanPayload` dictionaries with full `CoordManager` objects and dispatcher-owned source provenance
- standardize `FiberIO.scan(..., snap: bool = True)` while keeping `dc.scan(...)` and `dc.scan_to_df(...)` output and signatures unchanged
- make `snap=False` preserve stored coordinate values through zero-tolerance `CoordSegmented.from_array(...)` construction in stored-array formats
- document the scan payload/snap contract and add public API, provenance, forwarding, common conformance, and Terra15 regression coverage

## Format audit

| Format | Coordinate source | `snap=False` behavior |
| --- | --- | --- |
| DASDAE | serialized range metadata or stored coordinate arrays | exact construction for array-backed coordinates |
| DASHDF5 | stored time, channel, trace, and spatial arrays | exact construction |
| Febus BSL/MTX/T1 | stored time/distance and auxiliary arrays | exact construction |
| GDR | stored `DasTimeArray` | exact construction |
| H5Simple | stored coordinate arrays | exact construction |
| NetCDF CF | stored xarray coordinate values | exact construction for 1D monotonic coordinates; other arrays remain value-backed |
| Neubrex RFS | stored Unix timestamp array | exact construction |
| OptoDAS | header-defined time plus stored channel array | exact construction for distance; time is already exact metadata |
| PRODML | stored per-sample time arrays plus header-defined distance | exact construction for time |
| Terra15 | stored GPS/Posix time array plus header-defined distance | exact construction for time |
| APSensing, DASVader, Febus A1/G1 CSV, MiniSEED, Neubrex DAS, SEG-Y, Sentek, Silixa H5, Sintela Binary, SR4731, TDMS, XMLBinary | start/step/count or equivalent header metadata | documented no-op |
| Pickle, RSF, WAV, and other default-scan formats | coordinates from the fully loaded patch | best-effort no-op; payload values match the loaded patch |

## Review notes

- Exact array construction uses `tolerance=0`. This keeps every stored value unchanged while still returning `CoordRange` for truly uniform arrays and `CoordSegmented` for exact monotonic runs separated by seams.
- Standardizing read-side `snap`/`snap_dims` naming remains a follow-up.
- A declarative capability flag such as `exact_scan_coords` remains deferred until planner integration needs it.

## Testing

- `pre-commit run --all`
- `pytest tests -q` — 7,368 passed, 87 skipped, 3 xfailed

## Changelog

- added: `dc.scan_payloads(...)` returns raw scan payloads with full coordinate managers and source provenance without loading data arrays.
- changed: `FiberIO.scan(...)` standardizes `snap: bool = True`; `snap=False` requests exact stored coordinate values, and a jittery per-sample array falls back to a plain monotonic coordinate rather than exploding into per-sample segments.
- changed: some formats gained coordinate units on their default (`snap=True`) scan and read output — Terra15 exact times, OptoDAS and NetCDF distances.
- changed: `ScanPayload` carries optional `source_path`, `source_format`, and `source_version`, populated by `dc.scan_payloads(...)` rather than by format implementations.

## Checklist

I have (if applicable):

## Summary by CodeRabbit

* **New Features**
  * Added/exposed `scan_payloads(...)` to retrieve raw scan payloads with full `CoordManager` coordinates and optional source provenance (`source_path`/`source_format`/`source_version`).

* **User-Facing Changes**
  * Standardized `scan(..., snap: bool = True)` across supported formats; `snap=False` requests exact stored per-sample coordinate values when available.

* **Documentation**
  * Updated scan-payload and IO docs to describe the payload contract and the `snap` vs exact-coordinate behavior.

* **Tests**
  * Expanded end-to-end coverage for payload completeness, provenance attachment, and `snap=False` exact-value conformance.


d-chambers:
@coderabbitai review

dascore/io/utils.py
```diff
@@ -0,0 +1,12 @@
+"""Shared utilities for IO implementations."""
+
+from dascore.core.coords import BaseCoord, CoordSegmented, get_coord
+from dascore.exceptions import CoordError
+
+
+def get_exact_coord(values, units=None) -> BaseCoord:
+    """Return an exact coordinate, including for non-monotonic values."""
+    try:
+        return CoordSegmented.from_array(values, tolerance=0, units=units)
```
d-chambers, line 26: This is where the segment explosion originates (see review body for numbers). With `tolerance=0`, `from_array` never raises on a jittery-but-monotonic array — it happily returns a `CoordSegmented` with ~n/2 segments — so the `CoordError` fallback below only catches non-monotonic/duplicate/multidim input and never the case that actually needs it.

Suggest an explicit degeneracy guard, e.g.:

```python
def get_exact_coord(values, units=None) -> BaseCoord:
    """Return an exact coordinate, including for non-monotonic values."""
    try:
        out = CoordSegmented.from_array(values, tolerance=0, units=units)
    except CoordError:
        return get_coord(data=values, units=units)
    # Jittery arrays segment into ~n/2 runs; a plain monotonic array is
    # smaller, faster to build, and equally exact.
    if len(getattr(out, "segments", ())) > _MAX_SEGMENT_FRACTION * len(values):
        return get_coord(data=values, units=units)
    return out
```

`get_coord(data=...)` is exact for these inputs — its tolerant inference only collapses to `CoordRange` when the array really is near-uniform, which is the case where segmentation would have produced one segment anyway. A pre-check inside `from_array` that avoids building all the segments first would be better still, since the cost is ~120 µs/segment.

dascore/io/terra15/utils.py
```diff
@@ -107,7 +108,8 @@ def _scan_terra15(h5_fi, data_node, extras=None) -> list[ScanPayload]:
 def _get_raw_time_coord(data_node):
     """Read the time from the data node and return it."""
     time = _get_time_node(data_node)[:]
-    return get_coord(data=to_datetime64(time))
+    values = to_datetime64(time)
+    return get_exact_coord(values, units="s")
```
d-chambers, line 112: Two behavior changes on the **read** path, not just scan — `_get_raw_time_coord` is what `_read_terra15(..., snap_dims=False)` calls:

1. `get_exact_coord` here is what makes `dc.read(terra15_path, snap_dims=False)` 5x slower and return a 1001-segment `CoordSegmented` instead of a `CoordMonotonicArray` (numbers in the review body). Fixing `get_exact_coord` resolves this.
2. `units="s"` is newly attached to a `datetime64` coord that previously had none. It matches the `snap_dims=True` branch above so it's probably the intended normalization, but it changes existing `dc.read` output and isn't in the changelog. Worth a line there either way.

dascore/io/core.py
```diff
@@ -1338,7 +1567,7 @@ def is_directory_format(path) -> bool:
         return False
     try:
         get_format(path)
-    except Exception:
+    except UnknownFiberFormatError:
```
d-chambers: This narrowing looks unrelated to the scan-payload work, and it makes directory indexing more fragile. `is_directory_format` is called from `indexer.py` for every directory encountered during traversal, and `get_format` dispatches to arbitrary `FiberIO.get_format` implementations — an unreadable directory (`PermissionError`) or a formatter that raises something unexpected on a weird path will now abort the whole index instead of the directory simply not being a scan unit.

The blanket `except Exception` was admittedly too broad, but I'd rather see the exception set widened (`UnknownFiberFormatError`, `OSError`, at minimum) than tightened to one type — or this split into its own PR where the traversal-robustness tradeoff can be argued on its own merits.

dascore/io/netcdf/core.py
```diff
@@ -149,6 +154,14 @@ def scan(self, resource: H5Reader, **kwargs) -> list[ScanPayload]:
             )
         ]
 
+    @staticmethod
+    def _get_scan_coord(coord, snap=True):
+        """Return raw xarray values or an exact monotonic coordinate."""
+        values = coord.values
+        if snap or np.ndim(values) != 1:
+            return values
+        return get_exact_coord(values, units=coord.attrs.get("units"))
```
d-chambers: Minor asymmetry: the `snap=False` path attaches `coord.attrs.get("units")`, but the `snap=True` path returns bare `values` with no units, so the two modes produce coords that differ in more than exactness. Optodas got the equivalent fix in this PR (`get_coord(data=distance, units=unit)` on the snap branch); worth doing the same here so `snap` only controls exactness.

Separately — that optodas units change is itself a user-visible change to `snap=True` scan output and isn't mentioned in the changelog.

dascore/io/prodml/utils.py
```diff
@@ -128,14 +129,17 @@ def get_distance_units(attrs):
...
-def _get_time_coord(node):
+def _get_time_coord(node, snap=True):
     """Get the time information from a Raw node."""
     time_names = [x for x in node.keys() if x.endswith("DataTime")]
     assert len(time_names) == 1, f"Found bad time information in prodml {node=}"
     time_name = time_names[0]
     time_array = node[time_name]
     array_len = len(time_array)
     assert array_len > 0, "Missing time array in ProdML file."
+    if not snap:
```
d-chambers, line 140: Worth a docs note: ProdML's `read` has no `snap`/`snap_dims` parameter, so `dc.scan_payloads(path, snap=False)` returns the exact stored per-sample array while `dc.read(path)` always returns the header-derived `PartStartTime`/`PartEndTime` range. The conformance test in `test_common_io.py` skips the value comparison for formats without an exact read, so nothing catches the divergence.

That's arguably correct — read is the lossy one — but a caller comparing payload coords to loaded patch coords will be surprised. The `snap` contract section in `new_format.qmd` would be a good place to say that `snap=False` describes stored values, not necessarily what `read` returns.

d-chambers (commented):
Nice shape overall — splitting `_iter_scan_results` out so `scan` and `scan_payloads` share one traversal is the right factoring, the validation helper gives a real contract boundary, and the format audit table in the description is genuinely useful. The docs work (coordinate_internals, new_format snap section) is a clear improvement.

My main concern is `get_exact_coord`. `CoordSegmented.from_array(..., tolerance=0)` degenerates badly on jittery stored timestamps — which is exactly what GPS-stamped DAS time arrays look like — and because `terra15._get_raw_time_coord` is on the **read** path, this lands as a performance and representation regression on an existing public API, not just on the new one.

Measured on this branch vs. `dev`, using the checked-in trimmed terra15 test file (3,620 samples):

| | dev | this branch |
|---|---|---|
| `dc.read(path, snap_dims=False)` | 0.154 s | 0.812 s |
| resulting time coord | `CoordMonotonicArray` | `CoordSegmented`, **1001 segments** |

And synthetic monotonic arrays with ±3 ns jitter on a 1 µs step:

| n | `get_exact_coord` | `get_coord(data=...)` | segments |
|---|---|---|---|
| 1,000 | 0.125 s | 0.0003 s | 122 |
| 10,000 | 1.21 s | 0.0007 s | 1,139 |
| 50,000 | 6.06 s | 0.0018 s | 5,512 |
| 100,000 | 11.6 s | — | ~11,000 |

Roughly 120 µs per segment, so a 1M-sample file extrapolates to ~2 minutes. Both problems come from the same place: `from_array`'s run detection (a difference joins a run only when it equals a neighbor) means jitter turns nearly every sample into its own seam, so you get an n/2-segment object that is slower to build, larger in memory, and less useful than the `CoordMonotonicArray` it replaced.

The clean cases are fine — uniform-with-one-gap gives 2 segments as intended — so the fix is a guard rather than a redesign: bail out to `get_coord(data=...)` when the detected segment count exceeds some small fraction of `len(values)`. That preserves the real win (exact values plus queryable seams for genuinely piecewise-uniform data) without the pathological path. Ideally detect this before materializing all the segments so the cost is bounded too. Worth a test with a jittery array asserting the coord type and that values round-trip exactly.

Three smaller items inline. Everything else — the validation helper, provenance ownership, forwarding tests, conformance test — reads well to me.

dascore/io/utils.py
d-chambers: Addressed in 0dca954f. Added a `_is_over_segmented` pre-check to `get_exact_coord`: for arrays large enough for the cost to matter, it cheaply predicts the segment count (mirroring `from_array`s run detection) and falls back to `get_coord(data=...)` when segments would exceed one per ten samples. I used a pre-check rather than the post-hoc `len(out.segments)` guard so the degenerate path never materializes the segments at all (the cost is ~120 µs/segment). Genuinely piecewise-uniform arrays still segment and keep their queryable seams. Regression tests (jittery array stays monotonic + round-trips exactly; piecewise-uniform stays segmented) added in aea075c9.

dascore/io/terra15/utils.py
d-chambers: Addressed in 0dca954f. The `get_exact_coord` guard resolves the read-path regression: `dc.read(terra15, snap_dims=False)` is back to ~0.12 s and returns a `CoordMonotonicArray` again instead of a 1001-segment `CoordSegmented`. The `units="s"` normalization is intentional (matches the `snap_dims=True` branch) and is now called out in the changelog.

dascore/io/core.py
d-chambers: Addressed in 0dca954f. Widened to `except (UnknownFiberFormatError, OSError)` so an unreadable directory (e.g. `PermissionError`) is simply not treated as a scan unit rather than aborting an index traversal, while genuinely unexpected errors still surface. Kept it in this PR since it sits on the scan path; happy to split it out if you would rather.

dascore/io/netcdf/core.py
d-chambers: Addressed in 0dca954f. The `snap=True` NetCDF branch now returns `get_coord(data=values, units=...)` too, so `snap` controls only exactness and both modes carry units. The changelog note now also mentions the `snap=True` units change for OptoDAS, NetCDF, and Terra15.

dascore/io/prodml/utils.py
d-chambers: Addressed in 0dca954f. Added a paragraph to the `snap` contract section of `new_format.qmd` stating that `snap=False` describes the values a format *stores*, which are not necessarily what `read(...)` returns (ProdML reconstructs its time coord from `PartStartTime`/`PartEndTime`), and noting the common-IO conformance test only compares scan vs read for formats whose `read` also accepts `snap`/`snap_dims`.

PR #762, opened 2026-07-22
Add PRODML write support

d-chambers:
## Summary

- add standalone PRODML 2.1-compatible HDF5 writing for one raw time-by-distance Patch
- derive acquisition, facility, units, sampling, and optional measurement metadata from Patch attrs and coordinates
- validate representability, preserve dimension order and dtype, round timestamps safely to microseconds, and replace existing targets only after validation
- clarify that full PRODML 2.3 EPC/XML packaging is outside this writer's scope

## Review

Claude ultrareview checked the branch against `dev`. Its one verified finding—invalid optional-measure units escaping the graceful omission path—was fixed with regression coverage.

## Validation

- `pytest -q tests/test_io/test_prodml` — 42 passed
- `pytest -q tests/test_io/test_io_core.py tests/test_io/test_common_io.py` — 1006 passed, 149 skipped
- `pre-commit run --all-files` — passed
- all 23 applicable official PRODML 2.3 DAS HDF5 examples still detect as PRODML 2.1

## Summary by CodeRabbit

* **New Features**
  * Added ability to write a single raw time-by-distance patch as a standalone PRODML 2.1-compatible HDF5 file via `dc.write`, with version-specific behavior and forwarded options support.
  * Added strict validation for patch shape, coordinates sampling/alignment, supported dtypes, units, timestamps, and required metadata (including UUIDs).
  * Implemented HDF5 string encoding utility for writing fixed-length UTF-8 strings.
* **Bug Fixes**
  * Improved PRODML metadata normalization/handling (including “UNKNOWN” units/type cleanup and safer version/fingerprint behavior).
* **Documentation**
  * Clarified current PRODML support level (write-only 2.1; no full 2.3 EPC/XML packaging).
* **Tests**
  * Expanded standalone PRODML write/read, precision rounding warnings, replacement semantics, and extensive negative/validation coverage.

## Changelog

- added: PRODML 2.1 write support for one raw time-by-distance patch as a standalone HDF5 file. Full PRODML 2.3 conformance requires EPC/XML packaging, which DASCore does not write.

### DIFF
--- a/dascore/io/prodml/utils.py
+++ b/dascore/io/prodml/utils.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import warnings
 from collections.abc import Iterator
 from dataclasses import dataclass
-from uuid import UUID, uuid4
+from uuid import uuid4
 
 import h5py
 import numpy as np
@@ -44,6 +44,8 @@ _ROOT_ATTRS = {
     "GaugeLengthUnits": "gauge_length_units",
     "GaugeLength.uom": "gauge_length_units",
     "AcquisitionId": "acquisition_id",
+    "FacilityId": "facility_id",
+    "ServiceCompanyName": "service_company_name",
     "schemaVersion": "schema_version",
 }
 
@@ -115,6 +117,15 @@ def _get_prodml_version_str(hdf_fi) -> str:
     # in some prodml schemaVersion is str, in other float; this handles both.
     version_str = str(unbyte(attrs["schemaVersion"]))
     return version_str
+
+
+def _get_root_attrs(attrs):
+    """Return normalized attributes from an Acquisition group."""
+    out = maybe_get_items(attrs, _ROOT_ATTRS)
+    if "facility_id" in out:
+        values = tuple(unbyte(x) for x in np.atleast_1d(out["facility_id"]))
+        out["facility_id"] = values[0] if len(values) == 1 else values
+    return out
 
 
 def _get_distance_coord(acq):
@@ -357,23 +368,22 @@ def _prepare_prodml_write(spool):
     )
     attrs = patch.attrs
     acquisition_id = str(attrs.acquisition_id or uuid4())
-    try:
-        acquisition_id = str(UUID(acquisition_id))
-    except (ValueError, TypeError, AttributeError) as exc:
-        raise PatchError("ProdML acquisition_id must be a valid UUID.") from exc
 
     facility = attrs.get("facility_id") or attrs.station or "UNKNOWN"
     service_company = attrs.get("service_company_name") or "UNKNOWN"
     data_units = get_quantity_str(attrs.data_units) or "UNKNOWN"
-    string_values = (facility, service_company, data_units)
-    if not all(isinstance(x, str) and 0 < len(x) <= 64 for x in string_values):
-        msg = "ProdML facility, service company, and data unit must be String64 values."
-        raise PatchError(msg)
+    string_values = {
+        "FacilityId": facility,
+        "ServiceCompanyName": service_company,
+        "RawDataUnit": data_units,
+    }
+    for name, value in string_values.items():
+        if not isinstance(value, str) or not 0 < len(value.encode("utf-8")) <= 64:
+            msg = f"ProdML {name} must contain 1 to 64 UTF-8 bytes (String64)."
+            raise PatchError(msg)
 
     start_time, end_time = _iso_time(times[0]), _iso_time(times[-1])
     output_rate = 1_000_000.0 / time_step_us
-    pulse_rate = _get_measure(attrs, "pulse_rate", "pulse_rate_units", "Hz")
-    pulse_rate = pulse_rate or (output_rate, "Hz")
     num_loci = len(patch.get_coord("distance"))
     file_uuid, acquisition_uuid, raw_uuid = (str(uuid4()) for _ in range(3))
     acquisition_attrs = {
@@ -385,8 +395,6 @@ def _prepare_prodml_write(spool):
         "MinimumFrequency": 0.0,
         "MinimumFrequency.uom": _fixed_string("Hz"),
         "NumberOfLoci": num_loci,
-        "PulseRate": pulse_rate[0],
-        "PulseRate.uom": _fixed_string(pulse_rate[1]),
         "ServiceCompanyName": _fixed_string(service_company),
         "SpatialSamplingInterval": distance_step,
         "SpatialSamplingInterval.uom": _fixed_string(distance_units),
@@ -396,6 +404,7 @@ def _prepare_prodml_write(spool):
         "uuid": _fixed_string(acquisition_uuid),
     }
     for name, units_name, target, hdf_name in (
+        ("pulse_rate", "pulse_rate_units", "Hz", "PulseRate"),
         ("pulse_width", "pulse_width_units", "s", "PulseWidth"),
         ("gauge_length", "gauge_length_units", "m", "GaugeLength"),
     ):
@@ -525,7 +534,7 @@ def _yield_prodml_attrs_coords(
     """Scan a prodML file, return metadata."""
     acq = fi["Acquisition"]
     # Get the information common to all from root attributes.
-    base_info = maybe_get_items(acq.attrs, _ROOT_ATTRS)
+    base_info = _get_root_attrs(acq.attrs)
     base_info.update(extras if extras is not None else {})
     d_coord = _get_distance_coord(acq)
     # Iterate the raw and processed data and return results in a list.
@@ -558,7 +567,7 @@ def _read_prodml(fi, distance=None, time=None, source_patch_id=None):
     """Read the prodml values into a patch."""
     out = []
     acq = fi["Acquisition"]
-    base_info = maybe_get_items(acq.attrs, _ROOT_ATTRS)
+    base_info = _get_root_attrs(acq.attrs)
     d_coord = _get_distance_coord(acq)
     source_patch_ids = _normalize_source_patch_ids(source_patch_id)
     for info in _yield_data_nodes(fi):
--- a/dascore/io/prodml/utils.py
+++ b/dascore/io/prodml/utils.py
@@ -2,8 +2,10 @@
 
 from __future__ import annotations
 
+import warnings
 from collections.abc import Iterator
 from dataclasses import dataclass
+from uuid import UUID, uuid4
 
 import h5py
 import numpy as np
@@ -11,7 +13,9 @@ import numpy as np
 import dascore as dc
 from dascore.constants import VALID_DATA_TYPES
 from dascore.core.coords import get_coord
+from dascore.exceptions import InvalidSpoolError, PatchError, UnitError
 from dascore.io.utils import get_exact_coord
+from dascore.units import convert_units, get_quantity_str
 from dascore.utils.io import _normalize_source_patch_ids
 from dascore.utils.misc import iterate, maybe_get_items, register_func, unbyte
 from dascore.utils.models import UnitQuantity, UTF8Str
@@ -19,9 +23,9 @@ from dascore.utils.models import UnitQuantity, UTF8Str
 # --- Getting format/version
 
 _EXPECTED_ATTRS = (
-    "PulseRate",
-    "PulseWidth",
     "NumberOfLoci",
+    "SpatialSamplingInterval",
+    "StartLocusIndex",
     "schemaVersion",
     "uuid",
 )
@@ -39,6 +43,7 @@ _ROOT_ATTRS = {
     "GaugeLengthUnit": "gauge_length_units",
     "GaugeLengthUnits": "gauge_length_units",
     "GaugeLength.uom": "gauge_length_units",
+    "AcquisitionId": "acquisition_id",
     "schemaVersion": "schema_version",
 }
 
@@ -105,9 +110,11 @@ def _get_prodml_version_str(hdf_fi) -> str:
         return ""
     attrs = acquisition.attrs
     is_prodml = set(_EXPECTED_ATTRS).issubset(set(attrs))
+    if not is_prodml:
+        return ""
     # in some prodml schemaVersion is str, in other float; this handles both.
     version_str = str(unbyte(attrs["schemaVersion"]))
-    return version_str if is_prodml else ""
+    return version_str
 
 
 def _get_distance_coord(acq):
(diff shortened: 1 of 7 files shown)


d-chambers:
@codex review

dascore/io/prodml/utils.py
```diff
@@ -216,12 +223,244 @@ def _get_data_unit_and_type(node):
...
+    data_units = get_quantity_str(attrs.data_units) or "UNKNOWN"
+    string_values = (facility, service_company, data_units)
+    if not all(isinstance(x, str) and 0 < len(x) <= 64 for x in string_values):
+        msg = "ProdML facility, service company, and data unit must be String64 values."
+        raise PatchError(msg)
+
+    start_time, end_time = _iso_time(times[0]), _iso_time(times[-1])
+    output_rate = 1_000_000.0 / time_step_us
+    pulse_rate = _get_measure(attrs, "pulse_rate", "pulse_rate_units", "Hz")
+    pulse_rate = pulse_rate or (output_rate, "Hz")
```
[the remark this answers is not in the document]
d-chambers: Addressed in 29c17e0e. `PulseRate` now uses the optional-measure path and is omitted unless valid `pulse_rate` metadata and units are present; `OutputDataRate` remains derived only from the time cadence. Tests cover both omission and an explicit 50 Hz pulse rate.

dascore/io/prodml/utils.py
```diff
@@ -216,12 +223,244 @@ def _get_data_unit_and_type(node):
...
+    try:
+        acquisition_id = str(UUID(acquisition_id))
+    except (ValueError, TypeError, AttributeError) as exc:
+        raise PatchError("ProdML acquisition_id must be a valid UUID.") from exc
+
+    facility = attrs.get("facility_id") or attrs.station or "UNKNOWN"
+    service_company = attrs.get("service_company_name") or "UNKNOWN"
+    data_units = get_quantity_str(attrs.data_units) or "UNKNOWN"
+    string_values = (facility, service_company, data_units)
+    if not all(isinstance(x, str) and 0 < len(x) <= 64 for x in string_values):
```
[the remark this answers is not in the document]
d-chambers: Addressed in 29c17e0e. String64 validation now checks `len(value.encode("utf-8"))` for each named HDF field and reports the offending field. The regression covers both 65 ASCII bytes and 33 two-byte characters.

dascore/io/prodml/utils.py
```diff
@@ -216,12 +223,244 @@ def _get_data_unit_and_type(node):
...
+    times, time_step_us = _round_times_to_microseconds(patch.get_coord("time"))
+    distance_step, distance_units, start_locus = _get_distance_info(
+        patch.get_coord("distance")
+    )
+    attrs = patch.attrs
+    acquisition_id = str(attrs.acquisition_id or uuid4())
+    try:
+        acquisition_id = str(UUID(acquisition_id))
+    except (ValueError, TypeError, AttributeError) as exc:
+        raise PatchError("ProdML acquisition_id must be a valid UUID.") from exc
```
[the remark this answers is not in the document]
d-chambers: Addressed in 29c17e0e. The writer now preserves any non-empty `PatchAttrs.acquisition_id` verbatim and calls `uuid4()` only when the attribute is empty. `test_acquisition_id_is_preserved` covers `run-001`, while the missing-metadata test verifies that an absent ID receives a generated UUID.

dascore/io/prodml/utils.py
```diff
@@ -39,6 +43,7 @@
     "GaugeLengthUnit": "gauge_length_units",
     "GaugeLengthUnits": "gauge_length_units",
     "GaugeLength.uom": "gauge_length_units",
+    "AcquisitionId": "acquisition_id",
```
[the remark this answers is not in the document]
d-chambers: Addressed in 29c17e0e. The root attribute map now restores `FacilityId` to `facility_id` and `ServiceCompanyName` to `service_company_name`; normalization collapses the writer singleton `FacilityId` array to a string while preserving multiple values as a tuple. Round-trip and station-fallback regression tests cover both paths.

dascore/io/prodml/utils.py
```diff
@@ -216,12 +234,254 @@ def _get_data_unit_and_type(node):
...
         out["data_type"] = clean if clean in VALID_DATA_TYPES else ""
     return out
 
 
+# --- Writing ProdML files.
+
+
+def _fixed_string(value):
+    """Return a fixed-length HDF5 byte string."""
+    return np.bytes_(str(value).encode())
```
d-chambers: This should go in the hdf5 utils

dascore/io/prodml/utils.py
```diff
@@ -216,12 +234,254 @@ def _get_data_unit_and_type(node):
...
+
+
+def _fixed_strings(values):
+    """Return an array of fixed-length HDF5 byte strings."""
+    encoded = [str(value).encode() for value in values]
+    return np.asarray(encoded, dtype=f"S{max(map(len, encoded))}")
```
d-chambers: Same for this. Probably better to combine to one function that optionally takes a string or sequence of strings. Use utils.iterate

dascore/io/prodml/utils.py
```diff
@@ -216,12 +234,254 @@ def _get_data_unit_and_type(node):
...
+        msg = "ProdML writing requires a nonempty numeric distance coordinate."
+        raise PatchError(msg)
+    values = values.astype(np.float64)
+    if not np.all(np.isfinite(values)):
+        raise PatchError("ProdML distance coordinates must be finite.")
+    diffs = np.diff(values)
+    step = coord.step
+    if step is None:
+        step = diffs[0] if len(diffs) else np.nan
+    step = float(step)
```
d-chambers: Just use the coord machinery here; this is probably redundant as coords already carry a step. You can also use require_evenly_sampled in get_coord to make this entire thing redundant.

dascore/io/prodml/utils.py
```diff
@@ -216,12 +234,254 @@ def _get_data_unit_and_type(node):
...
+        msg = "ProdML distance coordinates must be increasing and evenly sampled."
+        raise PatchError(msg)
+    units = coord.unit_str
+    if not units:
+        raise PatchError("ProdML distance coordinates require length units.")
+    try:
+        convert_units(1.0, to_units="m", from_units=units)
+    except UnitError as exc:
+        msg = "ProdML distance coordinate units must represent length."
+        raise PatchError(msg) from exc
```
d-chambers: Just assume in m if not provided. Again, use coordinate methods for converting units rather than redoing it here.

dascore/io/prodml/utils.py
```diff
@@ -216,12 +234,254 @@ def _get_data_unit_and_type(node):
...
+    ratio = float(values[0] / step)
+    start_locus = int(np.round(ratio))
+    integer_tolerance = max(1e-9, abs(float(np.spacing(ratio))) * 4)
+    if not np.isclose(ratio, start_locus, rtol=0, atol=integer_tolerance):
+        msg = "ProdML requires distance_start / distance_step to be an integer."
```
d-chambers: Is this right? Cant it be non-integer m? Please double check and simplify this.

dascore/io/prodml/utils.py
```diff
@@ -216,12 +234,254 @@ def _get_data_unit_and_type(node):
...
+        raise PatchError(msg)
+    return step, units, start_locus
+
+
+def _get_measure(attrs, name, units_name, target_units):
```
d-chambers, line 320: This smells of a general utility. Consider adding it to utils somewhere or double check that there isnt something like this already.

dascore/io/prodml/utils.py
```diff
@@ -216,12 +234,254 @@ def _get_data_unit_and_type(node):
...
+        value = float(value)
+        if not units or not np.isfinite(value) or value <= 0:
+            return None
+        convert_units(value, to_units=target_units, from_units=units)
+    except (TypeError, ValueError, UnitError):
+        return None
+    return value, units
+
+
+def _iso_time(microseconds):
```
d-chambers: This isnt needed. Cant we just use str(to_datetime64) then truncate the str?

dascore/io/prodml/utils.py
```diff
@@ -216,12 +234,254 @@ def _get_data_unit_and_type(node):
...
+    value = np.datetime64(int(microseconds), "us")
+    return _fixed_string(f"{np.datetime_as_string(value, unit='us')}+00:00")
+
+
+def _prepare_prodml_write(spool):
+    """Validate input and prepare all values needed to write ProdML."""
+    patch = _get_single_patch(spool)
+    if patch.data.ndim != 2 or set(patch.dims) != {"time", "distance"}:
+        msg = "ProdML writing requires a 2D Patch with time and distance dimensions."
+        raise PatchError(msg)
```
d-chambers: wrap these checks all up in _get_single_patch so it does all patch validation then either raises or returns the valid patch.

dascore/io/prodml/utils.py
```diff
@@ -216,12 +234,254 @@ def _get_data_unit_and_type(node):
...
+    acquisition_id = str(attrs.acquisition_id or uuid4())
+
+    facility = attrs.get("facility_id") or attrs.station or "UNKNOWN"
+    service_company = attrs.get("service_company_name") or "UNKNOWN"
+    data_units = get_quantity_str(attrs.data_units) or "UNKNOWN"
+    string_values = {
+        "FacilityId": facility,
+        "ServiceCompanyName": service_company,
+        "RawDataUnit": data_units,
+    }
```
d-chambers: This is ugly, maybe we can wrap it up into a pydantic model? EG subclass the current prodml patch attrs but add the extras/validation logic there.

dascore/io/prodml/utils.py
```diff
@@ -216,12 +234,254 @@ def _get_data_unit_and_type(node):
...
+    return value, units
+
+
+def _iso_time(microseconds):
+    """Return an ISO-8601 UTC timestamp with microsecond precision."""
+    value = np.datetime64(int(microseconds), "us")
+    return _fixed_string(f"{np.datetime_as_string(value, unit='us')}+00:00")
+
+
+def _prepare_prodml_write(spool):
```
d-chambers: This function is way too long. Please break into several sub functions (can be closures).

d-chambers: You can also make a small class for this if it is better.

tests/test_io/test_prodml/test_prodml_write.py
```diff
@@ -0,0 +1,461 @@
...
+        facility_id="facility-a",
+        service_company_name="service-a",
+        data_type="strain_rate",
+        data_units="m/s",
+    )
+
+
+def _decode(value):
+    """Decode one HDF5 string attribute."""
+    return str(unbyte(value))
```
d-chambers: just inline this

tests/test_io/test_prodml/test_prodml_write.py
```diff
@@ -0,0 +1,461 @@
...
+            assert set(file) == {"Acquisition"}
+            assert "sentinel" not in file.attrs
+            assert set(file["Acquisition"]) == {"Raw[0]"}
+
+
+class TestProdMLWriteRoundTrip:
+    """Representable Patch content should survive a write/read cycle."""
+
+    @pytest.mark.parametrize("dims", (("distance", "time"), ("time", "distance")))
+    def test_round_trip(self, dims, prodml_patch, tmp_path):
```
d-chambers, line 212: Just compare the patches directly here; the Patch.__eq__ will handle the drudgery. (you can pop out some attrs if you need to before compare)

d-chambers (commented):
Much of this code is overly verbose and duplicates existing functionality already part of dascore's objects or utils. Please address specific findings listed here and look for was to reduce line count. Ensure each helper function is more than a single line and actually needed.

dascore/io/prodml/utils.py
d-chambers: Addressed in a2ab81eb. The PRODML-local scalar encoder was removed and replaced by `encode_h5_strings` in `dascore.utils.hdf5`, so the HDF5 representation logic now lives with the shared backend utilities.

dascore/io/prodml/utils.py
d-chambers: Addressed in a2ab81eb. `encode_h5_strings` accepts either a string or a sequence and uses `iterate`; it returns an HDF5 scalar for a scalar input and a fixed-width ndarray for sequences, preserving the existing `FacilityId` and `Dimensions` shapes.

dascore/io/prodml/utils.py
d-chambers: Addressed in a2ab81eb. `_get_single_patch` now obtains both coordinates with `require_sorted=True` and `require_evenly_sampled=True`, eliminating the manual distance-difference/tolerance machinery. An explicit ascending check is retained because DASCore intentionally treats reverse-sorted coordinates as sorted; the regression suite now covers and rejects reversed distance.

dascore/io/prodml/utils.py
d-chambers: Addressed in a2ab81eb. Distance normalization now delegates to `coord.convert_units("m")`: it assigns meters when units are absent, converts compatible length units, and raises `UnitError` for incompatible units. The writer emits meters consistently, with a unitless-distance regression test.

dascore/io/prodml/utils.py
d-chambers: Addressed in a2ab81eb after checking the schema. `StartLocusIndex` is an integer locus index, not an integer-meter restriction. The simplified check only requires the physical start to align with the sampling grid; fractional coordinates such as start 0.5 m / step 0.5 m now have explicit successful round-trip coverage, while an unrepresentable start 0.5 m / step 2 m remains rejected.

dascore/io/prodml/utils.py
d-chambers: Checked in a2ab81eb. There is no existing DASCore utility combining a value/unit attribute pair with PRODML's format-specific 'omit when invalid' policy. `_get_measure` is substantive and reused for pulse rate, pulse width, and gauge length, so keeping it local avoids adding a nominally general API with only one consumer.

dascore/io/prodml/utils.py
d-chambers: Addressed in a2ab81eb. `_iso_time` was removed and the two timestamps are formatted inline from `datetime64[us]`. I did not route through `dc.to_datetime64`, because that normalizes through nanoseconds and would reintroduce overflow for the existing year-1600/year-2500 cases; `str(datetime64[us])` already provides exact microsecond precision.

dascore/io/prodml/utils.py
d-chambers: Addressed in a2ab81eb. `_get_single_patch` now owns cardinality, dimensions, nonempty supported dtype, even-sampling, and ascending-coordinate validation before returning the valid Patch. Representation-specific microsecond rounding and locus-grid conversion remain in their transform helpers so values are computed only once.

dascore/io/prodml/utils.py
d-chambers: Addressed in a2ab81eb with the smaller alternative. A Pydantic model would require aliases, byte-length validators, quantity handling, and exception translation for representation-only fields. `_get_write_metadata` now encapsulates the defaults and String64 validation directly, while the attribute builders consume its normalized values.

dascore/io/prodml/utils.py
d-chambers: Addressed in a2ab81eb. The former ~100-line preparation function is now a short coordinator, with substantive helpers for metadata normalization, Acquisition attributes, and raw/data/time array attributes. Together with coordinate and HDF5 utility reuse, the production implementation is five lines smaller while covering more edge cases.

tests/test_io/test_prodml/test_prodml_write.py
d-chambers: Addressed in a2ab81eb. The one-line `_decode` test helper was removed and its uses now call the existing `unbyte` utility directly.

tests/test_io/test_prodml/test_prodml_write.py
d-chambers: Addressed in a2ab81eb. The round-trip test now compares the patches directly after normalizing the two intentionally unstored core attrs (`tag` and `station`). It retains focused assertions for data dtype, coordinate units, and facility/service metadata because those are not covered by `Patch.__eq__`.

PR #765, opened 2026-07-23
Add no-op fast paths for transpose/squeeze and idempotent coordinate snapping

d-chambers:
## What

`Patch` and `CoordManager` are treated as immutable, but several metadata
operations rebuilt a new manager/patch even when the operation was a no-op. This
adds identity ("return `self`") fast paths so no-op operations skip
reconstruction while preserving **bit-for-bit** output, coordinate arrays, dim
maps, attrs, history, and invalid-value masks.

## Why

Profiling a large archive-rebuild workload showed meaningful time spent
rebuilding `Patch`/`CoordManager` metadata for operations that change nothing —
notably defensive canonical transposes, `squeeze()` on patches with no length-1
dims, and coordinate snapping where only one coordinate actually needs to change.
Because these objects are immutable, returning the original object is both
cheaper and clearer.

## Changes

- `CoordManager.transpose` returns `self` when the resolved dims equal the current dims.
- `CoordManager.squeeze` returns `self` when there are no length-1 dims to drop.
- `CoordManager.sort` returns `(self, array)` when no coordinate needs sorting.
- `CoordManager.snap` replaces only coordinates whose `.snap()` returns a
  different object, rebuilds the manager at most once, and returns the same
  manager unchanged when nothing needs snapping.
- Patch-level `transpose`, `squeeze`, `snap_coords`, and `sort_coords` return the
  original patch when the coordinate manager (and data) are unchanged.

Why this is safe: the `patch_function` decorator only records history when
`out is not patch`, so a no-op that returns `self` correctly records no history,
while real transforms are unaffected. `BaseCoord.snap`/`CoordRange` already
return `self` when even, which is the identity signal `CoordManager.snap` keys
off of.

The rolling engine dispatch (`_get_engine`) still uses `patch.squeeze()`; with
the squeeze fast path it now returns `self` for ordinary 2-D patches (no
allocation), so no dispatch change is needed. Added engine-selection tests lock
that behavior in.

## Tests / benchmarks

- Identity + regression tests for transpose, squeeze, sort, and snap
  (already-even returns self; only the uneven coordinate is replaced; reverse
  snapping; data reordering only when required).
- Engine-selection tests confirming `(1, N)`/1-D → pandas and `(M, N)` → numpy,
  plus explicit `engine=` precedence.
- No-op CodSpeed benchmarks: `test_transpose_noop`, `test_squeeze_noop`,
  `test_snap_coords_already_even` (existing real-work benchmarks retained).

## Verification

- `pytest tests` → 6316 passed, 84 skipped, 4 xfailed.
- Doctests for the three touched modules pass.
- `pre-commit` (ruff, ruff-format, etc.) clean.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved coordinate-processing efficiency by avoiding unnecessary work when sorting, snapping, transposing, or squeezing produces no effective change.
  * Preserved existing patch and coordinate objects when operations are no-ops.
  * Improved rolling-operation engine selection across multidimensional and one-dimensional data.

* **Bug Fixes**
  * Ensured coordinate data remains consistent when snapping and reordering uneven dimensions.

### DIFF
--- a/benchmarks/test_patch_benchmarks.py
+++ b/benchmarks/test_patch_benchmarks.py
@@ -114,6 +114,18 @@ class TestProcessingBenchmarks:
         patch.transpose(*dims)
 
     @pytest.mark.benchmark
+    def test_transpose_noop(self, example_patch):
+        """Time a no-op transpose (same dimension order)."""
+        patch = example_patch
+        patch.transpose(*patch.dims)
+
+    @pytest.mark.benchmark
+    def test_squeeze_noop(self, example_patch):
+        """Time a no-op squeeze (no length-1 dimensions)."""
+        patch = example_patch
+        patch.squeeze()
+
+    @pytest.mark.benchmark
     def test_roll(self, example_patch):
         """Time roll/shift operations."""
         patch = example_patch
@@ -124,6 +136,12 @@ class TestProcessingBenchmarks:
         """Time coordinate snapping."""
         patch = patch_uneven_time
         patch.snap_coords("time")
+
+    @pytest.mark.benchmark
+    def test_snap_coords_already_even(self, example_patch):
+        """Time snapping an already even/sorted patch (no-op fast path)."""
+        patch = example_patch
+        patch.snap_coords("time", "distance")
 
     @pytest.mark.benchmark
     def test_hampel_filter_non_approximate(self, example_patch):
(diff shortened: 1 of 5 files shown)


d-chambers:
@codex review

d-chambers:
@codex review

PR #767, opened 2026-07-23
Restore 100% coverage and enforce it in codecov

d-chambers:
## Why

`dev` coverage slipped below 100% and CI never flagged it. Tracing Codecov history, coverage was a clean **100.0%** through #758 and first read **99.96%** at **#759** (*"Prune dead code and collapse single-implementation abstractions"*), whose patch coverage was 89.29% — it added 6 uncovered defensive lines in `dascore/io/index/indexer.py`.

It slipped through because **`codecov.yml` never set an explicit coverage target**. The 100% "gate" was only Codecov's default `auto` status, which compares against the *base commit* rather than an absolute floor. Since the network-test split (#752) + `after_n_builds: 7` (#751) made per-commit uploads land as partial (unittests-only, ~48%), #759's head (99.96%) was compared against a broken ~48% base and read as an **increase** — so both project and patch checks passed.

## What

1. **Make the gate absolute and deterministic.** Add explicit `project` + `patch` `target: 100%` (threshold 0%), scoped to the **`unittests` flag** — the non-network + doctest suite, which now covers every line. The gate no longer depends on the carried-forward `network` upload, so a skipped or stale network report cannot mask a drop.
2. **Restore coverage to 100%.** Cover every previously-uncovered line with **non-network** tests:
   - `indexer.py` (the #759 regression): atomic-swap temp-file cleanup, and the `_walk` guard that skips a file deleted mid-scan.
   - `_coord_record_from_row`: empty-string units normalize to `None`.
   - `suppress_warnings`: the message-filter branch.
   - `_warn_remote_cache_download`: metadata-scope guidance text.
   - ProdML: relative (non-datetime) time rejected on write; non-positive optional measures dropped; `_get_prodml_version_str` returns `""` for a non-ProdML file.
3. **Remove one unreachable branch from the count.** The `_round_times_to_microseconds` NaT guard is marked `# pragma: no cover`: a NaT breaks even sampling, so `_get_single_patch`'s `require_evenly_sampled` check rejects such coords first (raising `CoordError`, a `PatchError` subclass), making the guard unreachable via `dc.write`.

## Verification

- Non-network suite: **8028 passed**; combined with the doctest suite (both upload under the `unittests` flag), **coverage is 100.0% — 0 misses**.
- `codecov.yml` validated against Codecov's validator (`Valid!`).
- Lint (ruff, ruff-format, pre-commit hooks): pass.

Network tests aren't run locally (no credentials); the `network` flag remains informational and no longer participates in the gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filesystem indexing resilience when files are removed during scanning or updates fail, preventing leftover temporary files and partial results.
  * Improved handling of coordinate records with empty unit values.
  * Clarified validation for PRODML time coordinates and optional measurement values.

* **Tests**
  * Expanded coverage for filesystem indexing, PRODML detection and writing, remote metadata warnings, warning suppression, and coordinate handling.
  * Enforced 100% unit-test coverage for project and patch changes.

## Changelog

none

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,21 +7,24 @@ comment:
 
 # Require full coverage. `target: 100%` makes the gate absolute rather than
 # relative to the base commit, so a regression cannot slip through when a base
-# upload is partial (see the coverage drop introduced by #759). The gate is on
-# combined coverage (all flags); it cannot be scoped to `unittests` alone
-# because some lines are only exercised by the network suite. Note the network
-# flag carries forward (below): if a network upload is skipped, its prior report
-# is reused, so a regression in network-only code could still read as 100%.
+# upload is partial (see the coverage drop introduced by #759). The gate is
+# scoped to the `unittests` flag (the deterministic non-network + doctest
+# suite, which covers every line): it never depends on the carried-forward
+# `network` upload, so a skipped or stale network report cannot mask a drop.
 coverage:
   status:
     project:
       default:
         target: 100%
         threshold: 0%
+        flags:
+          - unittests
     patch:
       default:
         target: 100%
         threshold: 0%
+        flags:
+          - unittests
 
 flags:
   unittests:
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,6 +5,24 @@ codecov:
 comment:
   after_n_builds: 7
 
+# Require full coverage. `target: 100%` makes the gate absolute rather than
+# relative to the base commit, so a regression cannot slip through when a base
+# upload is partial (see the coverage drop introduced by #759). The gate is on
+# combined coverage (all flags); it cannot be scoped to `unittests` alone
+# because some lines are only exercised by the network suite. Note the network
+# flag carries forward (below): if a network upload is skipped, its prior report
+# is reused, so a regression in network-only code could still read as 100%.
+coverage:
+  status:
+    project:
+      default:
+        target: 100%
+        threshold: 0%
+    patch:
+      default:
+        target: 100%
+        threshold: 0%
+
 flags:
   unittests:
     carryforward: false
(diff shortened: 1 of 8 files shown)


tests/test_io/test_indexer.py
```diff
@@ -172,6 +172,54 @@ def test_get_reads_fresh_each_call(self, tmp_path):
...
+        def is_dir_then_vanish(self, *args, **kwargs):
+            # Delete the file immediately after the is_dir() probe so the walk's
+            # subsequent real stat() hits the concurrent-deletion guard. This
+            # reproduces the race without assuming how is_dir() is implemented.
+            result = real_is_dir(self, *args, **kwargs)
+            if self.name == "vanisher.h5":
+                self.unlink()
+            return result
+
+        monkeypatch.setattr(Path, "is_dir", is_dir_then_vanish)
```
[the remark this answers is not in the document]
d-chambers: Addressed in 3823cbd: the test no longer patches `Path.is_dir` globally or deletes a file mid-walk. It now mocks `_iter_filesystem` to deterministically yield a real file plus a never-created "vanished" path, whose real `stat()` raises `FileNotFoundError` and exercises the guard — independent of is_dir/stat call ordering. Verified it still covers indexer.py lines 273-277.

codecov.yml
```diff
@@ -5,6 +5,27 @@ codecov:
 comment:
   after_n_builds: 7
 
+# Require full coverage. `target: 100%` makes the gate absolute rather than
+# relative to the base commit, so a regression cannot slip through when a base
+# upload is partial (see the coverage drop introduced by #759). The gate is
+# scoped to the `unittests` flag (the deterministic non-network + doctest
+# suite, which covers every line): it never depends on the carried-forward
+# `network` upload, so a skipped or stale network report cannot mask a drop.
```
d-chambers: remove these lines; too verbose.

d-chambers: Done in 40247596 — trimmed to a single line.

PR #768, opened 2026-07-23
Skip reparsing canonical CoordRange coords in CoordManager update/select

d-chambers:
## What

`CoordManager.update` / `update_coords` / `select` funnel every coordinate
through `_get_coord` (nested in `_get_coord_dim_map`), which dumps each coord via
`model_dump()` and rebuilds it with `get_coord(**dict)` — re-running full
validation and, for array coords, the monotonic / even-sampling scan — **even
when the coordinate was already a valid, canonical object**. Because `BaseCoord`
is a pydantic model, this bypassed `get_coord`'s own
`if isinstance(data, BaseCoord): return data` fast path.

This adds a single short-circuit for the one type where returning the coord
unchanged is provably a no-op:

```python
if isinstance(coord, CoordRange):
    return coord
```

A `CoordRange` is the canonical evenly-sampled representation and round-trips
identically through `model_dump()` → `get_coord()`.

## Why only CoordRange (not all BaseCoords)

The reparse also *canonicalizes* non-canonical coordinates, so it cannot be
blanket-skipped:

- **Array coords** (`CoordArray` / `CoordMonotonicArray`) can be left
  non-canonical by slicing — e.g. an evenly spaced subset produced by
  `__getitem__` during sample selection should collapse to a `CoordRange`.
- A **fully-specified `CoordPartial`** should canonicalize to a `CoordRange`
  (otherwise value-based selection on it wrongly raises).

Both cases are covered by regression tests. `CoordRange` was verified (and
independently confirmed in review) to round-trip identically for forward,
reversed, singleton, datetime, unit-bearing, and dtype-bearing variants.

## Performance (micro-benchmark on master-equivalent code)

| Operation | Before | After |
|---|---:|---:|
| `update_coords`, CoordRange dims | ~393 µs | ~22 µs (**~17×**) |
| value-based `select`, CoordRange | ~596 µs | ~180 µs (**~3.3×**) |

Array-coord paths are intentionally unchanged (they still reparse for
canonicalization). Output is bit-for-bit identical: coordinate arrays, dim maps,
dtypes, data, attrs, and history are unchanged.

## Tests / benchmarks

- Identity preservation for `CoordRange` update/select; value-equivalence for
  array coords; the select result is unchanged.
- Regression guards: even sample-select subset of an irregular coord →
  `CoordRange`; fully-specified `CoordPartial` → `CoordRange` with working
  value-select.
- New `test_update_existing_coords` CodSpeed benchmark.

## Verification

- `pytest tests` → 6318 passed, 84 skipped, 4 xfailed.
- Doctests for `coordmanager.py` pass; `pre-commit` clean.
- Reviewed via Codex CLI across three iterations; two real edge cases it found
  (array-slice and fully-specified-CoordPartial canonicalization) were fixed and
  now have regression tests.

Follow-up to #765 / #757; see `.scratch/optimization_ideas.md` (PR 3).

### DIFF
--- a/benchmarks/test_patch_benchmarks.py
+++ b/benchmarks/test_patch_benchmarks.py
@@ -89,6 +89,16 @@ class TestProcessingBenchmarks:
         patch.select(time=(t1, t2))
 
     @pytest.mark.benchmark
+    def test_update_existing_coords(self, example_patch):
+        """Time update_coords re-passing already-validated BaseCoords."""
+        patch = example_patch
+        coords = patch.coords
+        patch.update_coords(
+            time=coords.coord_map["time"],
+            distance=coords.coord_map["distance"],
+        )
+
+    @pytest.mark.benchmark
     def test_sobel_filter(self, example_patch):
         """Time the Sobel filter."""
         patch = example_patch
--- a/dascore/core/coordmanager.py
+++ b/dascore/core/coordmanager.py
@@ -54,7 +54,12 @@ from typing_extensions import Self
 
 import dascore as dc
 from dascore.constants import dascore_styles, select_values_description
-from dascore.core.coords import BaseCoord, CoordSummary, get_coord
+from dascore.core.coords import (
+    BaseCoord,
+    CoordRange,
+    CoordSummary,
+    get_coord,
+)
 from dascore.exceptions import (
     CoordDataError,
     CoordError,
@@ -1272,6 +1277,15 @@ def _get_coord_dim_map(coords, dims):
 
     def _get_coord(coord):
         """Get a coordinate from various inputs."""
+        # A CoordRange is already canonical (it is the evenly-sampled
+        # representation), so re-parsing it via model_dump -> get_coord is pure
+        # overhead; return it directly. Other coord types are NOT short-circuited:
+        # array coords (CoordArray/CoordMonotonicArray) can be left non-canonical
+        # by slicing (e.g. an evenly spaced subset that should collapse to a
+        # CoordRange), and a fully-specified CoordPartial should canonicalize to a
+        # CoordRange -- get_coord performs that inference.
+        if isinstance(coord, CoordRange):
+            return coord
         if hasattr(coord, "model_dump"):
             coord = coord.model_dump(exclude_defaults=True)
         if isinstance(coord, Mapping):  # input is a dict
--- a/tests/test_core/test_coordmanager.py
+++ b/tests/test_core/test_coordmanager.py
@@ -17,6 +17,8 @@ from dascore.core.coordmanager import (
 )
 from dascore.core.coords import (
     BaseCoord,
+    CoordPartial,
+    CoordRange,
     get_coord,
 )
 from dascore.exceptions import (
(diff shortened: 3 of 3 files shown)


d-chambers:
@codex review

d-chambers:
### Why the guard is `CoordRange` only (the "canonical coords" reasoning)

The core fact behind this change: **the same coordinate values can be stored in more than one representation, and DASCore has a preferred ("canonical") one.**

A coordinate is a `BaseCoord`, which can be any of:

| Type | Stores | Used when |
|---|---|---|
| `CoordRange` | `start, stop, step` | values are **evenly spaced** |
| `CoordMonotonicArray` | the full array | increasing but **uneven** (e.g. `[0, 1, 3, 9]`) |
| `CoordArray` | the full array | not even sorted |
| `CoordPartial` | just a length | a dim with a size but **no values** |

`[0, 10, 20, 30]` is evenly spaced, so its canonical form is `CoordRange(start=0, stop=40, step=10)` — compact, and it reports `evenly_sampled == True`. The same values as a `CoordMonotonicArray` would be correct-but-non-canonical (bulkier, and reports `evenly_sampled == False`).

`get_coord(data=...)` is what performs this inference ("evenly spaced → `CoordRange`, monotonic-uneven → `CoordMonotonicArray`, empty → `CoordPartial`"), so anything built through it comes out canonical.

**The redundancy this PR removes.** Every `update`/`select` funneled each coord through `_get_coord`, which did:

```python
coord = coord.model_dump()   # dump a valid coord object to a plain dict
coord = get_coord(**dict)    # ...and rebuild it from scratch
```

even when `coord` was already a valid, canonical `CoordRange`. That round-trip re-runs the full even-spacing scan over the array for nothing — the ~393 µs → ~22 µs (`update_coords`) and ~596 µs → ~180 µs (value `select`) cost.

**Why we can't skip the rebuild for every coord.** That rebuild wasn't *only* wasted work — for some coords it was silently **canonicalizing** a coord that had gone stale. Coords go non-canonical via **slicing**: sample-selection does array indexing (`coord[indices]`) which keeps the same type and does **not** re-infer spacing. So:

- `[0, 1, 3]` → `CoordMonotonicArray` (uneven, correct)
- sample-select indices `0,1` → subset `[0, 1]`
- the slice stays a `CoordMonotonicArray([0, 1])` — but `[0, 1]` **is** now evenly spaced and *should* be a `CoordRange`.

The old rebuild fixed that (`get_coord([0,1])` → `CoordRange`). A blanket "skip all `BaseCoord`s" guard would drop the fix-up, leaving a `CoordMonotonicArray` that wrongly reports `evenly_sampled == False` — same values, wrong representation. (A second variant: a fully-specified `CoordPartial` should expand into a `CoordRange`, else value-based selection on it raises.)

**Why `CoordRange` alone is safe.** A `CoordRange` can never be non-canonical — it *is* the canonical even form, and slicing a `CoordRange` yields another `CoordRange`. So rebuilding one is *always* pure waste. Every other type might be sitting in a non-canonical state, so those keep going through the rebuild:

```python
if isinstance(coord, CoordRange):
    return coord      # already canonical — no rebuild needed
# all other types fall through to the rebuild, which canonicalizes them
```

The two regression tests pin exactly the traps that ruled out a broader guard: an even sample-select subset must become a `CoordRange`, and a fully-specified `CoordPartial` must become a `CoordRange`.

**TL;DR:** the old code rebuilt every coordinate partly to waste time and partly to *canonicalize* coords that slicing had left in the wrong representation. This skips the rebuild only for `CoordRange` — the one type guaranteed already-canonical — dropping the wasted work without dropping the fix-up.

Issue #508, opened 2025-07-11
JSONDecodeError: Extra data when trying to read a directory for SPOOL


d-chambers:
### Discussed in https://github.com/DASDAE/dascore/discussions/507

<div type='discussions-op-text'>

<sup>Originally posted by **pongchayut** July 11, 2025</sup>
Hey everyone,

I’m running into a weird issue with dascore. I tried this simple code:

```
import dascore as dc
spool = dc.get_example_spool("random_directory_das")
```
But it throws this error:
```
JSONDecodeError: Extra data: line 1 column 999 (char 998)
```
And this also happened with real data. It used to work just fine before, and I didn’t change anything in the code or the data. I'm working with .tdms files from a Silixa system. What’s strange is that if I pass a single .tdms file directly to dc.spool(), it works perfectly. The error only appears when I attempt to load the entire directory.

I'm not sure what went wrong or what caused this. Has anyone seen this before or know how to fix it?

Thanks in advance!

Python Version: 3.11.13
DASCore Version: 0.1.7

Full Traceback:
```
---------------------------------------------------------------------------
JSONDecodeError                           Traceback (most recent call last)
Cell In[17], line 1
----> 1 spool = dc.get_example_spool("random_directory_das")

File ~/anaconda3/envs/das311/lib/python3.11/site-packages/dascore/examples.py:732, in get_example_spool(example_name, **kwargs)
    727     msg = (
    728         f"No example spool registered with name {example_name} "
    729         f"Registered example spools are {list(EXAMPLE_SPOOLS)}"
    730     )
    731     raise UnknownExampleError(msg)
--> 732 return EXAMPLE_SPOOLS[example_name](**kwargs)

File ~/anaconda3/envs/das311/lib/python3.11/site-packages/dascore/examples.py:570, in random_directory_spool(path, **kwargs)
    568 spool = random_spool(**kwargs)
    569 path = spool_to_directory(spool, path)
--> 570 return dc.spool(path)

File ~/anaconda3/envs/das311/lib/python3.11/functools.py:909, in singledispatch.<locals>.wrapper(*args, **kw)
    905 if not args:
    906     raise TypeError(f'{funcname} requires at least '
    907                     '1 positional argument')
--> 909 return dispatch(args[0].__class__)(*args, **kw)

File ~/anaconda3/envs/das311/lib/python3.11/site-packages/dascore/core/spool.py:711, in _spool_from_str(path, **kwargs)
    708 if path.is_dir():
[...]

PR #772, opened 2026-07-24
ENH: per-directory files for the directory index-location map

d-chambers:
## Description

Simplified successor to #764: it replaces the shared JSON directory index-location map with **one small JSON file per data directory**, instead of the SQLite database (with corruption-recovery locks and `os.register_at_fork` hooks) that #764 proposed.

The index-location map records where a directory's index lives when the data directory itself is read-only, or when a user supplies a custom index path. Its goals are (a) concurrent writers to distinct data directories must not lose each other's entries, and (b) a corrupt entry (see #508) must not break indexing.

### What changed

- **One entry file per data directory** (`<sha256(dir)>.json`) under a cache directory, replacing the single shared `cache_paths.json`. Different directories use different files, so concurrent writers to distinct directories cannot clobber one another.
- **Corruption is per-entry and self-healing.** A corrupt/unreadable/malformed entry simply reads as a miss; the next atomic write for that directory self-heals it. No SQLite, no advisory locks, no recovery protocol, no at-fork hooks. Writes swap a sibling temp file into place, so readers never see a half-written entry.
- **The read-only fallback records nothing.** It derives a deterministic index filename from a full-sha256 digest of the directory path, so it is rediscoverable without a map entry. Only a user-specified custom index path is recorded.
- Config option `directory_index_map_path` becomes `directory_index_map_dir` (now a directory). The map is a disposable cache; any old shared `cache_paths.json` is ignored and left in place.

### Why not the SQLite approach from #764

For a disposable convenience cache whose only production reads/writes are single-key, per-entry files give the same lost-update and corruption guarantees by construction, with ~10% of the code and none of the process-wide side effects (the `os.register_at_fork` hooks in #764 serialize every fork in the host process through a library lock). It is also friendlier to NFS home directories, where SQLite locking is unreliable.

Supersedes #764. Part of the free-threading work extracted from #763.

### Validation

- `tests/test_io/test_indexer.py`, `tests/test_io/test_index/`, `tests/test_utils/test_config.py`: pass.
- Full `tests/test_io/`: 2053 passed, 63 skipped, 1 xfailed.
- `prek`/pre-commit hooks pass on changed files.
- Independent Codex review: its four findings (read-side deletion racing a concurrent repair; non-UTF-8/malformed payload handling; digest truncation risking a read-only index-file collision; non-UTF-8 path bytes) were all addressed with regressions.

## Changelog

- changed: the external index-location map is stored as one small file per data directory under `directory_index_map_dir` instead of a shared `cache_paths.json`, so a corrupt entry only affects its own directory. Any old `cache_paths.json` is ignored and left in place.

## Checklist

I have (if applicable):

## Summary by CodeRabbit

- **New Features**
  - Index-location mappings are now stored as separate files in a cache directory, improving isolation between data directories.
  - Unwritable directories receive stable, deterministic index locations.
  - Corrupt mapping entries are safely ignored and can self-heal without affecting other directories.
  - Concurrent updates for different directories no longer overwrite one another.

- **API Changes**
  - Replaced the `directory_index_map_path` setting with `directory_index_map_dir`.

- **Bug Fixes**
  - Corrupt or mismatched mapping data now impacts only the affected directory rather than broader index behavior.


dascore/io/index/indexer.py
```diff
@@ -32,52 +32,68 @@
...
-def _get_index_map(cache_path) -> dict:
+def _path_digest(path) -> str:
+    """Stable per-path digest (hash() of a str/Path is per-process random).
+
+    Uses the full sha256 and os.fsencode so distinct paths never collide
+    (a collision would make two read-only directories share one index
+    file) and non-UTF-8 filename bytes digest without error.
     """
-    Return a fresh dict of index locations read from disk.
+    return hashlib.sha256(os.fsencode(path)).hexdigest()
```
d-chambers: We may also need to support URLs in the future. Please just use fsencode if possible, fallback to string only hash if not.

d-chambers: Done in 34c4bb18: `_path_digest` now tries `os.fsencode` first (exact local-path bytes) and falls back to a plain string encoding for inputs it rejects — e.g. a remote URL/UPath directory whose `__fspath__` isn't available — instead of raising. Added a regression covering the fallback branch.

PR #784, opened 2026-07-25
Fix remote HTTP test stalls and tune HTTP HDF5 caching

d-chambers:
## Description

Tracks down the intermittent remote-HTTP deadlock/stall noted in the `test_remote_http.py` module docstring (the Windows skip rationale). Pinning the suite to 2 CPUs made it reproducible, which separated it into three independent causes.

### 1. Single-threaded fixture servers (test-only)

Two localhost fixture servers (`http_regression_das_path`, `http_range_das_path`) were single-threaded `HTTPServer`s. One slow or abandoned connection parks the accept loop, and every new connection then sits in TCP SYN retry backoff — observed as reproducible 15 s stalls (the SYN retry schedule), 30 s pytest-timeout failures, or an indefinite hang when pytest-timeout isn't installed. All three fixtures now share one threaded-server context manager. Before: 2/3 pinned module runs hit the 30 s timeout; after: 5/5 stable.

### 2. A real deadlock between h5py's lock and garbage collection (product bug)

**This is a known failure mode which h5py documents**, under [Python file-like objects](https://docs.h5py.org/en/stable/high/file.html#python-file-like-objects):

> When using a Python file-like object, using service threads to implement the file-like API can lead to process deadlocks. `h5py` serializes access to low-level hdf5 functions via a global lock. This lock is held when the file-like methods are called and is required to delete/deallocate `h5py` objects. Thus, if cyclic garbage collection is triggered on a service thread the program will deadlock.

fsspec's event loop is exactly that service thread. DASCore reads remote HDF5 through h5py's `fileobj` driver, so each fetch is serviced on the loop thread while the reader holds the global lock; a cyclic collection landing there needs the same lock, and neither side moves.

h5py names two mitigations: avoid reference cycles which keep h5py objects alive, or temporarily disable garbage collection. **We take the second.** The first is not available to a library — the cycle can be anywhere in the process. (h5py's third suggestion, avoiding file-like objects altogether, is what DASCore's existing `ensure_local_file` fallback does when streaming is not viable; it costs a full download, which is what the streaming path exists to avoid.)

This is platform-agnostic. Windows simply lost the timing coin flip more often, which is why it was previously filed as Windows flakiness. HDF5 handles backed by an **async fsspec filesystem** (HTTP, S3, ...) now pause automatic collection for the handle's lifetime (`dascore.utils.remote_io.pause_gc`) and resume it on close; reference counting still frees non-cyclic garbage. Local paths and synchronous backends such as `memory://` are untouched.

The pause is balanced across the cases that can strand it: an interrupt anywhere inside the paused region, a `BaseException` raised during teardown, a handle inherited through a fork (handles record the pid that paused for them, so a child cannot resume a pause it never took), an interrupt inside a `FiberIO` read or write, and buffered wrappers that hide the fsspec filesystem from the loop-backed check. A rate-limited `gc.collect()` on the next remote open recovers a handle leaked inside a reference cycle, which `__del__` cannot reach while collection is off.

**Because this is fixed, the Windows skip is removed and the `network_tests` matrix reuses the shared OS list** — Linux, macOS, and Windows now all exercise the localhost-HTTP path. That job is `continue-on-error`, so it reports without gating unrelated changes.

Note that the pause is process-global for the lifetime of an open remote handle: cyclic garbage from every thread accumulates until the last one closes, and a handle leaked inside a cycle keeps it paused until the next remote open, if the program makes one. That is a deliberate trade against a hard deadlock. It is documented for users in the [remote patches tutorial](https://dascore.org/tutorial/remote_patches.html) and for developers on `pause_gc`, both of which cite the h5py page above.

### 3. HTTP HDF5 reads overfetched (performance)

Remote HDF5 opens over HTTP amplify transfers: h5py's metadata probe alternates between the file header and footer, and fsspec's default single-window `BytesCache` refetches a multi-MB block on every jump — debug logs showed 18+ MB fetched to probe a 12.7 MB file, and range-less servers re-stream the entire file per miss. `H5Reader._get_open_kwargs` already tuned S3 for exactly this reason; HTTP now gets the tuned block size too, with `cache_type="blockcache"` (a block LRU that keeps both file ends resident, capped at 8 blocks).

Trade-off worth knowing: `BlockCache` issues one range request per block, where the default cache coalesces a large contiguous read into one request. That favours the metadata probe this PR is about and costs requests on bulk data reads over HTTP, which is the path DASCore already steers to the local cache.

Separately, servers that report no size stream rather than range, so `is_no_range_http_error` now also routes "cannot seek streaming HTTP file" to the same local-file fallback. Such a server therefore materializes a file locally to probe it, which is the behaviour range-less servers already had.

## Also here

`dc.write` of HDF5 to a remote path no longer commits a partial upload when the write fails. A remote HDF5 write buffers into a temp file and uploads it on close, so a mid-write error used to upload whatever had been written; both `IOResourceManager.__exit__` and the `FiberIO` type caster now abort instead, and share one `release_handle` helper. `close_all` isolates per-handle failures so one bad close cannot skip the rest, since remote handles resume the GC pause in `close`, and a cleanup failure no longer replaces the error that caused it.

## Review notes

Taken over from the original branch: rebased onto current `dev` and reviewed end to end, then put through three internal adversarial reviews plus a Codex review. Fixes those surfaced, beyond the original scope:

- Local and `memory://` `UPath` HDF5 reads took the remote branch and disabled process-wide GC on every read; the pause is now gated on `fs.async_impl`.
- A failed HDF5 open closed a caller-supplied file object, which breaks `get_format` (it offers the same object to every FiberIO in turn).
- `__del__` closed handles DASCore does not own; it now only releases the pause.
- `_type_caster`'s failure path committed a partial remote HDF5 upload; it aborts now.
- `pause_gc` could leave its depth inconsistent when interrupted, which would have silently disabled the fix for the rest of the process.

Known and accepted: the GC pause is process-global (see above); `hasattr(handle, "abort")` matches only the remote HDF5 writer, so other remote writers still stream directly; `_type_caster` closes a caller-supplied stream when a read fails, matching what it already did on success.

## Changelog

- fixed: remote HDF5 reads over an async fsspec backend (HTTP, S3) could deadlock. Automatic garbage collection is now paused while such a handle is open — [h5py's recommended mitigation](https://docs.h5py.org/en/stable/high/file.html#python-file-like-objects). See [Working with Remote Patches](https://dascore.org/tutorial/remote_patches.html).
- fixed: a failed HDF5 write to a remote path no longer uploads a partial file.
- fixed: HTTP servers that report no file size no longer fail on seek; DASCore falls back to a locally cached copy.
- changed: remote HDF5 over HTTP uses a bounded block cache, cutting metadata overfetch but retaining ~40 MB per open handle (8 x `remote_hdf5_block_size`).

## Checklist

I have:

I have (if applicable):


d-chambers:
Note on scope: this fixes the reproducible stall class (accept-loop parking — before: 2/3 pinned runs of the module hit the 30 s pytest-timeout; after: 5/5 stable) and the transfer amplification, but one intermittent stall remains in `test_http_range_hdf5_read_succeeds`: under CPU contention the ranged `dc.read` sometimes blocks for ~15 s (suspiciously the aiohttp keep-alive quantum) before completing, and the first CI run here hit it on the non-gating macOS network job (30 s timeout; the test's internal `skip_on_timeout(15)` can only fire after the blocked call returns). It's heisenbug-grade — enabling debug logging or the thread-method timeout makes it vanish — so I've left it rather than guess. If it keeps surfacing, the next lead is the aiohttp connection-reuse path for sequential range requests against connections the HTTP/1.0 server has just closed.

d-chambers:
Update: the residual stall is root-caused and fixed in ba6ba5cd — it was never a network problem. It's a platform-agnostic ABBA deadlock: h5py holds its global `phil` lock while a remote fileobj read blocks on fsspec's event-loop thread, and an automatic GC cycle triggered **on that loop thread** needs `phil` to deallocate dead h5py objects (from the `get_format` call earlier in the same test). Main thread holds phil waiting on the loop; the loop's GC waits on phil.

Evidence: faulthandler burst dumps during a live stall show the loop thread frozen at an allocation point with `Garbage-collecting` set and ~0 CPU; `gc.disable()` eliminated the stall 6/6 (it reproduced by run 1–2 otherwise); `gc.callbacks` tracing shows collections running on the fsspecIO thread exactly during the 5 MB range fetches. This also explains the "Windows flakiness" the module docstring described, the heisenbug behavior (instrumentation shifts allocation counts), and why the in-test `skip_on_timeout` sometimes couldn't fire.

Fix: remote h5py handles pause automatic collection for their bounded lifetime (resumed on close) — the mitigation h5py's [file-object docs](https://docs.h5py.org/en/stable/high/file.html) recommend for exactly this pattern. Pre-pause `gc.collect()` bounds cycle buildup, the pause is BaseException-safe, close is race-safe, and a `__del__` backstop covers leaks. The `skip_on_timeout` band-aid and stale TODO are removed so CI exercises the fixed path. Validated: 8/8 pinned repro runs clean (formerly deterministic stall), full suite green.

d-chambers:
Ran three independent adversarial reviews on the GC-pause mechanism (lock-ordering, CPython GC semantics, handle lifecycle/platforms) — commit 6a0a8ef9 addresses everything they surfaced:

**Real gaps fixed:**
- `io.IOBase` branch bypassed the pause — a user passing an fsspec file object directly (`dc.read(fsspec_file)`) still hit the original deadlock. Loop-backed file objects now pause too.
- `_type_caster` leaked its opened handle when a FiberIO method raised → GC could stay paused indefinitely. Exception path now closes.
- `IOResourceManager.__exit__` committed unconditionally → a mid-write exception **uploaded a partial file** to remote targets (pre-existing, adjacent). Now aborts on error.
- Maintenance collect moved outside the pause lock (a finalized leaked handle would self-deadlock), made unconditional (leaked handles self-heal on later opens) and rate-limited (measured 7–160 ms per full collect — matters at spool scale).
- KeyboardInterrupt-safe ordering in pause/resume; fork handler resets pause state in children (matching the existing `_reinit_after_fork` pattern).

**Verified sound by the reviews** (now documented in code): refcount-driven h5py deallocs can never land on the loop thread because references flow strictly one way through the sync bridge; the implementation matches h5py's file-object docs' recommended mitigation; wasm and free-threaded CPython carry over.

**Windows:** the review confirmed the fixture stack is Windows-correct and the skip's stated cause is the now-fixed platform-agnostic deadlock, so the win32 skip is removed and `windows-latest` joins the non-gating `network_tests` matrix as a canary.

dascore/utils/hdf5.py
```diff
@@ -199,9 +199,14 @@ def _get_open_kwargs(resource: UPath) -> dict[str, object]:
             # h5py performs many small seeks while opening HDF5 metadata.
             # s3fs defaults to 50 MB readahead blocks, which can pull most of
             # a large remote file just to satisfy metadata probes.
+            # HTTP needs a block LRU cache instead: the metadata probe
+            # alternates between the file header and footer, and fsspec's
+            # default single-window cache refetches a full block (or the whole
+            # file on range-less servers) on every jump.
+            cache_type = "blockcache" if protocol in ("http", "https") else "readahead"
```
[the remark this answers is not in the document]
d-chambers: Addressed. \`H5Reader._get_open_kwargs\` now passes \`cache_options={"maxblocks": 8}\` alongside \`cache_type="blockcache"\`, so retained data is bounded at 8 blocks rather than fsspec's default 32.

With the default \`remote_hdf5_block_size\` of 5 MiB that is ~40 MiB per open HTTP handle (up from ~4 MiB under fsspec's previous single-window default), which is now called out in the changelog so it is not a silent regression. Covered by \`test_remote_h5_open_kwargs_are_tuned\`, parametrized over s3/http/https.

The hard-coded 8 is worth promoting to a config field next to \`remote_hdf5_block_size\`; tracked as a follow-up rather than done here.

d-chambers:
## Bot review disposition

All CodeRabbit and Codex findings on this PR are addressed; every review thread is now resolved. Summary for reviewers, since the branch was rebased and several threads are marked outdated:

| Finding | Source | Disposition |
|---|---|---|
| Cap the HTTP block cache | Codex | `cache_options={"maxblocks": 8}` bounds retained data at 8 blocks instead of fsspec's default 32. The resulting ~40 MiB per open HTTP handle is stated in the changelog. |
| `close_all` aborts the whole cleanup loop if one handle raises | CodeRabbit | Per-handle isolation; the first error is re-raised only after every handle was attempted. `test_close_all_survives_failing_handle`. |
| `resume_gc()` skipped if `fileobj.close()` raises a `BaseException` | CodeRabbit | Nested `try/finally` in `_open_h5_fileobj`, mirroring `_ManagedH5pyFile.close`. `test_teardown_error_still_resumes`. |
| Document that GC resumes after the *last* handle closes | CodeRabbit | Wording corrected. The in-repo changelog was retired by #864, so this now lives in the remote-patches tutorial and the PR changelog. |
| Committed Pint cache metadata (`.xdg-test/`) | CodeRabbit | Stray files removed; no longer in the diff. |
| Cover HTTPS as well as HTTP in the cache test | CodeRabbit (nitpick) | `test_remote_h5_open_kwargs_are_tuned` parametrized over s3 / http / https. |
| Restore `_gc_collect_after` via `monkeypatch` | CodeRabbit (nitpick) | Both tests now use `monkeypatch.setattr`. |
| Make the gc-guard failure message match the detected condition | CodeRabbit (nitpick) | The root conftest fixture reports pause depth, actual gc state, and expected state separately. |

Beyond the bots, the branch also went through three internal adversarial reviews (concurrency/lock-ordering, IO semantics/ownership, tests/CI) and a Codex CLI review. Those surfaced several issues the bots did not, the most significant being: `__del__` closed handles DASCore does not own; the type caster's failure path committed a partial remote HDF5 upload; an interrupt inside `pause_gc` could leave its depth inconsistent and silently disable the fix for the rest of the process; and local/`memory://` `UPath` reads were pausing process-wide GC unnecessarily. All are fixed and covered by tests.

.github/workflows/runtests.yml
```diff
@@ -154,7 +154,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        # The same OS list as test_code; the deadlock these tests cover was
+        # platform-agnostic, so every supported platform should exercise it.
```
d-chambers: Just remove this

dascore/io/core.py
```diff
@@ -860,22 +865,31 @@ def _wrapper(*args, _pre_cast=False, **kwargs):
...
+            # including on KeyboardInterrupt: leaking a remote handle leaves
+            # garbage collection paused for as long as the traceback is
+            # retained. Abort rather than close, so a failed remote write
+            # discards its temp file instead of uploading a partial one.
+            if new_resource is not None and new_resource is not resource:
+                with suppress(Exception):
+                    release_handle(new_resource, abort=True)
+            # get_format can't raise; it must return False instead.
+            if fun_name != "get_format" or not isinstance(e, Exception):
+                raise
```
d-chambers: If it cant raise, why do we raise right after this check? Let's make sure the comment and the check are correct.

.github/workflows/runtests.yml
d-chambers: Removed in 3ea0a242.

dascore/io/core.py
d-chambers: Good catch — the check is right, the comment was wrong. Fixed in 3ea0a242.

`get_format` reports "not my format" by returning `False`, so the format-detection loop can try the next FiberIO. That is true for an ordinary `Exception` only. The `raise` covers two cases: any other method (which propagates normally), and a `BaseException` that is not an `Exception` raised inside `get_format` — `KeyboardInterrupt`, `SystemExit`.

That second case is new to this PR and is why the explicit check exists. On `dev` the clause was `except Exception`, so a Ctrl-C escaped on its own. This PR widens it to `except BaseException` purely so the handle cleanup runs on an interrupt — a leaked remote handle would leave garbage collection paused for as long as the traceback is retained. Widening the catch means the interrupt no longer escapes by itself, so it has to be re-raised deliberately. Without that, Ctrl-C during format detection would be swallowed and turned into "not this format".

The comment now says that instead of the misleading "get_format can't raise".

PR #788, opened 2026-07-26
Stop persisting GITHUB_TOKEN in CI checkouts

d-chambers:
## Description

`actions/checkout` defaults to `persist-credentials: true`, which writes the job's `GITHUB_TOKEN` into `.git/config` as an `http.extraheader`. It then stays readable by every subsequent step in the job, including the third-party tooling our workflows run (mamba/conda, npm, quarto, pyodide-build, codspeed). `runtests.yml` already set `persist-credentials: false` on the network job; this applies the same setting to the remaining 11 checkout steps.

Nothing in CI needs the persisted credentials — no workflow pushes with git:

- docs deploy via `quarto publish` (Netlify token) and `actions/deploy-pages` (OIDC)
- release assets via `softprops/action-gh-release`
- PyPI via `pypa/gh-action-pypi-publish`
- `build_deploy_stable_docs.yaml` only runs local `git tag --list` / `git restore`

The one remote git call is `git fetch --tags --force` in `.github/actions/mamba-install-dascore`, which works anonymously against this public repo and already runs credential-free today in the network test job.

This clears all `artipacked` findings from zizmor (`uvx zizmor@1.26.1 .github/workflows/` goes from 12 to 0; the remaining `unpinned-uses` findings are pre-existing and out of scope).

Raised by CodeRabbit on #783, which flagged the new WASM job's checkout. Rather than fix that one step in isolation, all of them are handled here. #783 keeps only the Pyodide version alignment.

## Changelog

none

## Checklist

I have (if applicable):

- [ ] included tests. See [testing guidelines](https://dascore.org/contributing/testing.html).

### DIFF
--- a/.github/workflows/build_deploy_master_docs.yaml
+++ b/.github/workflows/build_deploy_master_docs.yaml
@@ -35,6 +35,7 @@ jobs:
         with:
           fetch-tags: "true"
           fetch-depth: '0'
+          persist-credentials: false
 
       - uses: ./.github/actions/load-shared-vars
 
--- a/.github/workflows/build_deploy_stable_docs.yaml
+++ b/.github/workflows/build_deploy_stable_docs.yaml
@@ -30,6 +30,7 @@ jobs:
         with:
           fetch-tags: "true"
           fetch-depth: '0'
+          persist-credentials: false
 
       - name: Set manual stable docs version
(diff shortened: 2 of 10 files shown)


d-chambers:
Coverage gap is unrelated (since this didnt touch any actual code)

PR #802, opened 2026-08-04
Convert 0-D arrays in the time converters

d-chambers:
## Description

`to_datetime64` and `to_timedelta64` raised `TypeError: len() of unsized object` for a 0-D array, unless its dtype happened to match the converter's own time type:

```python
dc.to_timedelta64(np.array(np.datetime64("2020-01-01")))  # TypeError
dc.to_datetime64(np.array(5.0))                           # TypeError
```

The empty-array guard is `len(array) == 0`, which a 0-D array cannot answer. Only the matching time type worked, because that branch returns before reaching the check. Every one of these has a 1-D equivalent that works fine, so the rank was the only thing standing in the way.

Both converters now reshape a 0-D input to the length-one array it stands for and unpack the scalar back out, so every rank takes the same path and a 0-D input converts exactly like `array.reshape(1)` would. `_array_to_timedelta64`'s early returns become an `if/elif/else` to give it a single exit to unpack at.

Nothing about 1-D or higher input changes. 2-D input still raises the same `ValueError` it always did.

Found while reviewing #800.

## Changelog

- fixed: `dc.to_datetime64` and `dc.to_timedelta64` convert a 0-D array instead of raising `TypeError: len() of unsized object`.
- fixed: `to_timedelta64` normalizes a `datetime64` array's unit before viewing it as integers, so a non-nanosecond array (e.g. `datetime64[s]`) is no longer off by a factor of a billion.

## Checklist

I have (if applicable):

### DIFF
--- a/dascore/utils/time.py
+++ b/dascore/utils/time.py
@@ -21,6 +21,17 @@ _NAT_TIMEDELTA64 = np.timedelta64("NaT", "ns")
 _EPOCH_DATETIME64 = np.datetime64(0, "ns")
 
 
+def _is_degenerate(array) -> bool:
+    """
+    Return True for a 0-D array.
+
+    The array converters index and iterate their input, which a 0-D array
+    supports for neither. Callers reshape it to the length-one array it
+    stands for, then unpack the scalar back out.
+    """
+    return array.ndim == 0
+
+
 def _float_array_to_ns(array):
     """Convert seconds as floats to signed integer nanoseconds."""
     # Integer inputs must be widened first; the default integer is only
@@ -126,6 +137,9 @@ def _float_to_datetime(num: float | int) -> np.datetime64:
 def _array_to_datetime64(array: np.ndarray) -> np.datetime64 | np.ndarray:
     """Convert an array of floating point timestamps to an array of np.datatime64."""
     array = np.asarray(array)
+    degenerate = _is_degenerate(array)
+    if degenerate:
+        array = array.reshape(1)
     nans = pd.isnull(array)
     # dealing with objects
     if np.issubdtype(array.dtype, np.dtype(object)):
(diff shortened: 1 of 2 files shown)


dascore/utils/time.py
```diff
@@ -126,6 +137,9 @@ def _float_to_datetime(num: float | int) -> np.datetime64:
 def _array_to_datetime64(array: np.ndarray) -> np.datetime64 | np.ndarray:
     """Convert an array of floating point timestamps to an array of np.datatime64."""
     array = np.asarray(array)
+    degenerate = _is_degenerate(array)
+    if degenerate:
+        array = array.reshape(1)
```
d-chambers: Too much code; just check if it is degenerate and delete the helper function.

dascore/utils/time.py
```diff
@@ -245,30 +257,30 @@ def _pass_time_delta(time_delta):
 def _array_to_timedelta64(array: np.ndarray) -> np.timedelta64 | np.ndarray:
     """Convert an array of floating point durations to np.timedelta64."""
     array = np.asarray(array)
+    degenerate = _is_degenerate(array)
+    if degenerate:
+        array = array.reshape(1)
```
d-chambers: same here

dascore/utils/time.py
d-chambers: Done in 0e6fd465 — inlined as `degenerate = array.ndim == 0` with a two-line comment, helper deleted.

dascore/utils/time.py
d-chambers: Done — same inline check, pointing at the note in `_array_to_datetime64` rather than repeating it.

dascore/utils/time.py
```diff
@@ -245,30 +257,30 @@ def _pass_time_delta(time_delta):
...
-    # immutable arrays. See #575.
-    if np.any(invalid):
-        array = np.array(array)
-        array[invalid] = 0
-    # inf/NaN complain, salience these types of warnings for this block.
-    with np.errstate(divide="ignore", invalid="ignore"):
-        out = _float_array_to_ns(array).astype("timedelta64[ns]")
-        out[invalid] = _NAT_TIMEDELTA64
-    return out
+        out = np.array(array.view(np.int64)).astype("timedelta64[ns]")
```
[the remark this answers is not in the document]
d-chambers: Correct, and worse than the 0-D case alone — the 1-D path had the same bug for every non-ns unit. `to_timedelta64(np.array([np.datetime64("2020-01-01", "D")]))` returned 18262 ns instead of 18262 days; only an array already in ns was right. Fixed in 0e6fd465 by casting to ns before the view, with a test parametrized over D/s/ms/ns asserting the epoch offset directly rather than against the other rank.

PR #822, opened 2026-08-05
Document sample range semantics and resolve unit ambiguity

d-chambers:
## Description

Follow-up to the tutorial review that produced #808. Those were outright errors; these are the places the docs are *correct but misleading*, which cost more reader time than the typos did.

### 1. Sample ranges are half-open and nothing said so

`docs/tutorial/patch.qmd` said "0 refers to the first sample along the dimension and -1 refers to the last." True for a scalar query, misleading for a range. Measured on this branch:

| query | selects |
|---|---|
| `select(distance=(0, 10))` | values 0 through 10, **both included** (11 channels) |
| `select(time=(0, 10), samples=True)` | samples 0 through 9 (10 samples) |
| `select(time=(0, -1), samples=True)` | 1999 of 2000 — every sample **except** the last |
| `select(time=-1, samples=True)` | only the last sample |

So value ranges are inclusive, sample ranges are half-open, and `-1` means opposite things depending on whether it is a scalar or a range endpoint. Added a callout with that table and extended the example with assertions that demonstrate it.

The `Patch.select` docstring had the same gap, so it gets the same note as a doctest.

These rules are not arbitrary — they are exactly xarray's, where `sel` is inclusive and `isel` is half-open. What DASCore loses relative to xarray is the *signal*: one method plus a keyword instead of two method names, and a `(a, b)` tuple that reads as a closed interval where a `slice` would not. Hence 4 below.

### 2. The concepts page asked a question and left it hanging

`docs/tutorial/concepts.qmd` says "the meaning of the values depends on the function, e.g. both frequency and period might make sense in the example above" — and then never answers it for its own example, `pass_filter(distance=(0.1, 0.2))`. Meanwhile `patch.qmd` uses `pass_filter(distance=(10*m, 100*m))` and calls it a wavelength, so a reader comparing the two pages gets contradictory signals.

Bare numbers on `distance` are wavenumber — specifically the inverse of whatever units that coordinate currently carries, not 1/m unconditionally. Verified:

```python
patch.pass_filter(distance=(0.01, 0.1)) == patch.pass_filter(distance=(10 * m, 100 * m))  # True
```

Now stated, with a runnable cell asserting the equivalence, plus a second assertion showing that a units-attached query survives `convert_units(distance="ft")` where a bare one does not.

### 3. One idiom for open intervals

The tutorials wrote open intervals three ways. Standardized `select`/`pass_filter` examples on `...`.

**Not** applied to `taper`: it requires `None` and raises `TypeError: unsupported operand type(s) for *: 'ellipsis' and 'int'` on `...`. That asymmetry is arguably worth fixing in the library, but not here.

### 4. Smaller `select` documentation gaps

`select_values_description` in `constants.py` listed "a Slice or a tuple of (min, max)" as though those were different things, when a slice is converted to a tuple on the first line of `sanitize_range_param`. It also omitted the plain-integer case entirely, despite that being the form the tutorial uses most. Both fixed. (That string composes into `Patch.select` and `CoordManager.select` only — not the `order` pair, which is just as well, since `order` reads a tuple as a sequence of values rather than a range.)

Also noted that `select` accepts a `slice` wherever it accepts a `(min, max)` tuple. This already worked; it is now stated as the preferred form for sample ranges, because `slice(0, -1)` makes the half-open behavior self-evident where `(0, -1)` does not — scoped to `select`/`order`, since `pass_filter` and `taper` reject slices, and with the caveat that a slice still needs `samples=True` to mean indices.

## Testing

- `pytest tests` — 8502 passed, 88 skipped, 2 xfailed
- `pytest dascore/proc/coords.py dascore/constants.py --doctest-modules` — 18 passed
- Every added/changed cell executed directly; all assertions hold
- `pre-commit` passes; new cross-references resolve

All 5 changed tutorial pages execute cleanly end to end (`patch.qmd` 44 cells, `concepts.qmd` 5, `spool.qmd` 15, `visualization.qmd` 4, `file_io.qmd` 10).

## Changelog

none

## Checklist

I have (if applicable):

## Summary by CodeRabbit

* **Documentation**
  * Clarified value- and sample-based selection, including ranges, slices, negative indexes, and open-ended bounds.
  * Added examples explaining inclusive value ranges versus half-open sample ranges.
  * Documented unit-aware frequency, wavenumber, wavelength, and distance filtering.
  * Updated tutorials to use ellipses for open-ended time and filter ranges.
  * Explained how sample-relative indexing differs from coordinate-value selection, including exclusive upper bounds and slice equivalence.

### DIFF
--- a/dascore/constants.py
+++ b/dascore/constants.py
@@ -184,10 +184,8 @@ Any dimension name can be passed as key, and the values can be:
     - an array of values to select, which must be a subset of the
       coordinate array.
     - an array of booleans of the same length as the coordinate where
-      `True` indicates values to keep.
-
-When `samples=True`, the values are indices along the coordinate rather
-than values in it.
+      `True` indicates values to keep. This form does not support
+      `samples=True`.
 """
 
 check_behavior_description = """
--- a/dascore/constants.py
+++ b/dascore/constants.py
@@ -187,12 +187,7 @@ Any dimension name can be passed as key, and the values can be:
       `True` indicates values to keep.
 
 When `samples=True`, the values are indices along the coordinate rather
-than values in it. Note that `samples` means an index here and in
-[`order`](`dascore.Patch.order`), but a *count* of samples in most other
-patch functions, such as [`rolling`](`dascore.Patch.rolling`),
-[`pad`](`dascore.Patch.pad`), and [`roll`](`dascore.Patch.roll`). For
-instance, `distance=10, samples=True` selects channel number 10 here, but
-means ten channels wide in [`rolling`](`dascore.Patch.rolling`).
+than values in it.
 """
 
 check_behavior_description = """
--- a/dascore/constants.py
+++ b/dascore/constants.py
@@ -178,12 +178,21 @@ for each attribute.
 
 select_values_description = """
 Any dimension name can be passed as key, and the values can be:
-    - a Slice or a tuple of (min, max) for that dimension.
+    - a tuple of (min, max) for that dimension, or an equivalent slice.
       `None` and ... both indicate open intervals.
+    - an integer, when `samples=True`, to select a single row or column.
     - an array of values to select, which must be a subset of the
       coordinate array.
     - an array of booleans of the same length as the coordinate where
       `True` indicates values to keep.
+
+When `samples=True`, the values are indices along the coordinate rather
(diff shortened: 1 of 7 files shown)


d-chambers:
@coderabbitai review

PR #831, opened 2026-08-06
Add Uptech Sensing HDF5 IO support

Shihao-Yuan:
## Description

Adds support for reading HDF5 files exported by the Uptech Sensing AS1000 DAS interrogator.

The added reader:

- Detects the `Acquisition/StrainRate` and `Acquisition/Time` datasets.
- Reads Uptech metadata such as acquisition frequency, gauge length, fiber length, and spatial sampling.
- Creates DASCore time and distance coordinates.
- Supports time and distance selection.
- Adds automatic format registration through `pyproject.toml`.
- Includes unit tests for format detection, reading, metadata, and selection.

No issue or discussion is currently linked.

## Checklist

I have (if applicable):

## Summary by CodeRabbit

- **New Features**
  - Added support for reading Uptech Sensing AS1000 HDF5 exports.
  - Added time- and distance-based data selection.
  - Added metadata scanning and coordinate handling for imported data.
  - Added support for empty selections, returning an empty result.
  - Registered a new sample data file for download.

- **Bug Fixes**
  - Unsupported HDF5 files without required identifying metadata are now rejected.
  - Invalid, irregular, or frequency-inconsistent acquisition timing is now detected and reported.


d-chambers:
Thanks for putting this together, @Shihao-Yuan! The reader itself is clean and follows the `FiberIO` patterns closely. Just a few things need to happen before we can merge.

## 1. We need a real test file (the main blocker)

Right now the tests fabricate an HDF5 file in `tmp_path` with hand-written attributes, so the reader is only ever validated against its own assumptions. 

Do you have the file you developed against? If so, the usual path works: shrink it with `scripts/modify_h5_file.py` (so it is about 5mb or less), add it to `dasdae/test_data`, register it in `data_registry.txt`, and add `UptechH5V1` to `COMMON_IO_READ_TESTS` in `tests/test_io/test_common_io.py`. That last step matters more than usual here — the `Acquisition/...` naming is close enough to ProdML that `test_all_other_files_arent_format` is worth having as a guard against detection collisions.

If you can't share the file, let me know and I can reach out to Uptech directly.

## 2. The `data_registry.txt` change is wrong

The added line is `DASDMSShot00_20230328155653619.das`, which is the *Sentek* sample and is already at line 19 — so the registry now contains a literal duplicate (`get_registry_df()` returns 48 rows with 1 duplicate). Looks like fallout from the "Resolve data registry conflict" commit. Please drop it; it should be replaced by an actual Uptech entry.

## 3. Is `sampling_interval` really the channel spacing?

`_get_coords` uses `sampling_interval` as the distance step while `spatial_resolution` is read but only stored as an attribute. In DAS usage an unqualified "sampling interval" normally means *time*, and if that's what it is here the distance axis would be silently wrong. The published Uptech deployment I could find (Muñoz et al., Sensors 2023, 23, 6599) confirms the two are physically distinct — 100 ns pulse gave 10 m spatial resolution, with 10 m channel spacing — but happens to have them equal, so it doesn't settle it.

Easy check once you have a real file: `fiber_length` should be roughly `n_channels * step`. Whichever attribute reproduces `fiber_length` is the channel spacing. (In the current fixture, 4 channels x 2.5 m = 10 m against a declared 12,800 m fiber, which suggests the numbers weren't taken from a real header.)

## 4. CI

`lint_code` and all 12 `test_code` jobs pass. Two things to note:

**Codecov is red, and it's a real gap.** Patch coverage is 97.83% (45/46 new lines hit); the single uncovered line is:

```python
# dascore/io/uptech/core.py:46
if not data.size:
    return dc.spool([])    # <- never exercised
```

`codecov/project` also fails because total misses go 8 -> 9 against a 99.93% baseline with no threshold configured, so this one line is what turns both checks red. A test that selects a distance or time range outside the data and asserts an empty spool comes back covers it and is worth having regardless — that path is easy to get wrong.

**The `test_code_min_deps` failures are not yours.** All six are a `requests.exceptions.ReadTimeout` from github.com during the pre-test data download (it died at `[11/48] prodml_2.1.h5`); the macOS 3.13 job failed and fail-fast cancelled the rest. A re-run should clear it.

## 5. Smaller items

- **Time is never validated.** `Acquisition/Time` is assumed to be float epoch seconds, and `acquisition_frequency` is never cross-checked against the actual step. A non-uniform or differently-scaled time array will silently produce a monotonic (non-`CoordRange`) coord, and `sampling_frequency` can end up contradicting the patch's own time coord.
-
- **No `data_units`** despite `data_type="strain_rate"` — most other readers set it.

- **Detection is attrs-on-dataset only.** If any firmware revision puts the metadata on the `Acquisition` group instead, `get_format` returns `False` rather than reading it. Worth a comment noting the assumption, and a second real file would confirm it.\
-
- **`fiber_length`** is required for detection but never used — fine as a fingerprint, just flagging it's load-bearing for `get_format`.
-
- **`__init__.py` docstring** — per `docs/contributing/new_format.qmd`, this should describe basic use of the format with a short example and note any non-obvious parser details (see `dascore/io/sentek/__init__.py` or `dascore/io/gdr/__init__.py`).
-
- **Tests should be grouped in a class** per our test conventions.

Nothing here is structural — the reader design is right, it just needs real data behind it. Happy to help with the test-data side if that's the sticking point.

d-chambers:
Thanks for the second round, @Shihao-Yuan — the real test file and the `sampling_interval` note are exactly what was needed. I went through it again and found one thing serious enough that I just pushed the fix to your branch rather than making you round-trip on it (bcb65ad5). Hope that's helpful rather than presumptuous; happy to revert any of it if you'd rather take a different approach.

## The time validator was rejecting valid files

`_get_time` compared pairwise diffs of `Acquisition/Time` with `rtol=1e-4`. The catch is that those are **float64 epoch seconds** — at a 2026 epoch (~1.78e9), `np.spacing` is 2.4e-7 s, so consecutive timestamps carry ~240 ns of representation jitter no matter how good the interrogator clock is. Meanwhile the tolerance is `1e-4 · step`, which *shrinks* as the sample rate rises. Feeding it exactly-uniform synthetic files (`t0 + arange(n)/freq`):

| epoch | 200 Hz | 250 Hz | 420 Hz | 450 Hz | 1 kHz | 10 kHz |
|---|---|---|---|---|---|---|
| 2026 | ok | ok | ok | **rejected** | **rejected** | **rejected** |
| post-2038 | ok | **rejected** | **rejected** | **rejected** | **rejected** | **rejected** |

The sample file only passes because 193.5 Hz sits just under the cliff (2.2x margin, which halves when float64 spacing doubles at 2^31 s in Jan 2038). And it fails quietly: `dc.scan`/`dc.spool` catch reader exceptions, so a directory of 1 kHz Uptech files would produce an **empty spool and one `UserWarning`**.

Worth noting the check was buying nothing — with it removed, `get_coord` ingests the jittered array and still produces the correct `CoordRange(step=0.005169152s)`. So the fix keeps the monotonicity check and the mean-step-vs-`acquisition_frequency` cross-check (loosened to `rtol=5e-2`, enough to catch a ms/us/ns unit mismatch without rejecting a rounded nominal rate) and lets `get_coord` decide even sampling. This branch was also the one codecov flagged as uncovered — the untested branch was the broken one.

## Still needs your input: is `sampling_interval` really the channel spacing?

The check I suggested last round doesn't settle it on the real file. With 81 channels and `fiber_length=12800`:

- `sampling_interval=2.5` → 202.5 m
- `spatial_resolution=2.0` → 162 m

Neither is anywhere near 12,800 m, so the export covers ~1.6% of the declared fiber and the heuristic gives no signal. Related: `_get_coords` hardcodes `start=0`, which assumes the export begins at the fiber head — if it's a windowed subset, the whole distance axis is offset. Can you confirm both with Uptech, or against a second export with a different channel count? I left the current interpretation in place since it's yours and documented, but a wrong choice here silently mis-scales every distance, so I'd like it nailed down before merge. **This is the one open item.**

## Also in the pushed commit

- `ruff-format` fix — `lint_code` should go green.
- Coverage: the module is now at **100%**. Added tests for empty selection, missing datasets, non-finite timestamps, irregular-but-preserved timing, and the frequency tolerance boundary.
- Added an `UptechPatchAttrs` subclass for the format-specific attrs, matching `OptoDASPatchAttrs` / `APSensingPatchAttrs`, so `*_units` become real `UnitQuantity` fields instead of untyped extras.
- Set `data_category="DAS"` (scan was emitting `''`).
- Dropped the `sampling_frequency` attr. Nothing updates it through `decimate`/`resample`, and it already disagreed with the patch on the real file: attr said 193.5, `patch.get_coord("time").step` is 5169152 ns → 193.436 Hz. The time coord is authoritative. Shout if you want the nominal value kept — `nominal_sampling_frequency` would be fine.
- Restored the `data_registry.txt` line order so the diff is a single added entry, and lowercased the `dasdae` org in the URL to match the neighbours (verified it still resolves).

Local checks on the pushed head: full suite **6497 passed, 84 skipped, 4 xfailed**; `pytest tests/test_io` 901 passed; doctests pass; `pre-commit run` clean.

## Minor things I left alone

- `float()` on the five attrs assumes numeric storage — a byte-string attr would raise `TypeError`. Other readers run values through `unbyte`. No evidence Uptech does this, so I didn't add speculative handling.
- `dc.to_datetime64` on a float treats it as epoch **seconds**. A firmware variant writing ms/us/ns would give silently absurd times; the mean-step check now catches that, but only because `acquisition_frequency` is there to compare against.
- Not from this PR, but noting it: `get_coord` infers the step from `diff[0]` (5169152 ns) rather than the endpoint fit (5169190 ns), leaving 74 us of accumulated error at the end of this 10 s file and growing ~27 ms/hour. Fine here, and a general `get_coord` question rather than a reader one.

The reader design is right and the rest is close. Just the `sampling_interval` confirmation left.

d-chambers:
I'm going to merge this for now. Let's take a look into the issue that I mentioned above, and if we find out it's wrong we'll come back and fix it, but I'd like to have this in master for the next release.

Shihao-Yuan:
Sounds good. Let’s merge it for the release and investigate the issue separately. If we confirm it’s wrong, we can fix it in a follow-up.
Thanks, Derrick. 

Shihao

Issue #832, opened 2026-08-07
Re-chunking the same dimension assembles one patch larger than its plan row


d-chambers:
## Description

Chunking the same dimension twice (the "collapse" re-plan path) assembles one patch far larger than the plan says it should be. The `ChunkPlan` is correct — every output row spans exactly the requested number of samples — but one *assembled* patch spans from the start of the whole spool instead of from its own planned envelope.

This is not specific to any argument form; it reproduces with plain numeric lengths on `dev`.

## Example

```python
import dascore as dc

sp = dc.get_example_spool("random_das")
first = sp.chunk(time=833 * 0.004)          # 833-sample patches
second = first.chunk(time=416 * 0.004)      # re-chunk the SAME dim

[x.shape[1] for x in second]
# [416, 416, 416, 416, 2082, 416, 416, 416, 416, 416, 416, 416, 416, 416]
#                     ^^^^ should be 416
```

The plan for that same call is uniform:

```python
plan = first.chunk_plan(time=416 * 0.004)
step = plan.outputs["time_step"].iloc[0]
spans = ((plan.outputs["time_max"] - plan.outputs["time_min"] + step) / step).round()
sorted(set(spans))          # [416.0]  -- every planned output is 416 samples
```

Output 4's planned envelope is `00:00:06.656 – 00:00:08.316`, but the assembled patch covers `00:00:00.000 – 00:00:08.316`.

Two observations that may localize it:

1. **Exactly one member row in the whole plan is marked `_modified=False`** (i.e. "load this member whole, no trim), and it belongs to exactly the one output that assembles wrong:

   ```
      output_id  _patch_id                time_min                time_max
   6          4          2 2020-01-03 00:00:06.664 2020-01-03 00:00:07.996
   ```

   Loading that member whole rather than trimming to the 1.332 s window accounts for the extra samples.

2. **The re-plan's `_patch_id` values live in a different namespace than either spool.** For the example above:

   ```
   member _patch_id values: [0, 1, 2, 3, 4, 5, 6, 7, 8]
   root   _patch_id values: [1, 2, 3]
   parent _patch_id values: [1, 2, 3, 4, 5, 6, 7]
   ```

   They look like positional indices assigned by `_ensure_patch_id` over the collapsed frame (which has 9 member rows), rather than ids of the patches they name. Whether that is intentional (the plan resolver may interpret them against the parent's member rows) or the source of the mis-trim, I did not determine.

## Expected behavior

[...]

PR #833, opened 2026-08-07
Chunk spools by data size and by unit-bearing lengths

d-chambers:
## Description

Adds a data-size chunk length so a memory budget can be stated directly:

```python
spool.chunk(time=25 * dc.units.megabytes)
```

Each output patch's **data array** is then at most 25 MB. `docs/recipes/low_freq_proc.qmd` hand-rolled exactly this (`pa.data.nbytes / pa.seconds` → a length in seconds, which needs a loaded patch and its dtype); that recipe now uses the new API.

The same change closes a nearby gap: `Spool.chunk` previously rejected *every* pint quantity, so `chunk(time=10 * dc.units.s)` raised `NotImplementedError` from `to_timedelta64`'s singledispatch. Unit-bearing lengths now work too, and `overlap` accepts both forms.

### How a size resolves

Per partition, since the conversion needs that partition's sampling interval, its extent along the other dimensions, and its element dtype:

```
bytes_per_sample = itemsize * (product of the other dimensions' sample counts)
n_samples        = floor(requested_bytes / bytes_per_sample)
length           = n_samples * min(|step|)
```

Three decisions worth reviewing:

- **Floored**, so the data never exceeds the request. When a single sample slab is already larger than the target, outputs hold one sample each and a warning says so — the request cannot be honored below one sample.
- **Sized against the partition's smallest step**, not `get_middle_value`'s median. Steps within `config.sampling_group_tolerance` (5%) share a partition, so sizing against the median lets a faster-sampled member fit more samples into the length and overshoot the byte bound. This makes the bound conservative instead of occasionally violated — and deterministic rather than flaky.
- **Mixed dtypes use `np.result_type`**, matching the upcast `patch_assembly` performs. dtype is deliberately *not* a partition key (making it one would change which patches merge), so a partition may legitimately mix element types. `result_type` is also strictly more accurate than max-itemsize: `int32 + float32 → float64`, itemsize 8 rather than 4.

### Index schema

Chunking must stay a pure metadata operation, so the element dtype comes from the index rather than from loading a patch. `PatchSummary.dtype` already existed and `scan_to_df` exposed it, but `patch_record()` dropped it on the way in. It is now stored (`INDEX_VERSION` 5 → 6) and reaches the planner as a private `_dtype` column.

The leading underscore is **load-bearing, not cosmetic**: `_police_columns` skips `_`-prefixed columns and nothing else protects it, so a public `dtype` column would raise `CoordMergeError` on every merge of patches with differing element types. There is a comment at the rename site saying so. `dtype` is also added to `RESERVED_ATTR_COLUMNS` so a patch attr of that name warns-and-drops instead of colliding.

The schema is unreleased (not even a beta), so the version bump costs only a local index rebuild.

### Units

pint models `byte = 8 * bit` with `bit` **dimensionless**, so `is_compatible_with("byte")` is `True` for `50%`, `1 strain`, and any bare dimensionless quantity. `is_data_size` therefore tests the base unit directly. `get_byte_count` converts with `.to("byte").magnitude` and never through `to_float`, whose fallback is `float(obj)` — pint converts a dimensionless quantity to base units, so `to_float(25 * MB)` returns `2e8`, the count in *bits*. A test pins that 8× difference so the two paths cannot be collapsed. (`to_float`'s own quantity handling is tightened separately in the follow-up PR.)

### Verification

Beyond the suite, checked against real data:

- a float32 Terra15 file-backed directory spool: dtype recorded correctly, `max nbytes 1999800 ≤ 2 MB`
- a partition mixing float32 and float64: sized against the upcast, bound held exactly at 100000 bytes
- three patches whose steps differ by ~4% (one partition under the 5% tolerance): bound held, which is the case the min-step decision exists for

### Known limitation

`chunk(dim=...).chunk(same_dim=...)` mis-assembles one patch, which breaks this bound in that specific chain. That is a **pre-existing bug**, filed as #832 — it reproduces on unmodified `dev` with plain numeric lengths and has nothing to do with sizes or units. `TestChainedChunk::test_size_then_size` therefore asserts on the plan (which is correct and is what this PR owns) and references #832; tighten it to assert `nbytes` once that is fixed.

## Changelog

- added: `Spool.chunk` accepts a quantity as the chunk length — a coordinate-unit length (`chunk(time=10 * dc.units.s)`) or a data size (`chunk(time=25 * dc.units.megabytes)`) bounding each output patch's array. `overlap` accepts both, and `chunk_plan(...).params["size"]` reports what a size resolved to.
- changed: the index records each patch's element dtype, so an attr named `dtype` is reserved and stays unindexed. Indexes built by an earlier unreleased version must be rebuilt, which the new index format already requires.

## Checklist

I have (if applicable):

- [x] referenced the GitHub issue this PR closes. (none closed; #832 is referenced as a pre-existing limitation)

## Summary by CodeRabbit

* **New Features**
  * Chunking and overlap support coordinate-based and data-size quantities with dtype-aware sizing and diagnostics.
  * Mixed data types are handled safely when combining patches.
  * Indexes preserve each patch’s element data type.

* **Bug Fixes**
  * Improved unit validation, size calculations, rounding, overlap handling, and descending-coordinate support.

* **Documentation**
  * Added guidance and examples for quantity-based chunking, data-size limits, and processing workflows.
  * Existing indexes may require rebuilding after the index schema update.


d-chambers:
## Update after adversarial review

Three internal reviews (correctness / index plumbing / test quality) found four real defects in this PR, now fixed in `e59e6d36` with a regression test each. Full suite 8392 passed.

**The size bound could be exceeded by up to 1.84×.** An output holds the *sum of its members'* sample counts, which equals `span/step + 1` only when the partition sits on one grid. Near-contiguous files whose boundaries miss the grid each contribute their own trailing sample, so they pack more samples into a span than the grid allows — silently, with no warning. Reproduced at 1.004×, 1.048× and 1.840× (30 files × 20 samples with a 10 ms gap; 60 files × 2 samples with a 50 ms gap), through a real directory index as well as in memory. Fixed by dividing the length by a measured **packing factor** — the partition's maximum local sample density in units of its step, exactly 1.0 when members tile the grid and reported in `params["size"]`. Those three cases now land at 96–99.6% of the request, and the gridded case is byte-identical to before.

**Correction to this PR's own description:** I claimed the bound depends on sizing against the partition's *smallest* step, and that the median would overshoot. That was true of the pre-packing code but is now wrong — packing is measured in units of whichever step is used, so the two cancel exactly and the length works out to `n_samples / max_density` either way. Mutation testing is what surfaced this: swapping min for median kills no test, because the code is genuinely insensitive to it. The comment and design note now describe packing as the mechanism; min-step remains only because it makes the flooring granularity finest.

**Element dtype is now resolved per output, not per partition.** Carrying the partition-wide `np.result_type` onto every output made a chunked spool compare **unequal to its own materialized twin** whenever the spool mixed dtypes — breaking the promise in `_strip_identity`'s docstring, and for plain `chunk(time=4)`, not just sizes. An output drawing only from float32 members really is float32; an output spanning a dtype boundary gets the upcast, which is what assembly produces. This also stops a chained size chunk under-filling.

**Non-finite and array-valued quantities are now rejected.** `chunk(time=nan*MB)` **silently merged the entire spool into one patch** (a NaN magnitude is null, so it read as merge mode) when the user asked for a size *cap*; `inf` raised a bare `ValueError` from the sample-count division. Both now raise `ParameterError`.

**NaN could reach the derived catalog as the string `"nan"`.** `np.nan` is truthy, so `str(row.get("_dtype") or "")` let it through, and `np.dtype("nan")` then discarded every good dtype in the partition.

### Test gaps closed

Mutation testing found `np.result_type` unguarded: both existing tests mixed float32 + float64, where it agrees with max-itemsize. The distinguishing case is `int32 + float32` — both 4 bytes, promoting to 8-byte float64 — where the max-itemsize rule silently produces **2× the requested budget**. Now tested, and each new test is confirmed to fail under its mutation. Also replaced `test_merge_mode_unaffected`, which passed with the entire feature absent, and `test_dtype_attr_is_reserved`, which restated a literal from the module under test rather than testing the behavior.

### Cross-PR note

This PR's `test_byte_count_is_not_to_float` asserted `to_float(25 MB) == 8*25e6`, which #834 makes raise — whichever merged second would have broken. Rewritten to assert the byte count directly; verified against simulated merged semantics.

PR #834, opened 2026-08-07
Handle pint quantities explicitly in to_float

d-chambers:
## Description

Follow-up to #833, split out so the index schema bump stays reviewable on its own.

`to_float` handled pint quantities backwards: it rejected everything *with* dimensions — including the seconds quantity its own docstring implies should work — and silently accepted everything *without*.

| input | before | after |
|---|---|---|
| `2 * s` | `DimensionalityError` | `2.0` |
| `2 * min` | `DimensionalityError` | `120.0` |
| `[1., 2.] * min` | `DimensionalityError` | `array([60., 120.])` |
| `25 * MB` | `200000000.0` | `UnitError` |
| `50%` | `0.5` | `UnitError` |
| `10 * m` | `DimensionalityError` | `UnitError` |

The silent half is the dangerous one. `to_float`'s fallback is `float(obj)`, and pint's `Quantity.__float__` converts a *dimensionless* quantity to base units. Bytes are dimensionless in pint (`byte = 8 * bit`, `bit` dimensionless), so a data size came back as its count in **bits** — eight times too large, with no exception and a plausible-looking number. It was found while adding size-based chunking, where an 8× overestimate of a memory budget is exactly the failure the feature exists to prevent.

### Why reject rather than pick a conversion

Two alternatives were considered and rejected:

- **Strip units (`.magnitude`)** is unit-blind: `to_float(10 * km)` and `to_float(10 * m)` would both give `10.0`, while callers treat the result as a number in canonical units (`proc/filter.py` compares it against Nyquist). It also makes `50%` → `50.0`, contradicting the percent-is-a-fraction convention in `maybe_convert_percent_to_fraction` and `coords._get_compatible_value`.
- **Convert to base units** is today's dimensionless behavior generalized — and bytes' base unit *is* the bit, so the 8× trap survives.

Instead, treat time as the function's domain, which is what every existing overload already does (datetime64 → seconds since epoch, timedelta64 → seconds, plain numbers passed through). A metre or a megabyte has no float representation in that domain, so it raises `UnitError` naming `convert_units` and `get_byte_count` as the explicit alternatives.

The magnitude is recursed back through `to_float` rather than returned directly, which is what makes array-valued quantities take the array path and an integer magnitude still widen to float.

### Is anything relying on the old behavior?

No. The full suite was run with `to_float` instrumented to record every call receiving a `Quantity`: **exactly one hit across 8382 tests**, and it was the test asserting the bits behavior itself. The only semantic loss is `to_float(get_quantity("50%")) == 0.5`, which nothing used and which is far more likely a bug than an intent when it reaches a seconds-converter.

### Drive-by

`patch.notch_filter(distance=5 * dc.units.m)` on a coordinate with **no units** leaked pint's `DimensionalityError` through `proc/filter.py`. It now raises `UnitError: Cannot filter 'distance' with 5 m: the coordinate has no units...`. The generic `to_float` message would have been misleading in a distance filter, hence the specific one.

## Changelog

- changed **breaking**: `to_float` converts a time quantity to seconds and raises `UnitError` for every other quantity, instead of silently reducing a dimensionless one to base units (`25 MB` came back as `2e8`, the count in bits); use `convert_units` or `get_byte_count` for an explicit conversion.
- fixed **breaking**: filtering a coordinate that carries no units with any quantity raises `UnitError`; a dimensionless one was previously read as a bare number (`20 %` became 0.2 Hz), so a call that silently did the wrong thing now stops.

## Checklist

I have (if applicable):

- [x] referenced the GitHub issue this PR closes. (no issue; follow-up to #833)

## Summary by CodeRabbit

* **New Features**
  * Added support for converting time quantities to seconds when using numeric time utilities.
  * Added clearer handling for unit-bearing values during filtering.

* **Bug Fixes**
  * Invalid unit combinations now raise a clear `UnitError` instead of proceeding silently or exposing a lower-level conversion error.
  * Non-time quantities, such as data sizes, are no longer implicitly converted to numeric values.

* **Documentation**
  * Updated the unreleased changelog with the new quantity-conversion and filtering behavior.


d-chambers:
## Update after adversarial review

Reviewed alongside #833; two findings addressed in `9366ad37`. Full suite 8323 passed.

**The notch-filter guard is broader than the changelog said, and its message was wrong.** It fires on *any* quantity when the coordinate has no units, including dimensionless ones — but the message claimed the value was "unit-bearing", which is false for `0.2 dimensionless`. The behavior is right and worth keeping: before this PR a dimensionless quantity was silently read as a bare number, so `20 %` became 0.2 Hz. The message now says a quantity cannot be interpreted against a unitless coordinate, the test is parametrized over metres / percent / dimensionless, and the changelog describes the real scope.

**The quantity contract was invisible in the rendered API docs.** It lived only in the private `_quantity_to_float` docstring, which the renderer does not surface; it is now on `to_float`'s public docstring.

### Verified clean

A reviewer instrumented `to_float` to record every `pint.Quantity` reaching it while emulating *pre-PR* semantics, then ran the full suite including the generated doc-code tests: **every hit originated in this PR's own new tests**, none in `dascore/` and none in any `docs/**/*.qmd` block. All 44 `to_float` call sites were also read individually — `units.py:146`, `utils/patch.py:1033`, `io/xml_binary/utils.py:242`, `coords.py`'s `_get_compatible_value` and tolerance handling, and `proc/filter.py` all strip units or return bare magnitudes before `to_float` sees them. Recursion termination was probed across 21 constructions (0-d, empty, masked, nested, timedelta64 and datetime64 magnitudes, foreign registries) with no unbounded recursion, and `.to("s")` was confirmed to raise only `DimensionalityError` — offset, log, angle and count units all produce a clean `UnitError`.

### Known, not fixed here

`to_float(Quantity(np.array(2.0), "s"))` raises `TypeError: len() of unsized object`. The cause is pre-existing in `_array_to_float`, which has the same hole for a bare 0-D array (`to_float(np.array(2.0))` fails on `dev` too), so #802 evidently did not cover `to_float`. It is hard to reach — pint downcasts 0-D magnitudes to `np.float64` on any arithmetic, so every natural spelling works — and fixing `_array_to_float` is a separate concern from this PR. Flagging rather than folding it in.

Issue #871, opened 2026-08-11
Re-chunking a selected chunked spool ignores the select, duplicating samples


d-chambers:
Chunk a spool, select a sub-range of the chunked view, then chunk the same dimension again: the second chunk loads more samples than the selection contains, and some sample values appear twice.

```python
import numpy as np
import dascore as dc
from dascore.units import m

pm = dc.get_example_patch().set_units(distance="m")
d = pm.get_coord("distance")
span = float(d.max() - d.min() + d.step)
pf = pm.update_coords(distance=d.data + span).set_units(distance="m")

first = dc.spool([pm, pf]).chunk(distance=200, conflict="keep_first", keep_partial=True)
selected = first.select(distance=(250 * m, 450 * m))
print(sum(p.shape[p.get_axis("distance")] for p in selected))   # 201 -- correct

second = selected.chunk(distance=None, conflict="keep_first")
patch = second[0]
values = np.asarray(patch.get_coord("distance").values)
print(len(values), len(np.unique(values)))                      # 252 201
```

The selection holds 201 samples; the re-chunked patch has 252, of which only 201 are distinct — 51 samples are duplicated, so the coordinate is no longer strictly increasing and the data array repeats a block.

Reproduces unchanged on dev (`541b25ae`), so it predates the native-units work; found while verifying that branch. The re-plan path is `collapse_working_df`, which rebuilds a working frame from the first plan's member rows and applies the view's residual ranges — a member whose row is not marked `_modified` then loads its whole source patch, and the parent residual alone does not narrow it to the intersection the second plan expects.

Worth checking whether the same shape of error affects re-chunking along a *different* dimension after a select, which shares the path.

d-chambers:
The select turns out not to be needed: re-chunking the same dimension duplicates samples on its own.

```python
import dascore as dc

patch = dc.get_example_patch("random_das")          # 2000 time samples
spool = dc.spool(patch)

sum(p.shape[1] for p in spool.chunk(time=3))                    # 1500 -- correct
sum(p.shape[1] for p in spool.chunk(time=2).chunk(time=3))      # 4500
```

Same `collapse_working_df` path and the same cause — a member whose row is not marked `_modified` loads its whole source patch — so this is the minimal reproduction rather than a separate defect. Reproduces on dev with no inventory or units involved.

[...]

Issue #857, opened 2026-08-11
Selecting on inventory-defined fields is not supported when an inventory is attached


d-chambers:
Once a spool carries an inventory, the fields that inventory defines should be selectable, and today they are not: `spool.attach_inventory(inv).select(coupling="cement")` raises `InvalidSpoolQueryError` because the index has no such column. The spec (`API/spool.qmd`) documents this workflow, so the gap is between the design and the implementation rather than in the design.

Two levels, with quite different mechanics:

**Acquisition-level (whole patch).** Fields of the acquisition or interrogator — `gauge_length`, `interrogator.model`. A patch matches or it does not. Resolution is pure metadata: for each distinct `data_source_id` and time window in the index, resolve the acquisition, test the predicate, keep or drop the row. No patch data is loaded and `len(spool)` stays exact.

**Channel-level (within a patch).** Track fields (`coupling.medium`), annotation groups (`zone`), geometry axes (`latitude`, `z`), and optical `distance`. These select a set of channels, which may be disjoint along the fiber, so a matching patch is trimmed and sometimes subdivided. That changes the number of patches, so it cannot be answered from the index alone; it needs a resolution pass over the inventory per row. `prune_to_inventory` (below) is the natural place for that pass, since it is already the documented one-time metadata cost.

**What comes back on the patch.** A name used in a select should arrive on the extracted patch, and nothing more: `select(coupling="cement")` yields patches carrying `coupling.medium` so the result says why each channel survived, without paying for blanket enrichment. Composed with an explicit `Spool.enrich`, the patch owes the union of what was asked for.

Related and not yet implemented:

- `Spool.prune_to_inventory(inventory=None)` — resolve the index against the inventory, drop rows it does not describe (`on_missing` of `raise`/`warn`/`drop`), and subdivide rows straddling an epoch boundary.
- Lazy attachment: `attach_inventory` should accept a path and not read it until something needs it, plus an auto-attach from a blessed name in a spool directory. This depends on the authoring-format loader, which does not exist yet.

`Spool.select` currently adds a note to its error naming the attached inventory and saying this is not supported yet, so the failure at least points somewhere useful.

d-chambers:
Small item to fold in whenever `Spool.enrich` is next touched, deliberately not done on its own because it is docstring-only and would re-run CI for nothing.

`Patch.enrich` composes its parameter documentation from shared fragments:

```python
@compose_docstring(
[...]

PR #882, opened 2026-08-12
Conform a spool to the inventory which describes it

d-chambers:
## Description

Phase 3 (b) of the inventory work (#857), on top of #877: `Spool.conform_to_inventory`, the one eager step of the inventory workflow.

Attaching an inventory promises nothing about the spool matching it. Conforming makes it so: every row is resolved now, patches the inventory does not describe are dropped, and a patch whose span crosses a change of optical path is subdivided into one patch per epoch — so the spool can grow as well as shrink, which is why the method proposed as `prune_to_inventory` is called `conform_to_inventory`. It is metadata work; no patch data is read.

```python
spool = dc.spool("/data/archive").attach_inventory(inventory)
raw = spool.conform_to_inventory().enrich()
```

**Subdivision is a plan, not a dataframe rewrite.** The derived-catalog machinery already models one source row feeding several outputs, so conform builds an outputs/members pair and hands it to `derived_catalog(..., mode="chunk")`. `_patch_id` stays a source identity on the members, read hints ride along through `PlanResolver._load_member`, and `len`/`get_contents` stay exact because the outputs *are* the contents rows.

**Splitting is exact along the sample grid.** An epoch boundary owes the sample grid nothing, so each piece opens at the first sample at or after its boundary. The obvious split — `[t_min, b - step]` and `[b, t_max]` — loses any sample falling between those two bounds, which for an arbitrary boundary is the common case rather than the edge one.

**A change of acquisition raises instead of subdividing.** Its two halves were recorded under different configurations, so no subdivision makes it one honest patch; `on_unresolved` has no say, since the inventory describes such a patch twice rather than not at all.

`resolve_contexts` now sits on the same epoch walk, which fixes it refusing a row that crosses an epoch bound nothing actually changes across — something `Patch.enrich` has always allowed, so the two no longer disagree about the same patch.

Not in scope, and unchanged: channel-level select/unselect and `split_by` are phase 3 (c), so conform subdivides along time only.

### Review notes

The full review pipeline ran before this PR (Codex plus five Claude reviewers). Two defects it found are worth calling out, since both would have been easy to ship:

- Three reviewers independently found that snapping a cut onto the sample grid could round an exactly-on-grid boundary a hair above its index, putting the boundary sample in the epoch *before* the boundary. The cause was converting both operands to float seconds with `to_float` before dividing — three roundings where one would do. (`CoordRange._get_index` divides the native types and is unaffected: measured over 200k random step/index pairs it is wrong 0 times, against ~21.5k for the `to_float` form. An earlier draft of this description blamed its fudge factor; that was wrong.) The ratio now only starts the search, and an exact comparison against the grid settles it either way it errs.
- Comparing two resolutions by object identity made conform refuse patches `Patch.enrich` accepts: a fiber array re-registered with a new description resolves to a fresh object whose acquisition and optical path say exactly what they said before. Resolutions are now compared by what they say, over the two fields that say anything about the patch.

While verifying composition I found that `chunk(time=2).chunk(time=3)` duplicates samples on dev with no inventory involved — the minimal form of #871, recorded there rather than fixed here. `chunk(...).conform_to_inventory()` is unaffected and tested; the reverse order walks into that pre-existing defect.

Closes nothing on its own; #857 stays open for phase 3 (c).

## Changelog

- added: `Spool.conform_to_inventory`, which resolves a spool against an attached DASDAE inventory: patches the inventory does not describe are dropped under an `on_unresolved` policy, and a patch spanning a change of optical path is subdivided into one patch per epoch.
- fixed: inventory-backed `Spool.select` refusing a patch whose span crosses an epoch bound across which nothing the inventory says about it changes; `Patch.enrich` already accepted such a patch.

## Checklist

I have:

I have (if applicable):

### DIFF
--- a/dascore/core/spool.py
+++ b/dascore/core/spool.py
@@ -38,6 +38,7 @@ from dascore.exceptions import (
     InvalidSpoolQueryError,
     MissingPatchError,
     ParameterError,
+    PatchError,
     UnresolvedPatchError,
 )
 from dascore.utils.chunk_plan import (
@@ -46,6 +47,7 @@ from dascore.utils.chunk_plan import (
     _combined_dtype,
     _ensure_patch_id,
     build_chunk_plan,
+    build_subdivision_plan,
     samples_adjusted_envelopes,
 )
 from dascore.utils.display import get_dascore_text, get_nice_text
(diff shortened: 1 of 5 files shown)


d-chambers:
@coderabbitai review

d-chambers:
Thanks — addressed in 6ca77f58.

**1. Cut order (`chunk_plan.py`)** — fixed, by sorting rather than documenting. Which cuts a row has is a *set*; the order carries no information, so a caller cannot meaningfully get it wrong and the helper should not be able to be broken by one. `_snapped_cuts` now returns them sorted, which is also where the deduplication already happens.

**2. `strict=` on the new `zip()` calls** — fixed, but the suggestion for `_epoch_changes` would have broken it. `zip(resolved, resolved[1:], boundaries, strict=True)` raises: there is one boundary between each consecutive pair, so `resolved` is always one longer than the other two. It now reads `zip(resolved[:-1], resolved[1:], boundaries, strict=True)`, which is what the equal-length claim actually requires. The other two sites took `strict=True` as-is.

(For the record, ruff as configured here does not report B905 — `uvx pre-commit run --all-files` is clean before and after. The change is worth making on its own merits, since the neighbouring code already passes `strict=True`.)

**3. Test placement** — the mismatch was real, but it was the docstring, so that is what changed. Those four cases are boundary policy: they are the rows that offer *no* usable boundary — no instants to place one against, or no identity to look one up with — and they were written from the same checklist as the rest of the class, before the implementation. `TestConformMembership` is about which patches survive a conform, which is a different question from where one can be cut. The class docstring now says what the class covers.

PR #895, opened 2026-08-13
Keep the Febus G1 sample window and fix two coordinate bugs it exposed

d-chambers:
## Description

Reading a directory of Febus G1 BSL files surfaced a cluster of problems, found while investigating why `spool.get_contents()['start_time']` came back as a float column rather than `datetime64`. Four are in the Febus G1 readers and the index; a fifth, found during review, is in coordinate merging and is the reason the new coordinate is worth having at all.

**The `start_time`/`end_time` attrs were epoch-second floats.** The G1 HDF5 attr map copied the root `start_time`/`end_time` header attrs straight into patch attrs. Both are bit-identical to `start_times[0]` and `end_times[-1]`, so they only restated the time datasets as untyped floats. They are now dropped, which is what the sibling A1 text path already does via `attr_exclude`. This also removes the confusing follow-on where those columns went `NaN` for every merged row under `spool.chunk(..., conflict='drop')` — they were per-file values that could not survive a merge. A test asserts the information is still recoverable from the coords, which is what justifies deleting rather than retyping them.

**The per-sample acquisition window was discarded.** The readers built the time coord from `start_times` and ignored the `end_times` dataset entirely, so a BSL patch presented as instantaneous samples when each value is really an average over a window — ~600 s in the data that prompted this, ~0.85 s in the test file, with a short dead gap before the next sample. Patches now carry a non-dimensional `sample_span` coord mapped to `time` holding that length.

The span is stored rather than the raw `end_times` on purpose. `start_times` and `end_times` are both near-regular, so each snaps to a `CoordRange` independently and they land on slightly different steps; differencing the two snapped coords turns the real per-sample jitter into a linear drift. Differencing the raw arrays first keeps it exact, and doing so in float seconds is more precise than differencing after the round to `datetime64[ns]`.

**`formatVersion` collided with a reserved index column.** It was mapped to a `format_version` attr, so every index build warned `Skipping reserved attr name 'format_version'` and the attr was left unqueryable. The same value is already reported as `source_version`, so the redundant mapping is removed.

**A stepless numeric coord's step became an int64 sentinel.** In `SQLIndexBackend._add_envelope_objects` the envelope column was assigned as a bare object array, letting pandas re-infer the dtype. When no numeric coord in a result carries a step, that array holds only `Timedelta`s and `None`, so it inferred `timedelta64` and turned the numeric nulls into `NaT`; `pd.to_numeric` in `_pivot_coords` then mapped those to `-9223372036854775808`. It is data-dependent rather than chunk-specific — chunking just tends to produce a small result where nothing anchors the column to a numeric dtype. Assigning an explicit object-dtype Series keeps the nulls as `NaN`. `_env_min`/`_env_max` shared the hazard.

**Merging replaced associated coords with the dimension coord.** `_get_merged_coords` used `dim` where it meant `coord_name`, so every coordinate mapped to the merge dimension had the dimension coord's values concatenated in place of its own. This is pre-existing and already corrupts `temperature` on any multi-file Febus G1 spool — after a chunk it comes back as datetime64 copies of `time` rather than temperatures. It would have silently destroyed `sample_span` in the same way, so the new coordinate is not meaningful without this fix. Snapping is now applied only to the dimension coordinate, since only it defines contiguity. It is a separate commit if you would rather it went in on its own.

### Notes

- `sample_span` is new user-visible state on Febus G1 HDF5 patches (both BSL and MTX, which share the coord builder). It slices with `select` and survives chunk/merge like `temperature`.
- It is deliberately not named `time_span`: a coord whose name starts with a dimension name shadows the `{dim}_{suffix}` envelope convention, and `update_coords(time_span_min=...)` died with an internal `ValueError: too many values to unpack` where other names raise a clean `CoordError`.
- `INDEX_VERSION` is intentionally not bumped — history shows it tracks schema changes, not reader metadata, and bumping would force every user to rebuild every index for a Febus-only change. Anyone with an existing index over Febus G1 data should rebuild it to pick up `sample_span` and drop the removed attrs.
- The reader now raises a named error when `start_times` and `end_times` disagree in shape; previously a truncated or mid-write file raised an opaque broadcast error, and a length-1 `end_times` broadcast silently into garbage spans.
- `start_time`/`end_time` were removed from the `VENDOR_ATTRS` allowlist in `test_common_io.py`; no reader emits them now, so leaving them would have let them back in unnoticed.
- One limitation worth knowing: a span array regular enough to look like a range is still normalized to a `CoordRange` by `get_coord_manager`, so exactness is guaranteed at the constructor but not through the coord manager. The comment says so rather than claiming more.

## Changelog

- fixed: The Febus G1 HDF5 readers no longer copy the redundant `start_time` and `end_time` header floats into patch attrs, where they restated the time coordinate as epoch seconds.
- added: Febus G1 HDF5 patches now carry a `sample_span` coordinate giving the length of each sample's acquisition window.
- fixed: The Febus G1 HDF5 readers no longer emit a `format_version` attr, which collided with a reserved index column and warned on every index build.
- fixed: The Febus G1 HDF5 readers now raise a clear error when a file's `start_times` and `end_times` datasets disagree in length.
- fixed: A numeric coordinate with no step no longer reports that step as a large negative integer instead of a null when no other numeric coordinate in the result carries one.
- fixed: Merging patches along a dimension no longer replaces the values of coordinates associated with that dimension, such as `temperature`, with the dimension coordinate's own values.

## Checklist

I have:

I have (if applicable):

## Summary by CodeRabbit

- **Documentation**
  - Expanded FEBUS format documentation with acquisition-window and timing-coordinate details.

- **Bug Fixes**
  - Added reliable `sample_span` timing coordinates for FEBUS G1 data.
  - Added validation for inconsistent timing datasets.
  - Prevented missing envelope values from being misinterpreted as timestamps.
  - Preserved associated coordinate values and data types during coordinate merges.
  - Preserved floating-point null values when loading indexed data.
  - Removed redundant timing and format metadata from patch attributes while retaining scan metadata.


d-chambers:
Good catch on the merge test — both points were right.

The two managers carried identical `quality` values (cm2 was derived from cm1 by offsetting time only), so concatenating the first manager's values twice would have passed. They now differ by a constant.

The more useful half was the snapping path: without a `snap_tolerance`, `_snap_coords` returns immediately, so the part of the change that restricts snapping to the dimension coordinate was never exercised. The test now merges with `snap_tolerance=1.3` and an offset that lands inside it.

Verified the strengthened test catches each half of the fix independently:

- reverting `coord_map[coord_name]` to `coord_map[dim]` → AssertionError on the values
- dropping the `coord_name == dim` guard around snapping → CoordMergeError, since a stepless coord has no step to build a tolerance from

Pushed in e957f1d7.

PR #901, opened 2026-08-14
Add a tunnel inventory recipe

d-chambers:
## Description

The [inventory tutorial](https://github.com/DASDAE/dascore/pull/899) shows the model and the verbs. This is the worked example: one deployment, from field notes to enriched patches.

The deployment is an underground tunnel — a 1500 m telemetry run in from an instrument room, a helically wound cable trenched along the floor past a buried coil, and three instrumented boreholes worked back along the tunnel. It comes from the survey drawing the DASDAE proposal site carried, which turns out to specify every distance and every coordinate; the recipe follows it rather than inventing its own.

### The tables are the point

Every track in the authoring format is a spreadsheet, so the page writes the CSV and renders it back rather than reciting twenty-three constructor calls a reader would skip. What stays as objects is what has no rows: the cables, enclosures, and interrogators, which share no fields, and the acquisitions.

The hardware is `resources/*.yaml`, the tracks are `optical_components.csv`, `geometry.csv`, `coupling.csv`, and `annotations.csv` under `path.00`, and the whole directory then loads with `dc.inventory`.

### What it documents that nothing else does

- **The wind angle stops at the geometry track.** The trench cable is wound off-axis, so 25 m of fiber covers 22.15 m of tunnel. Two control points say so, every channel between them is placed by interpolating that, and nothing downstream has to know about $\cos\phi$. The page asserts the interpolated position rather than describing it.
- **A repair is an epoch, not an edit.** A contractor puts a bucket through the trench cable fifteen meters of fiber in; it is fixed with two splices and a 2 m patch cord, and every channel past the repair moves. A second directory named `path.00@2024-09-01` records that, the first path runs until the second begins, and a patch recorded across the boundary is subdivided by `conform_to_inventory` rather than assigned to whichever epoch happened to be checked first. The same channel is ten meters down borehole 3 in one piece and eight in the other, which the page asserts.
- **The repair slack has no position.** The patch cord is coiled in a splice box, so the geometry describes the trench up to the break and again from it, and the two meters between get no coordinates. The page asserts that gap rather than describing it.
- **Selection can return more patches than it was given.** `select(section="borehole")` on a spool of one patch gives three, because the three holes are three disjoint runs of fiber and a patch is a dense block of channels.

### Executable, and derived rather than typed

Every `{python}` block runs under the doc-code tests. The components table is the only place a distance is written down: the other three tracks are generated from its running totals, so they cannot drift out of step with it, and the repair is a five-row edit to one table rather than forty hand-shifted distances across four.

The two figures come from the DASDAE inventory proposal site, which is being retired.

### Review

Codex reviewed it and found eight things, of which the sharpest was that the figures and the page described different tunnels — the drawing has a 1500 m lead-in and 15 m links where the page had invented 40 m and 5 m. The page was rebuilt on the drawing. It also caught the `distance_map` running out before the repaired path did, so the last two channels of the longer epoch resolved to nothing; the patch cord being described as buried trench fiber when the point of it is that it has no position; and three smaller claims that were not true of the implementation (`sequence` is not a model field, `Interrogator` has no firmware, and a file named plainly `.inventory` is not one of the accepted spellings).

## Changelog

- added: a [tunnel inventory recipe](https://dascore.org/recipes/tunnel_inventory.html) building a complete inventory for a deployment as an authoring directory of YAML and CSV files, then attaching, selecting, enriching, and conforming with it.

## Checklist

I have:

I have (if applicable):

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive tunnel-inventory recipe covering DAS/DSS deployment inventory creation and validation.
  * Documented hardware, fiber arrays, optical components, geometry, coupling, annotations, acquisition settings, repairs, serialization, and automatic inventory discovery.
  * Added the recipe to the documentation’s Recipes navigation under a new Inventory section.

### DIFF
--- a/docs/_static/borehole_fiber.svg
+++ b/docs/_static/borehole_fiber.svg
@@ -1,0 +1,491 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="140.14871mm"
+   height="235.66652mm"
+   viewBox="0 0 140.14871 235.66652"
+   version="1.1"
+   id="svg1"
+   inkscape:version="1.4.4 (1:1.4.4+202605061436+dcaf3e7d9e)"
+   sodipodi:docname="borehole_fiber.svg"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <sodipodi:namedview
+     id="namedview1"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:showpageshadow="2"
+     inkscape:pageopacity="0.0"
+     inkscape:pagecheckerboard="0"
+     inkscape:deskcolor="#d1d1d1"
+     inkscape:document-units="mm"
+     inkscape:zoom="1.0393815"
+     inkscape:cx="310.28068"
+     inkscape:cy="164.03987"
+     inkscape:window-width="1854"
+     inkscape:window-height="1011"
+     inkscape:window-x="66"
+     inkscape:window-y="32"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="layer1" />
+  <defs
+     id="defs1">
+    <inkscape:path-effect
+       effect="simplify"
+       id="path-effect4"
+       is_visible="true"
+       lpeversion="1.3"
+       threshold="30"
+       steps="5"
+       smooth_angles="360"
+       helper_size="10"
+       simplify_individual_paths="false"
+       simplify_just_coalesce="false" />
+    <inkscape:path-effect
+       effect="simplify"
+       id="path-effect3"
+       is_visible="true"
+       lpeversion="1.3"
+       threshold="30"
+       steps="1"
+       smooth_angles="360"
+       helper_size="10"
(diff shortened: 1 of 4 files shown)


d-chambers:
All eight inline comments were written against `e6cd27fe`, the first push. The page was rebuilt in `3a007f48` and `d3862a8c` after a local review made the same catch the diagram comment did — the figures specified a different deployment than the page — so each of them is resolved, though mostly by the ground moving rather than by a targeted fix.

**The diagram against the inventory.** Taken, and it was the important one. The drawing is not an illustration: it is a survey, with lettered waypoints, a 1500 m telemetry lead-in, 15 m links between every feature, and coordinates on the tunnel's own grid. Every distance in the page now comes from it, and the winding factor is the 0.886 its coordinates are computed with rather than the 0.866 the old prose claimed.

**The coil annotation, and the repair slack in the trench coupling.** Both gone by construction. The three interval tracks are now generated from the components table's running totals rather than written against it, so a component whose distance moves takes its coupling and its annotation with it. The patch cord is excluded from the trench parts, and channel 1518 — inside the repair — now comes out with no `section` and a `nan` position, which is the point of it.

**The distance map bounded before the repaired path.** Taken. The map now runs to 2000 m, past both epochs, and the page says why: a map that stops where today's fiber does stops working the day the fiber gets longer.

**The trench out of reach of borehole 3.** Resolved by the survey. The trench is between boxes A and E, and the boreholes are worked back along the tunnel afterwards, which is the order the drawing shows; there is no longer a 60 m jump between consecutive components.

**Both DSS findings** — the single-fiber trench cable, and `path.01` having no geometry. The second optical path is gone. The drawing shows one fiber route, and a DSS path nobody surveyed would have been the same fiction in a different place. That a second interrogator would be a second path at its own location code is now one sentence of prose, with no invented data under it.

Issue #905, opened 2026-08-15
Inventory round trip loses a deliberately-blank field, reloading it as its default


d-chambers:
## Description

Serializing an inventory drops fields whose value is empty, so that reloading gives back the field's *default*. That is lossless only when the default is itself the empty value — and for four fields it is not. The most consequential is `CoordinateReferenceSystem`: a frame deliberately left without an authority code (a local or derived CRS described by WKT) reloads as **EPSG:4979 / WGS 84 3D**, silently changing what every coordinate in the inventory means.

The pruning happens in `_drop_empty` (`dascore/core/inventory.py`), which walks the dumped mapping by key without knowing which model each mapping came from, so it cannot tell "empty and defaults to empty" from "empty and defaults to something else".

Fields with a non-empty string default, and so affected:

| Model | Field | Default |
|---|---|---|
| `CoordinateReferenceSystem` | `authority` | `"EPSG"` |
| `CoordinateReferenceSystem` | `code` | `"4979"` |
| `CoordinateReferenceSystem` | `name` | `"WGS 84 3D"` |
| `Interrogator` | `instrument_type` | `"interrogator"` |

(`object_type` on the resource models has the same shape but is a discriminator tag, never blanked by a caller.)

This is long-standing, not a regression — it reproduces identically on `dev` before the inventory-model-trims branch. `_UNPRUNED_KEYS` was an exemption list for exactly this hazard but only ever held `"value"`; that branch removes it because the one entry had become unreachable (an annotation `value` cannot be the empty string — a validator refuses it), and corrects the docstring to state this limitation instead of claiming losslessness.

## Example

```python
import dascore as dc
import dascore.core.inventory as inv

# a local frame described by WKT, deliberately with no authority code
crs = inv.CoordinateReferenceSystem(
    authority="", code="", name="", wkt='LOCAL_CS["mine"]',
    coordinate_labels=("x", "y"), units=("m", "m"),
)
one = inv.Inventory(coordinate_reference_system=crs)
back = dc.inventory(one.to_yaml()).coordinate_reference_system

print((crs.authority, crs.code, crs.name))    # ('', '', '')
print((back.authority, back.code, back.name)) # ('EPSG', '4979', 'WGS 84 3D')

# same shape, second field
i = inv.Interrogator(resource_id="i1", instrument_type="")
two = inv.Inventory(resources={"i1": i})
print(dc.inventory(two.to_yaml()).resources["i1"].instrument_type)  # 'interrogator'
```

[...]

PR #906, opened 2026-08-15
Trim the inventory model

d-chambers:
## Description

Fifth PR of the pre-release simplification roadmap: trimming the inventory model.

**One value per component.** `loss_db` and `reflectance_db` (and their measurement fields) held either a scalar or a tuple of per-wavelength values. Several wavelengths are several components' worth of facts with no channel to give them to, so the tuple form is gone — and with it `_check_measurement_pairing`, the "many-shaped field" tracking in `get_names` (a field stated as several values had to be dropped from the names an inventory contributes), the tuple branch of `attenuation_db_per_km`, and the tuple branch of `_normalize_resources`, which only those measurement tuples ever reached.

**The vocabularies are derived, not listed.** A field's own type already says whether it holds a value or a reference, and each track model already declares its identity field, so nothing needs to restate them. This PR goes one step further than `dev`: `_normalize_resources` carried a hand-written `{model: {field: allowed types}}` dict, which is now computed from the annotations. I verified the derived version reproduces the hand-written one exactly. That also makes the drift test the roadmap asked for unnecessary — a resource-valued field cannot be missing from a list computed from the fields.

The one pairing still pinned by a test is `INVENTORY_ATTRS` in `constants.py` against the model-derived names: that vocabulary genuinely lives in two places, since readers write it without importing the models, so it can only be checked rather than derived. `DATA_STATE_ATTRS` moved to `constants.py` beside it, where `proc/inventory.py` had been keeping a private copy.

**Deletions:** the `_reject_start_distance` migration shim (for a field that never shipped), a one-element list in `Inventory.resolve`, three `CoordinateReferenceSystem` validators merged into one `_check_axes` (they were three statements about one thing — what this CRS's axes are), and `_UNPRUNED_KEYS`, whose hazard a later validator had already made unreachable.

**Deleted, but not as a claim of losslessness.** `_drop_empty` prunes by key without knowing which model a mapping came from, so a field whose default is *not* the empty value reloads as that default. That is a real, pre-existing bug — a CRS deliberately left without an authority comes back as EPSG:4979 WGS 84 — filed as #905 with a reproduction and a suggested round-trip property test. `_UNPRUNED_KEYS` was the exemption list for exactly that hazard but only ever held `"value"`, whose empty form a validator refuses; the docstring now states the limitation rather than claiming pruning is lossless.

Source is −45 lines, tests −65.

### Review round

Six reviewers; four findings were mine and worth naming, because two were invisible to a green suite:

- The `_SYSTEM_FACT_NAMES` revert **silently did not apply** — and the check I ran to confirm it compared the resulting value against the constants, which both the derived and the literal form satisfy. I verified the value, not the derivation.
- `DascoreBaseModel._cache` was **not unused**: it is the pydantic private-attr declaration that keeps `cached_method`'s dict out of `__dict__`, where a `deep_equality_check` over `__dict__` sees two equal models differ by which one has been asked a cached question.
- The derived reference table **could not see through a collection** — exactly `OpticalPath.measurements`' shape, and why a hand-written row survived in `norm_path`. It sees through now and the row is gone.
- The tests meant to pin the trims mostly did not: one of four scalarized fields covered, an annotation-value test that passed against a `_drop_empty` replaced by the identity function, one reference row no other test constructs, and a guard whose test I had deleted. Each is now controlled — verified by breaking the thing it guards and watching it fail.

### A note on two roadmap bullets

The roadmap asked to replace the annotation-derived vocabularies with literal tables, and to add a drift test for the resource table. That was done first and it was worse: the tables plus the reflection needed to notice them going stale came to ~123 lines *that could diverge*, replacing ~64 *that could not*. Both bullets are reversed here, and the roadmap now records the generalisation — a bullet asking for a test to pin something is usually a signal that the something wants restructuring so the test is unnecessary. `_UNPRUNED_KEYS` was the third instance of the same shape: the behavior was unreachable, so the test could not have failed.

## Changelog

- changed **breaking**: optical component `loss_db` and `reflectance_db` (and their `_measurement` fields) hold one value rather than a tuple of per-wavelength values; `attenuation_db_per_km` returns one value accordingly.
- changed **breaking**: `Acquisition` no longer explains the removed `start_distance` field; an unknown field is refused as an unknown field.
- added: `dascore.constants.DATA_STATE_ATTRS` names the attrs which describe the data as it stands rather than the system which recorded it.

## Checklist

I have:

I have (if applicable):

## Summary by CodeRabbit

- **Improvements**
  - Improved inventory validation and consistency for coordinate reference systems, resources, tracks, and acquisition data.
  - Optical loss, reflectance, and attenuation values now accept a single measurement value, providing clearer and more predictable validation.
  - Resource references are normalized consistently across inventory paths and nested specifications.
  - Acquisition resolution and metadata handling are more reliable, including improved support for track identity fields.
- **Bug Fixes**
  - Fixed handling of invalid non-numeric inventory values and empty annotation values.
  - Improved discovery of coordinate-related fields, including fields containing collections.

### DIFF
--- a/dascore/constants.py
+++ b/dascore/constants.py
@@ -134,6 +134,13 @@ max_lens = {
 # inventory's: seconds, hertz, and meters. Readers convert at the parse
 # boundary instead of shipping a companion units attr.
 # Tested against the models in tests/test_core/test_attrs.py.
+# Attrs describing the data as it now stands rather than the system which
+# recorded it. Processing maintains them, so blanket enrichment leaves
+# them alone; naming one restores the as-acquired value. Kept beside
+# INVENTORY_ATTRS because the two together are what an inventory can say
+# about a patch.
+DATA_STATE_ATTRS = ("data_type", "data_category", "data_units")
+
 INVENTORY_ATTRS = (
     "closed_fiber_loop",
     "firmware_version",
--- a/dascore/constants.py
+++ b/dascore/constants.py
@@ -134,13 +134,6 @@ max_lens = {
 # inventory's: seconds, hertz, and meters. Readers convert at the parse
 # boundary instead of shipping a companion units attr.
 # Tested against the models in tests/test_core/test_attrs.py.
-# Attrs describing the data as it now stands rather than the system which
-# recorded it. Processing maintains them, so blanket enrichment leaves
-# them alone; naming one restores the as-acquired value. Kept beside
-# INVENTORY_ATTRS because the two together are what an inventory can say
-# about a patch.
-DATA_STATE_ATTRS = ("data_type", "data_category", "data_units")
-
 INVENTORY_ATTRS = (
     "closed_fiber_loop",
     "firmware_version",
@@ -157,6 +150,13 @@ INVENTORY_ATTRS = (
     "software_version",
     "spatial_interval",
 )
+
+# Attrs describing the data as it now stands rather than the system which
+# recorded it. Processing maintains them, so blanket enrichment leaves them
+# alone; naming one restores the as-acquired value. Together with
+# INVENTORY_ATTRS these are what an inventory can say about a patch.
+DATA_STATE_ATTRS = ("data_type", "data_category", "data_units")
+
 
 # Methods FileFormatter needs to support
 FILE_FORMATTER_METHODS = ("read", "write", "get_format", "scan")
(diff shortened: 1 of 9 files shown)


d-chambers:
Actually, I think I am going to reject this. The tuple form of reflectance and loss were part of the original design spec, and the other deletions are not very substantial.

PR #910, opened 2026-08-15
Let a geometry state named numeric columns

d-chambers:
## Description

An optical path could state a value varying along distance only when that value was a CRS position, and a CRS holds at most three axes. Borehole depth where the CRS is already spent on easting/northing/elevation, pipeline chainage, burial depth, fiber azimuth — all of them fell in the gap, and the only place left for them was `annotations`, which is a set of intervals rather than a curve and carries no units.

`Geometry.coordinates` becomes a mapping of column name to values, with a `units` mapping beside it for the columns which are not axes.

```csv
segment,distance,depth (m),azimuth (degree)
hole 1,1580.0,0.0,0.0
hole 1,1620.0,20.0,3.5
```

**Which columns are axes is the CRS's to say.** A column it declares, or the canonical `x`/`y`/`z` alias of one, is that position axis and takes the CRS's units. Anything else is a quantity in its own right. So `depth` is an axis in a CRS which declares it and a plain column in one which does not, and no data has to move for that to be true.

**A segment states every axis or none of them.** Half a position is not a position, and deciding what the missing axis meant is not a reader's job. A segment stating no axis at all — the chainage case — is now an ordinary thing to write, and contributes no position rather than a broken one.

**Each column is its own function track.** Two segments may cover the same distance as long as they do not state the same column over it, which is what lets depth and azimuth be surveyed independently along one hole. Interpolation, half-open coverage, the run-end rule, and `interval` are unchanged, and a column never bridges two segments: distance between them is uncovered, whatever either side holds.

### Reading and writing

In a CSV, a header the CRS does not name is such a column and may carry its units in parentheses, `depth (m)`. Text there is refused with an error pointing at `annotations.csv` — a value which varies along the fiber without being a number is what annotations are for. A unit suffix on an axis header is refused too, since the CRS states the units of its own axes.

Enrichment resolves a column after the CRS-label step and before the annotation fallback, so `spool.select(depth=(0, 50))`, a blanket `enrich()`, and `get_names` all pick it up with no further work. The value is a point sample at the channel's nominal position; there is no gauge-length averaging, and the docs say so.

### Signature change

`OpticalPath.coordinates_at(distances)` becomes `coordinates_at(distances, crs)`. Without the CRS the method cannot tell an axis from a column, and guessing is exactly what this PR removes. It also now raises rather than returning a half-width row when an unchecked path holds a segment stating some axes and not others.

### v1 exclusions

Stated so they are choices rather than oversights: axes are all-or-none per segment; interpolation is linear only; there is no angular wraparound for azimuth-like columns. FDSN export is not addressed because none exists — the only `fdsn` in the inventory module is the prose describing location codes.

### Not in this PR

`Patch.coords_from_df` (`dascore/proc/coords.py`) is the manual version of this projection, and its gap-bridging across missing spans is worth a separate look.

**This PR is red for a reason that is not in it: #912.** `tests/test_autogenerated_doccode/recipes/test_tunnel_inventory.py` fails on `dev` and therefore here, on every `test_code` job and in `test_build_docs`. `OpticalPathAnnotation.value` defaults to `True`, and `1 == True`, so serialization drops a numeric value of `1` as equal to the default; the document then reloads with a boolean in a numeric group and is refused. The tunnel recipe numbers its boreholes 1, 2, 3, so it has been failing since it merged. #912 fixes that in one place; with it applied, the full suite here is 10,209 passed and 0 failed.

### Review

Codex reviewed the change and found eight things, taken in `c2a728c7`. Two of them placed the fiber wrongly rather than loudly: a segment stating `x`/`y`/`z` and another stating `longitude`/`latitude`/`elevation` over the same distance are two spellings of one axis, and checking columns by name alone let the overlap through; and a segment stating `x` *and* `longitude` has three distinct axes, so it passed the partial-position guard, after which whichever spelling the mapping held last silently won. Both are now refused, the second where the position is assembled rather than only where it is checked.

The rest: a segment which measures without placing no longer claims the position track's run end, or offers an axis of `nan` where nothing places the fiber at all; a dotted column name is refused, since that is how a field of a typed track is asked for; `select` and `reverse` revalidate rather than copying past the validators, so a selected segment's columns stay frozen and checked; `distance (m)` beside `distance` is a duplicate rather than a pandas `TypeError`; and `z` is a column of its own under a CRS declaring two axes, in a CSV as it already was in the model.

## Changelog

- added: a geometry segment can state named numeric columns which are not positions, such as borehole depth or chainage, with their own units, and they reach a patch as coordinates like any other inventory name.
- changed: `Geometry.coordinates` is a mapping of column name to values rather than a tuple of coordinate rows, and `OpticalPath.coordinates_at` takes the coordinate reference system as a second argument.

## Checklist

I have:

I have (if applicable):

### DIFF
--- a/dascore/core/_spool_inventory.py
+++ b/dascore/core/_spool_inventory.py
@@ -671,9 +671,9 @@ def _get_geometry_coord(inventory, path, label, distances):
     # on_missing policy rules rather than a column of nan.
     if not any(axis_columns(x, crs) for x in path.geometry):
         return None
+    # axis_index refuses a label this CRS has no axis for, and the array is
+    # one column per axis, so the index is always one of its columns.
     coords = path.coordinates_at(distances, crs)
-    if index >= coords.shape[1]:
-        return None
     return get_coord(data=coords[:, index], units=crs.units[index])
 
 
--- a/dascore/core/_spool_inventory.py
+++ b/dascore/core/_spool_inventory.py
@@ -668,10 +668,18 @@ def _get_geometry_coord(inventory, path, label, distances):
         return None
     if not path.geometry:
         return None
-    coords = path.coordinates_at(distances)
+    coords = path.coordinates_at(distances, crs)
     if index >= coords.shape[1]:
         return None
     return get_coord(data=coords[:, index], units=crs.units[index])
+
+
+def _get_geometry_column_coord(path, name, distances):
+    """Return one geometry column which is not a position, with its units."""
+    values = path.column_at(name, distances)
+    if values is None:
+        return None
+    return get_coord(data=values, units=path.column_units(name) or None)
 
 
 def get_coord_values(inventory, path, name, distances):
@@ -688,7 +696,13 @@ def get_coord_values(inventory, path, name, distances):
             path, track, field or TRACK_IDENTITY_FIELDS[track], distances
         )
     if name in VALID_COORDINATE_LABELS:
-        return _get_geometry_coord(inventory, path, name, distances)
+        # A coordinate label this CRS does not declare is not an axis here,
+        # and is free to be a geometry column of its own -- depth, where the
+        # CRS is spent on easting, northing, and elevation.
+        if (coord := _get_geometry_coord(inventory, path, name, distances)) is not None:
+            return coord
+    if (coord := _get_geometry_column_coord(path, name, distances)) is not None:
+        return coord
     return _get_annotation_coord(path, name, distances)
 
 
--- a/dascore/core/_spool_inventory.py
+++ b/dascore/core/_spool_inventory.py
@@ -682,7 +682,10 @@ def _get_geometry_column_coord(path, name, distances):
     values = path.column_at(name, distances)
     if values is None:
         return None
-    return get_coord(data=values, units=path.column_units(name) or None)
+    # The path has already refused two segments measuring one column in two
+    # units, so whichever states it states the one they agree on.
+    units = next((x.units[name] for x in path.geometry if name in x.units), None)
+    return get_coord(data=values, units=units)
 
 
 def get_coord_values(inventory, path, name, distances):
--- a/dascore/core/_spool_inventory.py
+++ b/dascore/core/_spool_inventory.py
@@ -34,6 +34,7 @@ from dascore.core.inventory import (
     Inventory,
     ResolvedContext,
     _annotation_kind,
+    axis_columns,
     interval_masks,
 )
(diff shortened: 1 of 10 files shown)


d-chambers:
Thanks — all eleven comments were written against `087386db`, the first commit. Eight of them were the same eight a local Codex run found, and were closed in `c2a728c7` before these arrived; one was new and real, and is closed in `674d94dc`.

**New, and taken: the bare `geometry` identity was ambiguous.** Allowing two segments to overlap where they state different columns left the `geometry` coordinate — which is each segment's *name* over its interval — with two candidates for one channel, and the tuple's order decided it:

```python
patch.enrich(inv, coords=("geometry",)).get_coord("geometry").values[0]
# 'azimuth survey' with geometry=(depth, azim); 'depth survey' with geometry=(azim, depth)
```

Rather than take the overlap away — a depth survey and an azimuth survey of one borehole is exactly what it is for — overlapping segments now have to share a name. They are two measurements of one stretch of fiber, so they state its name: both `"hole 1"`, not `"depth survey"` and `"azimuth survey"`. The identity is single-valued again and the overlap still works.

**Already closed in `c2a728c7`**, each verified against the current branch:

- *Aliases of one axis overlapping* — `x`/`y`/`z` against `longitude`/`latitude`/`elevation` is now caught, by checking the axes again through the CRS rather than by name.
- *Absent position axes* — a path whose geometry states only columns returns `None` for `x`, so `on_missing` rules, and a blanket enrich no longer manufactures all-NaN axes.
- *Endpoint masks* — the position track's coverage is computed from the segments which place the fiber, so a column-only segment starting where one ends no longer takes its last control point.
- *Canonical aliases in a CSV* — the loader decides axis-hood with `crs.axis_index`, so `x,northing` under an `easting,northing` CRS loads, and `z (m)` under a two-axis CRS is a column of its own rather than a refused axis.
- *Frozen mappings after `select`/`reverse`* — both revalidate now instead of copying past the validators, so `coordinates` stays a `FrozenDict`.
- *Qualified typed-track names* — a column named `coupling.medium` is refused: a dotted name is how a field of a typed track is asked for.
- *`distance (m)` beside `distance`* — counted against the structural columns, so it is a duplicate with the loader's own message rather than a pandas error. The same check covers `distance (ft)` on its own, which would otherwise have renamed onto the optical-distance column and quietly discarded the unit.

Local state on this branch: 10,210 passed, lint clean twice, doctests 155 passed. The one remaining failure is the tunnel recipe's doc-code test, which fails on `dev` too and is fixed by #912.

PR #912, opened 2026-08-15
Keep an annotation value of 1 through serialization

d-chambers:
## Description

`tests/test_autogenerated_doccode/recipes/test_tunnel_inventory.py` has been failing on `dev` since the tunnel recipe merged in #901, which also means `TestDocBuild` is red for anything branched from it. The cause is not in the recipe.

`OpticalPathAnnotation.value` defaults to `True`, so a bare flag — `noisy` over an interval — needs no value written. `Inventory.to_yaml` leaves out any field still holding its default. But `1 == True` in Python, and pydantic's `exclude_defaults` compares with `==`, so an annotation whose value is the number `1` is dropped as though it were the flag default:

```python
annotation = OpticalPathAnnotation(start_distance=0.0, end_distance=10.0, group="borehole", value=1)
# ...in an inventory:
text = inventory.to_yaml()          # no `value:` line for this annotation
dc.inventory(text)                  # InvalidInventoryError: group 'borehole'
                                    # mixes ['boolean', 'numeric'] values
```

Reloading gives the annotation `True`, the group then holds a boolean beside the numbers next to it, and the group-holds-one-kind rule refuses the whole document. Any group numbered from one — boreholes 1, 2, 3, which is what the recipe writes — cannot survive a round trip.

Identity, not equality, is what "still its default" means for a field whose type admits both `bool` and `int`, so a wrap serializer puts the value back unless it is `True` itself. A flag annotation still writes no value, so documents which use the default stay as terse as they were.

This is the same shape as the existing fix for the `object_type` tag in `dascore/models/base.py`, which `exclude_defaults` also drops and which is also put back by a wrap serializer.

## Changelog

- fixed: an optical path annotation whose value is the number `1` is no longer dropped when an inventory is serialized, which had made any annotation group numbered from one fail to reload.

## Checklist

I have:

I have (if applicable):

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed serialization of optical path annotations so numeric values, including `1`, are preserved correctly.
  * Ensured annotation type information survives YAML round trips.
  * Maintained support for explicit field inclusion and exclusion during serialization.
  * Default-valued boolean fields continue to be omitted when appropriate.

* **Tests**
  * Added regression coverage for annotation serialization, filtering, round trips, and invalid mixed value types.

### DIFF
--- a/dascore/core/inventory.py
+++ b/dascore/core/inventory.py
@@ -37,6 +37,7 @@ from pydantic import (
     BeforeValidator,
     Field,
     field_validator,
+    model_serializer,
     model_validator,
 )
 from typing_extensions import Self
@@ -696,6 +697,21 @@ class OpticalPathAnnotation(_IntervalModel):
             )
             raise ValueError(msg)
         return value
+
+    @model_serializer(mode="wrap")
+    def _keep_the_value(self, handler, info):
+        """
+        Put back a value which is the default only by ``==``.
+
+        ``exclude_defaults`` compares with ``==``, and ``1 == True``, so a
+        group numbered from one loses every ``1`` on the way out and reloads
+        holding a boolean -- which then mixes kinds with the numbers beside
+        it and is refused. Identity is what "still its default" means here.
+        """
+        out = handler(self)
+        if "value" not in out and self.value is not True:
+            out["value"] = self.value
+        return out
 
 
 # The coordinates a DistanceMap may be written in, in preference order.
--- a/dascore/core/inventory.py
+++ b/dascore/core/inventory.py
@@ -37,7 +37,6 @@ from pydantic import (
     BeforeValidator,
     Field,
     field_validator,
-    model_serializer,
     model_validator,
 )
 from typing_extensions import Self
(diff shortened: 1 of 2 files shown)


d-chambers:
Both taken, in `6ea85c67`. They were the same defect seen from two sides: a second `@model_serializer` on the subclass does not sit beside the base's, it replaces it.

**The class tag was being dropped.** Pydantic runs one model serializer per class, so `_keep_the_value` took `DascoreBaseModel._write_object_type`'s place and every annotation document lost the `object_type` it is dispatched by:

```python
annotation.model_dump(mode="json")
# before: {..., 'group': 'g', 'value': 2}          <- no object_type
# after:  {..., 'group': 'g', 'value': 2, 'object_type': 'OpticalPathAnnotation'}
```

The base's serializer is now extended rather than declared beside, so the tag is written first and the value goes back after it.

**A deliberately excluded value came back.** Restoring on absence alone overrode the caller's own filtering. Restoration is now limited to what `exclude_defaults` dropped, and `include`/`exclude` are honoured:

```python
annotation.model_dump(mode="json", exclude={"value"})   # no value
annotation.model_dump(mode="json", include={"group"})   # no value
```

Both are pinned by tests. `tests/test_models.py` passes with the rest: 10,180 passed, lint clean twice, doctests 154 passed.

PR #1049, opened 2026-08-26
Let istft rebuild associated coordinates

### LINKED ISSUE #1039: istft raises on any patch with an associated coordinate
## Description

`Patch.istft` raises on any patch that carries an associated coordinate. `stft` accepts the patch and keeps the coordinate, but the inverse rebuilds the coordinate manager with the non-dimensional coordinate passed bare, and `get_coord_manager` refuses it. Every patch that has been through `Spool.enrich` has associated coordinates, so on an enriched archive `stft` is a one-way trip.

## Example

```python
import dascore as dc

p = dc.get_example_patch()
p.stft(time=1.0).istft().shape        # (300, 2000)  fine

p_depth = p.update_coords(depth=("distance", p.get_array("distance") * 2.0))
p_depth.stft(time=1.0).istft()
# CoordError: Coordinates that are not named the same as dimensions must be
# passed as a tuple of the form: (dimension, coord)
```

## Expected behavior

`istft` returns a patch with the same coordinates as the one `stft` was given, associated coordinates included, as `dft(...).idft()` does for a single dimension.

## Versions
 - Python Version: 3.12.3

d-chambers:
## Description

`Patch.istft` raised on any patch carrying an associated (non-dimensional) coordinate. `_get_istft_coord` rebuilt the coordinate manager from `coords.coord_map`, which yields bare coordinates, so `get_coord_manager` rejected every coordinate whose name isn't a dimension name.

The fix uses `get_coord_tuple_map()` — the same pattern `idft` already uses in `_get_idft_coords_and_sizes` — so associated coordinates keep their dimensions. It also drops the frequency and window dimensions with `drop_coords` first, since coordinates associated with those are indexed by frequency or by window and cannot survive the inverse.

Round-trip behavior now matches `dft`/`idft`: coordinates associated with un-transformed dimensions come back unchanged, and coordinates associated with the transformed dimension are gone because `stft` already drops them (documented in its Notes). The `istft` docstring gained a Notes section saying so.

Closes #1039

## Changelog

- fixed: `Patch.istft` no longer raises on patches that carry associated (non-dimensional) coordinates.

## Checklist

I have:

I have (if applicable):

### DIFF
--- a/dascore/transform/fourier.py
+++ b/dascore/transform/fourier.py
@@ -641,15 +641,18 @@ def _get_istft_coord(coords, frequency_axis, time_axis):
     Get the coordinate manager for the inverse of the short time fourier transform.
     """
     dims = coords.dims
-    # Create new time coordinate.
-    coord_map = dict(coords.coord_map)
+    freq_name, time_name = dims[frequency_axis], dims[time_axis]
+    # Drop the transformed dims and any coords associated with them; those
+    # coords are indexed by frequency or window and cannot survive the inverse.
+    stripped = coords.drop_coords(freq_name, time_name)[0]
+    # Use the tuple map so associated coords keep their dimensions.
+    coord_map = stripped.get_coord_tuple_map()
     coord_map.pop("_stft_window")
-    time = coord_map.pop("_stft_old_coord")
-    coord_map.pop(coords.dims[frequency_axis])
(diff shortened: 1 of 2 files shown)


d-chambers:
@codex review

PR #1051, opened 2026-08-26
Let a scalar adopt a dimensionless unit instead of being scaled into it

### LINKED ISSUE #1037: patch + scalar corrupts the units of a prefixed dimensionless unit (µϵ, mrad, %) and later operations silently scale the data
## Description

Adding a bare scalar to a patch whose data units are a *prefixed dimensionless* unit — `µϵ`, `mϵ`, `µrad`, `mrad`, `percent` — leaves the data right but writes a scaled unit label on the result: `1 µϵ` becomes `428572.0 µϵ`. Nothing complains, and the label stays wrong through every later operation. Anything that then folds the unit magnitude into the data (dividing by another patch, `convert_units`, a second addition) is silently off by a factor of ~4e5. Units with a real dimension (`mm`, `km`, `µm`, `kHz`) and the unprefixed `ϵ`/`rad` are fine, so this is easy to miss in tests and easy to hit with BOTDR data, which arrives in `µϵ`.

The cause is in `dascore/utils/array.py::_apply_op_one_unitful`. The probe pair `2.0 * µϵ + 1.5 * dimensionless` is meant to raise `DimensionalityError` so that the bare side adopts the patch's units, but pint accepts it, because microstrain *is* dimensionless: it converts `1.5` to `1.5e6 µϵ`, the probe comes out as `1500002 µϵ`, the plain result is `3.5`, and `_label(probe / plain)` is `428572 µϵ`. The adopt branch has to run for every unit whose base is dimensionless, not only when pint refuses.

## Example

```python
import dascore as dc
import numpy as np

p = dc.get_example_patch().set_units("µϵ")

(p + 1).attrs.data_units      # '428572.0 µϵ'    expected '1 µϵ'
(p - 1).attrs.data_units      # '-2999996.0 µϵ'
(1 - p).attrs.data_units      # '-2.999996'
((p + 1) + 1).attrs.data_units  # '673469.714286 µϵ'
np.allclose((p + 1).data, p.data + 1)   # True: the data are right, the label is not

# and then the label bites
q = p / (p + 1)
np.nanmedian(q.data / (p.data / (p.data + 1)))   # 2.3e-06, expected 1.0
np.nanmedian((p + 1).convert_units("ϵ").data / (p.data + 1))   # 0.43, expected 1e-6

for u in ["mrad", "percent", "mm", "ϵ"]:
    print(u, (dc.get_example_patch().set_units(u) + 1).attrs.data_units)
# mrad 429.142857143 mrad
# percent 43.4285714286 %
# mm 1 mm
# ϵ 1 ϵ
```

Comparisons are unaffected (`p > 1` drops the units), and patch-with-patch arithmetic in `µϵ` is fine; it is only the one-unitful branch.

## Expected behavior

`p + 1` on a `µϵ` patch keeps `1 µϵ` as its units, as it does for `mm`. In `_apply_op_one_unitful`, treat a probe that pint accepts by coercion the same as one it refuses: when the known unit is dimensionless-compatible (its base units are dimensionless), take the adopt path for `add`, `subtract` and the comparisons directly rather than letting pint scale the bare side.

## Versions
 - Python Version: 3.12.3

d-chambers:
## Description

Closes #1037.

Adding a bare scalar to a patch whose data units were a *prefixed dimensionless* unit (`µϵ`, `mrad`, `percent`, `ppm`) left the data right but wrote a scaled unit label on the result: `dc.get_example_patch().set_units("µϵ") + 1` came out as `428572.0 µϵ` instead of `1 µϵ`, and the wrong label then rode through every later operation, so a `convert_units` or a division by another patch was silently off by ~1e6.

`_apply_op_one_unitful` decides how to read the operand without units by probing the registry: dimensionless first (right for products and quotients), and, if the registry raises `DimensionalityError`, the other side's units (right for sums, differences, and comparisons). Microstrain *is* dimensionless, so the registry never refused — it coerced `1.5` into `1.5e6 µϵ` — and the adopt branch never ran. The units of one unit of output, taken as the probe's result over the bare result, then came out as the unit's magnitude.

`docs/notes/patch_compatibility.qmd` already documents the behaviour this PR produces — "takes the other operand's units where the operation needs equal units" — so the code was violating its own published contract for these units. No doc change is needed.

### The fix

Ask the operator, rather than the unit, what it wants of its operands. `_needs_equal_units` puts the same probe to metres, where the registry does refuse a bare number, and a second one to metres beside metres: an operator which wants **one unit on both sides** refuses the first and accepts the second, and there the bare side adopts the known units. An operator which wants its operands **dimensionless outright** — `logaddexp`, an exponent in `2 ** patch` — refuses both, and is left exactly as it was, since a dimensionless unit already satisfies what it asked for.

I kept this as a probe rather than the hand-written `frozenset` that `_OFFSET_UNIT_OPERATORS` uses three lines above, even though the probe's output today is a fixed 14-ufunc set. The two answer different kinds of question: `_OFFSET_UNIT_OPERATORS` is dascore's own policy on what stays meaningful on a temperature, so writing it out states a decision, while this one restates pint's classification of its ufuncs — a fact about a dependency, which a literal set would silently get wrong the day pint reclassifies one. The probe re-derives it. The two costs a set would avoid are addressed directly: the `reversed` argument is gone (I checked all 41 distinct numpy 2-in/1-out ufuncs; the answer never depends on operand order), and the cache is `lru_cache(maxsize=64)` rather than unbounded, so a runtime-built `np.frompyfunc` cannot accumulate in it.

### Blast radius

I checked every binary ufunc against every unit class before and after. The data are identical everywhere; only labels move, and only for units whose base is dimensionless. Nothing at all changes for a unit with a dimension (`mm`, `km`, `kHz`, `µm`, `m/s`, `100 cm`), for an offset unit (`degC`), or for a patch with no units.

For the dimensionless units, `add`, `subtract`, `maximum`, `minimum`, `hypot`, `copysign`, `nextafter` and `arctan2` now agree with what `mm` does. Three cases worth calling out beyond the `µϵ`/`mrad`/`%` of the issue:

- `ϵ`, `rad` and `count` used to lose their label entirely on a reversed operand, so that `1.5 - patch` had no units; they keep it now.
- A level in `dB` is dimensionless in pint, so it is one of these units too: `maximum`, `minimum`, `hypot`, `copysign`, `nextafter` and `arctan2` against a bare number now label `1 dB` (or `1 rad`) instead of `0.79`, `None`, `0.87`. Adding a bare number to a level is still refused earlier, by the logarithmic guard.
- A unit carrying its own magnitude keeps it: `set_units("2 percent") + 1` gives `2.0 %`, exactly as `set_units("100 cm") + 1` gives `100.0 cm`.

### Not fixed here

An operator which needs its operands genuinely dimensionless still runs on the bare data, so `2 ** patch` and `np.logaddexp(patch, 1.5)` on a `µϵ` patch are off by the unit's magnitude and carry a meaningless label — the same shape of bug as #1037, through a different operator. This PR leaves that behavior bit-identical to `dev` rather than papering over it; it is filed as #1059.

## Changelog

- fixed: Adding or subtracting a bare number no longer scales the units of a patch whose data units are a prefixed dimensionless unit such as `µϵ`, `mrad`, or `percent`.

## Checklist

I have:

I have (if applicable):

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of operations involving unit-aware values.
  * Dimensionless and logarithmic units now propagate correctly during supported operations.
  * Comparisons, products, scaled units, and unsupported operations now receive more consistent unit handling.
  * Behavior is consistent across supported computation backends.

### DIFF
--- a/dascore/utils/array.py
+++ b/dascore/utils/array.py
@@ -307,6 +307,38 @@ def _is_offset_unit(quantity) -> bool:
     return np.isclose(two - one, one - zero) and not np.isclose(zero, 0)
 
 
+# Bounded rather than unbounded: a ufunc can be made at runtime
+# (np.frompyfunc) and the cache holds a reference to whatever it is given.
+@functools.lru_cache(maxsize=64)
+def _needs_equal_units(operator) -> bool:
+    """
+    Return True when an operator wants both of its operands in one unit.
+
+    Metres are the yardstick: the registry refuses a bare number beside
+    them, so what comes back is the operator's own requirement rather than
+    the unit's. An operator which wants dimensionless operands refuses
+    metres beside metres too, which is how the two are told apart. The
+    registry decides by dimensionality, so the operand order does not
+    change the answer.
+    """
+    meter, dimensionless = get_quantity("meter"), get_quantity("dimensionless")
+    assert meter is not None and dimensionless is not None
+
+    def _refuses(other):
+        # DimensionalityError is a TypeError, so these clauses may not be
+        # reordered. The second catches every other way the registry can
+        # decline (an unimplemented ufunc, a dtype, a gufunc's shapes).
+        try:
+            operator(2.0 * meter, other)
+        except DimensionalityError:
+            return True
+        except (TypeError, ValueError):
+            return False
+        return False
+
+    return _refuses(1.5 * dimensionless) and not _refuses(1.5 * meter)
+
+
 def _is_logarithmic_unit(quantity) -> bool:
     """
     Return True for a logarithmic unit (dB), whose base conversion is not linear.
@@ -439,12 +471,16 @@ def _apply_binary_ufunc(
         The side without units conflicts with nothing: it is taken as
         dimensionless first (right for products and quotients) and, if the
         operator rejects that, as sharing the other side's units (right for
-        sums, differences, and comparisons). The units of one unit of output
-        are settled on scalars — the probe's result over the bare result —
-        so the data stay bare, a scale in the units ("100 cm") rides along
-        unchanged, and nothing large is ever wrapped by the unit registry.
-        A ufunc the registry does not implement falls through to numpy with
-        the units left as they were.
+        sums, differences, and comparisons). A unit whose base is itself
+        dimensionless (µϵ) is one the registry coerces the bare side into
+        rather than rejecting, so there the operator is asked outright and
+        the bare side adopts the units wherever it wants one unit on both
+        sides. The units of one unit of output are settled on scalars —
+        the probe's result over the bare result — so the data stay bare, a
+        scale in the units ("100 cm") rides along unchanged, and nothing
+        large is ever wrapped by the unit registry. A ufunc the registry
+        does not implement falls through to numpy with the units left as
+        they were.
         """
         known = data_units if data_units is not None else other_units
         if _is_offset_unit(known):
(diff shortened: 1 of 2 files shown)


d-chambers:
@codex review

PR #1052, opened 2026-08-26
Trim a chunk only on the dimension it was chunked on

### LINKED ISSUE #1038: In-memory spools trim chunks on associated coordinates: channels lost, and chunk(distance=...) raises
## Description

A spool built from an in-memory patch that carries associated (non-dimensional) coordinates loses channels when it is chunked, and can crash. The in-memory planner appears to feed every `<coord>_min`/`<coord>_max` pair of the patch summary back into `patch.select` when it trims a chunk, associated coordinates included. A string coordinate then keeps only the channels matching one value; a numeric coordinate with `nan` in it keeps only the channels where it is not `nan`; and chunking along the dimension the coordinate shares raises `CoordError` from `UnCoord`. Plain indexing (`spool[0]`) and `select` are fine, and a file-backed spool of the same patch chunks correctly, so it is specific to `dc.spool(patch)` / `dc.spool([patches])` going through the planner.

Every patch that has been through `Spool.enrich` has such coordinates (`x`, `y`, `z`, `hole_depth` with `nan` between the labelled stretches, and string `borehole`/`leg` coordinates), which is how I hit it: `dc.spool(enriched_patch).chunk(time=1)` returned 18 of 267 channels, the stretch matching the *last* label in the inventory.

Possibly related to #813, which is about non-dim coords sharing the concatenation dimension; here the coordinates share the *other* dimension and it is the trim, not the concatenation, that goes wrong.

## Example

```python
import dascore as dc
import numpy as np

p = dc.get_example_patch()                      # (time: 300, distance: 2000)
d = p.get_array("distance")
p_str = p.update_coords(label=("distance", np.where(d > 150, "far", "near")))
p_nan = p.update_coords(xcoord=("distance", np.where(d > 150, d * 2.0, np.nan)))
p_num = p.update_coords(depth=("distance", d * 2.0))

[x.shape for x in dc.spool(p_str).chunk(time=2)]   # [(300, 500), (300, 500), ...]  expected (300, 2000)
[x.shape for x in dc.spool(p_nan).chunk(time=2)]   # [(149, 500), ...]
[x.shape for x in dc.spool(p_num).chunk(time=2)]   # [(300, 500), ...]  fine: no nan, no strings
dc.spool(p_str)[0].shape                            # (300, 2000)      fine
dc.spool(p_str).select(time=(1, 3), relative=True)[0].shape   # (300, 501)  fine

dc.spool(p_nan).chunk(distance=100)
# CoordError: UnCoord does not support relative and samples must be True for both select and order methods.

# selecting by a range on a string coordinate is refused on the patch, which is
# what the planner seems to be doing under the hood
p_str.select(label=("far", "near"))
# CoordError: String coordinates do not support range selection.
```

(The `(300, 500)` for `p_str`: 500 is the number of channels the trim happens to keep, not the length of the label.)

## Expected behavior

`dc.spool(p).chunk(time=2)` returns patches with all 2000 channels, exactly as a file-backed spool of the same patch does, and `chunk(distance=100)` does not raise. The trim after a chunk should select on the chunked dimension (and on the other dimensions only when the plan actually narrows them), never on associated coordinates.

## Versions
 - Python Version: 3.12.3

d-chambers:
## Description

Closes #1038.

A chunked spool assembles each output from *member rows*: the source row with the planned dimension's envelope replaced by the member's trim range. Every other `<name>_min`/`<name>_max` pair in that row still describes the whole source, but `PatchAssembler._load_trimmed_patch` fed all of them into `Patch.select` — and, because the converted kwargs travel on to `PlanResolver._load_member`, down to the reader as trim hints as well.

Re-selecting a coordinate to its own extent is a no-op for a sorted numeric range, and harmless for the dimensions. It is not harmless for an associated (non-dimensional) coordinate:

- a string coordinate becomes `select(label=[first, last])`, a membership test only the channels labelled with an endpoint pass;
- a numeric coordinate holding `NaN` loses the channels it says nothing about, since a missing value is outside every range;
- a coordinate that refuses range selection at all raises `CoordError` from `UnCoord`, which is why `chunk(distance=...)` crashed.

Every patch that has been through `Spool.enrich` carries such coordinates, so this hit ordinary use.

The fix gives `PatchAssembler` the plan's dimension — the only range a member row narrows — restricts the post-load trim to it, and drops the associated coordinates' ranges before the kwargs become reader trim hints. Dimensional ranges still travel as hints, so multi-patch file filtering is unchanged; patch binding goes by `source_patch_key`, never by the hints.

### Two corrections to the issue

**File-backed spools were broken too.** The issue reports them as correct; they are not. The converted `label: ['end', 'start']` kwarg reaches the DASDAE reader as a hint and `coords.select` performs the membership trim inside `_read_patch` (`dascore/io/dasdae/utils.py:339-353`), so a directory spool loses the same channels and raises the same `CoordError`. Both paths are fixed together, which the new equivalence test pins.

**The damage was wider than lost channels.** When the associated coordinate rides the *chunked* dimension, dev also produced wrong-length outputs, dropped whole output patches with `#583` "no patch remained" warnings, and left `get_contents()` describing patches the spool does not yield. A contents-vs-patch agreement check (each patch's real coord min/max against its `get_contents()` row) fails on dev in 4 of 11 spool constructions — in-memory and file-backed `chunk(distance=100)`, both `select`-then-chunk variants, and `chunk(time=2).chunk(distance=100)` — and passes on all 11 here. The fix restores an invariant rather than trading one bug for another.

### On coverage

The two remaining defensive guards — the planned dimension being on the loaded patch, and a member row naming its dimensions — are never reached by the suite, so they are written as `assert` statements on their own lines rather than folded into a covered `or` chain. This repo measures line coverage only, so folding them would have read as covered while saying nothing about them, which is what a `# pragma: no cover` would do. A third guard, `plan_dim is None`, is gone: the field is now required, since an assembler built without one trims nothing and hands every dimensional range on as a read hint. `dascore/utils/patch_assembly.py` and `dascore/io/index/planned.py` are both at 100%.

Three pre-existing problems on this seam are filed separately and untouched here: #1061 (associated coordinates disable the streaming merge), #1062 (concatenating a file-backed spool raises `KeyError`), #1063 (the unit-mismatch read-hint guard is inert in chunk mode).

## Changelog

- fixed: Chunking a spool no longer trims its patches on associated (non-dimensional) coordinates, which dropped channels, mis-sized or dropped whole output patches when the coordinate rode the chunked dimension, and could raise a `CoordError`.

## Checklist

I have:

I have (if applicable):

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved planned patch loading so only the intended dimension is trimmed.
  * Preserved associated coordinates and unrelated dimensions during patch chunking.
  * Ensured consistent behavior for in-memory and file-backed patches.

* **Tests**
  * Added coverage for string, numeric, and missing-value coordinates.
  * Verified that all channels remain available after chunking.

### DIFF
--- a/dascore/io/index/planned.py
+++ b/dascore/io/index/planned.py
@@ -523,10 +523,10 @@ class PlanResolver(PatchResolver):
         return nested
 
     def _assembler(self):
-
         return PatchAssembler(
             load_patch=self._load_member,
             merge_kwargs=self.merge_kwargs,
+            plan_dim=self.dim,
         )
 
     def _load_member(self, kwargs: Mapping) -> dc.Patch:
--- a/dascore/utils/patch_assembly.py
+++ b/dascore/utils/patch_assembly.py
@@ -102,13 +102,45 @@ def _match_merge_units(patch, merge_dim, target_units):
     return patch, target_units
 
 
-def _coord_only_kwargs(patch, kwargs) -> dict:
-    """Keep only the kwargs naming a dim or coordinate of patch."""
-    return {
-        k: v
-        for k, v in kwargs.items()
-        if k in patch.dims or k in patch.coords.coord_map
-    }
+def _drop_associated_ranges(row, kwargs, plan_dim) -> dict:
+    """
+    Drop the ranges of coordinates which merely ride a dimension.
+
+    `_convert_min_max_in_kwargs` turns every `<name>_min`/`<name>_max`
+    pair the row carries into a `<name>: [min, max]` range, and those
+    ranges travel on as read hints. Only the planned dimension's range
+    is a trim; an associated (non-dimensional) coordinate's is its whole
+    extent, and asking a reader to select on it drops the channels a
+    string coordinate labels with neither endpoint, or a numeric one
+    leaves NaN.
+    """
+    # a row which doesn't name its dimensions keeps only the plan's range,
+    # costing the reader a hint rather than risking a wrong trim
+    raw = row.get("dims")
+    dims = {x for x in raw.split(",") if x} if isinstance(raw, str) else set()
+    ranged = {x.rsplit("_", 1)[0] for x in row if x.endswith(("_min", "_max"))}
+    drop = ranged - dims - {plan_dim}
+    return {k: v for k, v in kwargs.items() if k not in drop}
+
+
+def _plan_trim_kwargs(patch, kwargs, plan_dim) -> dict:
+    """
+    Keep only the trim the plan actually narrows.
+
+    A member row is its source row with the *planned* dimension's
+    envelope replaced by the member's trim range; every other range
+    column still describes the whole source. Selecting on those would
+    re-select a coordinate to its own extent, which is a no-op for a
+    sorted numeric range but not for a string coordinate (a range of
+    labels), one holding NaN (missing values fall outside every range),
+    or one which cannot be range-selected at all.
+    """
+    # an unmodified member states no range, and a patch without the
+    # planned coordinate has nothing of that name to trim
+    coord_map = patch.coords.coord_map
+    if plan_dim is None or plan_dim not in kwargs or plan_dim not in coord_map:
+        return {}
+    return {plan_dim: kwargs[plan_dim]}
 
 
 def _as_plan_units(patch, kwargs, row) -> dict:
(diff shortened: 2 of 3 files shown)


d-chambers:
@codex review

PR #1055, opened 2026-08-26
Enrich what the acquisition and the path state, and keep the file's own values

### LINKED ISSUE #1043: Spool.enrich copies an unclear subset of acquisition attrs, overwrites under keep_first, and does not enrich coupling
## Description

`Spool.enrich` copies some of an acquisition's fields onto the patch attrs and not others, and from the outside the selection looks arbitrary. On the galileo_2026 archive it fills `interrogator.instrument_type`, `interrogator.manufacturer`, `interrogator.model`, `interrogator.name`, `gauge_length` and `closed_fiber_loop`, and in doing so overwrites the file's own `interrogator.model` (`"Peta"` from the Sintela header becomes `"ONYX peta"` from the inventory) although `conflict="keep_first"` is the default and the file's value came first. At the same time it leaves `data_category` empty on the low-frequency DAS patches whose acquisition says `data_category: DAS`, and it does not copy `sample_rate`, so a patch enriched from an acquisition can still not say which category it is.

Coupling is the other gap. Labels and geometry become per-channel coordinates (`borehole`, `leg`, `hole_depth`, `x`, `y`, `z`), which is what makes `select(borehole="N*")` possible, but the path's `coupling` conditions do not, so there is no `select(coupling_type=...)`; the "grouted versus hanging" comparison, which is the first thing anyone asks of this deployment, has to be done by hand.

## Example

```python
import dascore as dc
from galileo_2026 import get_data_path, get_inventory_path   # DASDAE/galileo_2026

inv = dc.inventory(get_inventory_path())
spool = dc.spool(get_data_path()).update().attach_inventory(inv)

raw = spool.select(tag="DAS")[0]
rich = spool.enrich().select(tag="DAS")[0]
raw.attrs.get("interrogator.model"), rich.attrs.get("interrogator.model")
# ('Peta', 'ONYX peta')            overwritten under conflict="keep_first"

lf = spool.enrich().select(tag="DAS_LF")[0]
lf.attrs.data_category, lf.attrs.get("sample_rate")
# ('', None)                       the acquisition says data_category: DAS, sample_rate: 5.0

lf.select(coupling_type="outside_borehole_casing")
# PatchCoordinateError: Coordinate(s) ['coupling_type'] not found in patch coordinates: [...]
```

## Expected behavior

- A documented rule for which acquisition fields become attrs (all of the acquisition's scalar fields seems the least surprising), applied consistently, so `data_category` and `sample_rate` arrive with the rest.
- `conflict="keep_first"` keeps the file's value and `conflict` governs the interrogator fields the way it governs everything else; if the inventory is meant to win, say so in the docstring.
- Coupling conditions enriched as a per-channel string coordinate (`coupling_type`, or one coordinate per `CouplingCondition` field), alongside labels and geometry.

## Versions
 - Python Version: 3.12.3

d-chambers:
## Description

Closes #1043.

Three things about `enrich` that the galileo_2026 archive made obvious: it copied a subset of the acquisition's fields whose membership looked arbitrary, its default `conflict` overwrote the file's own header, and the one per-channel fact anybody wants to compare that deployment by — how the fiber is coupled — was not projected at all.

**One rule on each side.** Blanket enrichment now copies every scalar field the resolved acquisition (and its interrogator) states, less four names: `data_type` and `data_units`, which describe the data as it now stands rather than the system that recorded it, and `sample_rate` and `spatial_interval`, which the patch's own coordinates already state. `data_category` was in the excluded set and does not belong there — no processing function rewrites which family of instrument recorded the data, so DAS data stays DAS data — which is why the low-frequency DAS patches came out with an empty `data_category` while their acquisition said `DAS`. It is now copied with the rest.

I did not move `sample_rate` and `spatial_interval` with it, though the issue asks for `sample_rate` too. The patch's time coordinate *is* its sample rate, and a decimated patch enriched with the as-acquired value would carry an attr contradicting its own coordinate. Naming it (`attrs=("sample_rate",)`) still asks for it, which is the escape hatch for wanting the as-acquired value on purpose.

**`conflict` means what it means everywhere else.** `keep_first` kept the *inventory's* value, which is the opposite of what the word means in every other dascore verb and is why the Sintela header's `interrogator.model` of `"Peta"` became the inventory's `"ONYX peta"` under a policy named for keeping the first value. `keep_first` now keeps what the patch stated — a reader read it out of the file header, so it was there first — and a new `keep_last` is the inventory correcting the file. Filling an attr the patch leaves unset is not a disagreement and still happens under every policy, which is most of what enrichment does.

This also makes the default cheaper at the spool level: pending enrichment with `keep_first` no longer rewrites any stated header, so `Spool.select` no longer resolves stated rows against the inventory to judge them.

**Coupling is projected.** A blanket coord request now adds `coupling`, whose value per channel is the coupling type — `spool.enrich().select(coupling="grouted")` is the grouted-versus-hanging comparison. It was already reachable as `coords=("coupling",)`; it was only missing from the blanket set, whose comment claimed the typed tracks restate what the patch's axis and the inventory already record. That is true of the geometry segments and the optical components, which is why those two stay opt-in, but a coupling condition states something about the fiber's surroundings that nothing else does. The projection machinery is unchanged; a path stating no coupling conditions still projects none.

One boundary worth naming: a patch which already carries a coordinate called `coupling` and disagrees with the inventory about it now raises under blanket enrichment, where before the name was not asked for. That is the rule every label group and geometry column already lives under — the error says to rename or drop the coordinate, and naming the coordinates wanted (`coords=("x", "y", "z")`) is the way past it — and `coupling` is a reserved track name in the inventory vocabulary, so a patch coordinate of that name genuinely is ambiguous. Both the collision and the re-enrich refresh are pinned by tests.

The tutorial now states the copying rule and the conflict policies, with the coupling selection as the example.

## Changelog

- fixed: `Patch.enrich` and `Spool.enrich` with `conflict="keep_first"` (the default) now keep the patch's own value of an attr the inventory disagrees with, rather than the inventory's.
- added: `enrich` accepts `conflict="keep_last"`, under which the inventory's value corrects the patch's.
- changed: Blanket enrichment copies `data_category`, which describes the acquisition rather than the state of the data.
- added: Blanket enrichment projects the optical path's coupling conditions as a `coupling` coordinate.

## Checklist

I have:

I have (if applicable):


d-chambers:
@coderabbitai review

d-chambers:
@codex review

PR #1024, opened 2026-08-26
Benchmark: the section level API sidebar on a full build

d-chambers:
## Description

Not for merge. This is the paired run for #1022: the same commit, on the same runner class, with `DASCORE_DOC_COMPACT_SIDEBAR` forced on so the doc build renders every page with a 16 entry sidebar instead of one naming all 1,388 documented objects.

On one API page rendered twice each way, the sidebar was worth 503,534 -> 35,592 bytes of html and 20.2 -> 8.2 s. This says what it is worth across the whole site: compare the build report artifact here against the one on #1022.

Close this once the two reports have been compared.

## Changelog

- none

PR #1067, opened 2026-08-26
List the sixteen API sections, not every object

d-chambers:
## Description

The API sidebar named all 1,388 documented objects. That put 165 KiB in the generated `_quarto.yml` and repeated it into every page rendered from it, and it cost more than everything else the build did.

Two full CI builds differing only in this (32960776336 and 32960821408, on #1024):

| | every object | sixteen sections | |
|---|---:|---:|---:|
| quarto wall | 110.1 min | 34.1 min | −69% |
| page phase | 105.0 min | 33.8 min | −68% |
| notebook preparation | 7.96 min | 0.67 min | −92% |
| all timed phases | 118.3 min | 35.0 min | −70% |
| rendered site | 802.2 MiB | 208.5 MiB | −74% |
| API html, total | 635.0 MiB | 41.5 MiB | −93% |
| API page, p50 | 489.4 KiB | 31.0 KiB | −94% |
| generated `_quarto.yml` | 175,216 B | 7,268 B | −96% |

Per page, static pages fall from 3.76 s to 0.68 s. The cost was fixed overhead per page, not content, which is why even `build_notebooks.py` got 12x faster: each tutorial render parses the same config.

That meets both goals the API surface plan set — a cold build under 60 minutes, and at least a 75% cut in API html — without touching how pages are generated.

## What a reader loses, and does not

The sidebar no longer lists every object, so a reader reaches one from the page of whatever owns it. That already works: every module page lists the modules, classes, and functions it holds, and every class page lists its methods. Search is unchanged, since quarto indexes rendered pages rather than sidebar entries.

## Notes

- `docs/api/_metadata.yml` says which sidebar the unlisted pages use. Without it quarto gives an unlisted page the first sidebar in the config, which is the one for the front matter. Verified on a rendered `utils/time/to_datetime64.html`: all sixteen sections in its nav, 41 KB against 503 KB before. It needs a `.gitignore` negation, since `docs/api/*` is generated and untracked.
- The exhaustive builder and the `DASCORE_DOC_COMPACT_SIDEBAR` switch are both gone; the experiment they existed for is the table above.
- Stacked on #1022 for the measurement harness. Rebase to dev once that merges.

## Changelog

- changed: The API documentation sidebar lists the sixteen top level sections rather than every documented object, which cuts the documentation build from about two hours to about half an hour.

### DIFF
--- a/.github/workflows/test_doc_build.yml
+++ b/.github/workflows/test_doc_build.yml
@@ -6,14 +6,8 @@ on:
   pull_request:
     types: [labeled, synchronize]
 
-  # A manual run is how the sidebar experiment is measured: the same build,
-  # on the same runner class, with only the navigation changed.
+  # A manual run is how a full build is measured without waiting for a label.
   workflow_dispatch:
-    inputs:
-      compact_sidebar:
-        description: "Build the API sidebar with one entry per section"
-        type: boolean
-        default: false
 
 permissions:
   contents: read
@@ -31,8 +25,6 @@ env:
   # point at the released site, where a page added since the last release 404s
   # and the preview misleads the reviewer.
   DASCORE_DOC_SITE_URL: https://dascore.netlify.app
-  # Empty unless a manual run asked for the section level sidebar.
-  DASCORE_DOC_COMPACT_SIDEBAR: ${{ inputs.compact_sidebar && '1' || '' }}
 
 jobs:
   test_build_docs:
--- a/.gitignore
+++ b/.gitignore
@@ -71,6 +71,8 @@ target/
 
 # docs
 docs/api/*
+# The generated pages are not tracked, but the sidebar they use is.
+!docs/api/_metadata.yml
 # Where a local `quarto render docs` puts the built site.
 docs/_site/
 _autosummary
--- a/docs/api/_metadata.yml
+++ b/docs/api/_metadata.yml
@@ -1,0 +1,4 @@
+# Every page under api/ uses the API sidebar. Only the 16 section pages
+# are listed in it, and quarto gives an unlisted page the first sidebar
+# it finds, which would be the one for the front matter.
+sidebar: API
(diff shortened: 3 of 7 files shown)


d-chambers:
## Acceptance build

Run 33005725355, 37.6 min wall.

| | before | this branch |
|---|---:|---:|
| all timed phases | 118.3 min | 36.3 min |
| quarto wall | 110.1 min | 35.3 min |
| page phase | 105.0 min | 34.9 min |
| notebook preparation | 7.96 min | 0.69 min |
| rendered site | 802.2 MiB | 168.2 MiB |
| API html total | 635.0 MiB | 48.6 MiB |
| API page p50 | 489.4 KiB | 36.5 KiB |
| uploaded artifact | 139.5 MiB | 105.1 MiB |
| generated API sidebar | 164.3 KiB | 1.1 KiB, 16 entries |

The p50 is 36.5 KiB rather than the 31.0 KiB the benchmark run showed, and that difference is the point of `docs/api/_metadata.yml`: the benchmark's unlisted pages were falling back to the front matter sidebar, and these carry the API one as they should.

URL baseline against the refrozen manifest in #1068: **0 added, 0 removed, 0 moved, 0 unpublished**. No page moved and no link broke.

**Search is unaffected**, which was the one reader-facing claim I could not check without a full render. `search.json` holds 2,582 entries, 2,338 of them API pages — quarto indexes what it renders, not what the sidebar lists. Spot checks resolve: `to_datetime64` -> `api/dascore/utils/time/to_datetime64.html`, `waterfall` -> `api/dascore/viz/waterfall/waterfall.html`, and the `Patch` and `Spool` class pages are both indexed.

The remaining page phase is executable pages, not page count: 161 source-only pages cost 15.5 min against 13.8 min for all 1,175 static ones. That is the next lever, not aggregation.

PR #1076, opened 2026-08-27
Stop doing timedelta arithmetic numpy has deprecated

### LINKED ISSUE #1074: Merging datetime patches performs deprecated scalar arithmetic on NumPy timedeltas
## Description

Merging ordinary contiguous patches with a datetime time axis reaches a NumPy `DeprecationWarning` for each input member when warnings are configured to show every occurrence. Library-originated `DeprecationWarning`s are normally hidden by Python's default filter, so users may see nothing until a future NumPy release promotes this path to an error. NumPy states that the generic timedelta conversion will become an error in the future, so a currently successful DASCore merge is on a path to failure.

## Example

```python
import warnings

import dascore as dc

patch = dc.get_example_patch("random_das")
parts = [
    patch.select(time=(None, 50), samples=True),
    patch.select(time=(50, None), samples=True),
]

with warnings.catch_warnings(record=True) as caught:
    warnings.simplefilter("always")
    merged = dc.spool(parts).chunk(time=None)[0]

print(merged.shape)
print([(item.category.__name__, str(item.message)) for item in caught])
```

Actual output begins:

```text
(300, 2000)
[
    (
        'DeprecationWarning',
        "The 'generic' unit for NumPy timedelta is deprecated, and will raise an error in the future. This includes implicit conversion of bare integers (e.g. `+ 1`).Please use a specific unit instead."
    ),
    (
        'DeprecationWarning',
        "The 'generic' unit for NumPy timedelta is deprecated, and will raise an error in the future. This includes implicit conversion of bare integers (e.g. `+ 1`).Please use a specific unit instead."
    ),
]
```

The merge succeeds today and emits two warnings for the two input members.

## Expected behavior

A normal merge of contiguous datetime patches should be warning-free and remain compatible with future NumPy releases.

## Diagnosis

`dascore/utils/patch.py::_middle_step` calls `convert_units` for each `np.timedelta64` coordinate step at lines 585–586. `dascore/units.py::convert_units` then evaluates `(data * mult1 + add) * mult2` at line 271. Adding the scalar affine offset, even when it is zero for a seconds-to-seconds conversion, invokes NumPy's deprecated generic timedelta conversion.

This is not specific to a file format or to non-contiguous data. It occurs in the in-memory example above while DASCore determines the merged coordinate step.

## Suggested fix

Return the input unchanged for a no-op unit conversion, or add a timedelta-aware conversion path whose multiplier and offset use explicit timedelta units. The generic affine scalar path should not receive `np.timedelta64` values.

## Suggested regression coverage

Run the reproduction inside `warnings.catch_warnings()` with `simplefilter("error", DeprecationWarning)` and assert the merged patch equals the original patch. A second case that genuinely converts compatible time units would exercise the non-no-op timedelta path.

## Versions

- OS: Ubuntu 24.04.4 LTS, Linux 6.8.0-138-generic, x86_64
- DASCore version: `0.1.22.dev307+gef094ba8d`
- DASCore commit: `ef094ba8d047422d388d719c080dc94f90cefb5e` on `origin/dev`
- Python version: `3.12.3`
- NumPy version: `2.5.2`

d-chambers:
## Description

Closes #1074.

Merging contiguous datetime patches ran arithmetic NumPy has deprecated for `np.timedelta64`. NumPy says the implicit generic-unit cast of a bare number will become an error, so operations that succeed today are on a path to failure.

The reported site is one of three, all reached by ordinary use:

- `dascore/units.py::convert_units` applied its affine factors as `(data * mult1 + add) * mult2`. `_middle_step` expresses each member's step in the merged coordinate's units, and a time-like coordinate may only carry seconds, so the conversion is always an identity — but adding the bare zero to a `timedelta64` is the deprecated cast, and it fires once per input patch. Fixed by returning the data untouched when the units already match, using the `units_match` helper written for exactly this.

  Skipping only the zero offset would have silenced the warning while leaving a second, quieter bug in place: `mult2` is a float `1.0`, so the identity conversion rounds a `timedelta64`'s integer payload through float64. `convert_units(np.timedelta64(2**53 + 1, "ns"), "s", "s")` came back one nanosecond short on every NumPy version. The fast path is exact.

- `dascore/utils/chunk_plan.py::_check_tolerance_value` tested `value < 0`, comparing a `timedelta64` against a bare integer. Every timedelta tolerance passes through it, so `spool.chunk(time=None, tolerance=to_timedelta64(10))` — the spelling added in #1065 — warns too.

- `dascore/utils/patch.py::_check_not_negative` did the same for a rolling window or step, so `patch.rolling(time=<timedelta>, step=<timedelta>)` warns.

Both comparisons now use the `to_timedelta64(0) if is_timedelta64(value) else 0` idiom `chunk_plan.py` already uses for the chunk value.

### Behavior notes

A conversion to the units already held now returns the input object rather than a float copy, so `convert_units(1, "m", "m")` is `1`, not `1.0`. Conversions that actually convert are untouched, offsets included.

One degenerate input changes: a `Quantity` whose *magnitude* is `pd.NA`, passed as units, previously propagated `<NA>` and now raises `TypeError` from the magnitude comparison. `get_quantity(pd.NA)` already raises, so such a value can only arrive by constructing the `Quantity` by hand; nothing in the repo does.

### Verification

The deprecation exists only on NumPy >= 2.4, so I built a second environment on NumPy 2.5.2 (the reporter's version) to check against. The issue's reproduction is warning-free there after the change, and I confirmed each of the three sites warned before it and does not after. The full suite passes on both 2.3.5 and 2.5.2, and coverage stays at 100% on all three changed files.

I found the second and third sites by sweeping the whole suite under 2.5.2 for this warning, not from the issue.

No tests added — these are deprecation fixes, and the existing suite already covers the changed lines.

### Follow-up, not in this PR

That sweep also found about a dozen test-only spellings that trip the same deprecation — `np.timedelta64("NaT")` with no unit, and `time_array + 10` on a `datetime64` array — across `test_display`, `test_file_spool`, `test_directory_spool`, `test_spool_inventory`, `test_dasdae`, `test_prodml_write` and `test_patch_chunk`. None is a library defect, but they will fail once NumPy promotes the deprecation, and they want one mechanical sweep rather than a partial fix here.

## Changelog

- fixed: Merging patches with a datetime coordinate, chunking with a timedelta tolerance, and rolling over a timedelta window no longer perform arithmetic NumPy has deprecated for `timedelta64`.
- fixed: `dascore.units.convert_units` no longer rounds a `timedelta64` through floating point when the units already match.

## Checklist

I have:

I have (if applicable):

- [ ] included tests. See [testing guidelines](https://dascore.org/contributing/testing.html).

## Summary by CodeRabbit

- **Bug Fixes**
  - Unit conversions now preserve data unchanged when the source and destination units are identical.
  - Improved validation for time-based tolerances, steps, and overlaps.
  - Prevented deprecated NumPy time comparison behavior and related warnings.

### DIFF
--- a/dascore/units.py
+++ b/dascore/units.py
@@ -243,8 +243,8 @@ def convert_units(
         follows the input rather than matching it (an int in yields a float
         out) and callers legitimately pass values the checker only knows as
         `object`. Note that data is returned untouched whenever from_units
-        is None, so None survives that path but raises a TypeError once
-        there are real factors to apply.
+        is None or already names to_units, so None survives those paths but
+        raises a TypeError once there are real factors to apply.
     to_units
         The desired units after the conversion
     from_units
@@ -264,6 +264,9 @@ def convert_units(
     elif to_units is None:
         msg = "Cannot convert units to_units are not specified"
         raise UnitError(msg)
+    # Factors of one aren't free: they round a timedelta64 through float.
+    if units_match(from_units, to_units):
+        return data
     try:
         mult1, add, mult2 = _get_conversion_factors(from_units, to_units)
     except DimensionalityError as e:
--- a/dascore/utils/chunk_plan.py
+++ b/dascore/utils/chunk_plan.py
@@ -304,7 +304,9 @@ def _check_tolerance_value(value, name, shown=None, *, allow_infinite=False):
         raise ParameterError(msg)
     try:
         null = bool(pd.isnull(value))
-        negative = not null and value < 0
+        # A bare 0 would make numpy cast the timedelta to a generic unit.
+        zero = to_timedelta64(0) if is_timedelta64(value) else 0
+        negative = not null and value < zero
     except TypeError:
         msg = (
             f"The tolerance for {name!r} must be a sample count, a quantity, "
--- a/dascore/utils/patch.py
+++ b/dascore/utils/patch.py
@@ -64,7 +64,7 @@ from dascore.utils.misc import (
     yield_sub_sequences,
 )
 from dascore.utils.paths import is_memory_uri
-from dascore.utils.time import is_timedelta64, to_float
+from dascore.utils.time import is_timedelta64, to_float, to_timedelta64
 from dascore.workflow.builtin import Concatenate, Stack
 from dascore.workflow.checks import attr_type, check_patch_attrs, check_patch_coords
 from dascore.workflow.identity import (
@@ -1077,7 +1077,9 @@ def get_window_axis_step(
     def _check_not_negative(val, name):
         """Ensure a step or overlap value isn't negative."""
         magnitude = val.magnitude if isinstance(val, dc.units.Quantity) else val
-        if magnitude is not None and magnitude < 0:
+        # A bare 0 would make numpy cast the timedelta to a generic unit.
+        zero = to_timedelta64(0) if is_timedelta64(magnitude) else 0
+        if magnitude is not None and magnitude < zero:
             msg = f"{name} must be non-negative"
             raise ParameterError(msg)
 


d-chambers (commented):
The tests are not needed and the comments are too long.

d-chambers:
Done — tests dropped and the comments cut to one line each. The diff is now just the three library edits; coverage stays at 100% on all three files without them.

PR #1105, opened 2026-09-02
Delete the dev-only code nothing calls

d-chambers:
## Description

An architecture review of `dev` looked for definitions with no production caller. This removes the subset that **never shipped**: every name below was added on `dev` after v0.1.21, so no code written against a release can import it. Names the last release did publish are left in place, dead or not — those need a deprecation cycle, not a sweep.

Each name was verified with a word-boundary grep across `dascore/`, `tests/`, `docs/**.qmd`, `scripts/`, and the entry points in `pyproject.toml`, then checked against `origin/master` to confirm it is absent from the last release. The only references were the definitions and their own tests.

**Removed**

- `dascore/utils/time.py`: `_saturate_time`, `saturate_add`, `saturate_subtract`
- `dascore/utils/patch.py`: `_spool_up`
- `dascore/core/summary.py`: `PatchSummary.dim_tuple`, which returned `self.dims`
- `dascore/utils/hdf5.py`: `LocalH5Reader` and `_RemoteH5Writer._abort`, an alias for `abort()`
- `dascore/config.py`: `config_attr` and `_ConfigDescriptor`
- `dascore/io/index/catalog.py`: `PatchCatalog.add`, `remove`, `_require_root`, `attr_names`, `coord_names`, `sources`, `get_metadata`

**Branches that cannot be taken**

- `Spool.update` looked up an `index` method on the file's formatter. No `FiberIO` defines one and it was never in `FILE_FORMATTER_METHODS`.
- `build_coord_clause` carried a `None` kind key and a guard for it, but `_range_bounds` raises before it can return that kind.
- `DBDirectoryIndexer.ext` was annotated and read but never assigned.

**One small unification.** `DBDirectoryIndexer.index_map_dir` becomes a property. The `config_attr` descriptor existed to serve this one field on this one class.

Tests for the deleted names go with them. Two that read a deleted name while covering surviving behavior were rewritten against it: the scan/patch summary comparison reads `dims`, and writer abort idempotence runs against the public `abort()`. The API URL baseline drops the 46 rows for pages that no longer exist.

### Reviewer notes

Two removals are visible to out-of-tree code even though nothing in this repo reaches them, and both are deliberate:

- A third-party `FiberIO` could have defined `index(path)` and had it called by `Spool.update`. The hook was undocumented and absent from the formal method list.
- `DBDirectoryIndexer.ext` could have been assigned from outside to filter the directory walk. Also undocumented, with no in-tree setter.

Both were added on `dev`, so neither is reachable from a released version.

Not in this PR, and worth separate decisions:

- The review also found roughly 20 dead names that **did** ship in v0.1.21 (eight constants, three exceptions, `patches_to_df`, `merge_compatible_coords_attrs`, `fill_defaults_from_pydantic`, `list_ser_to_str`, `format_dtypes`, `TextWriter`, `assert_dtype_compatible_with_units`, `get_stencil_coefs`, `apply_operator`, `CoordManager.drop_disassociated_coords`, `PatchAttrs.drop_private`, and the `DTypeLike` / `CommaSeparatedStr` aliases). Removing them is a breaking change and belongs in a deprecation PR.
- `sources.last_indexed_ns` is written and never read, but dropping the column needs an `INDEX_VERSION` bump and an index rebuild.

## Changelog

- removed: Several functions and methods added since the last release but never called are gone: `saturate_add`, `saturate_subtract`, `PatchSummary.dim_tuple`, `LocalH5Reader`, `config_attr`, and the `PatchCatalog` mutation and introspection methods.
- removed: `Spool.update` no longer looks up an `index` method on a file's formatter, and `DBDirectoryIndexer.ext` no longer filters the directory walk. Neither hook was documented and no `FiberIO` in DASCore defined one.

## Checklist

I have:

I have (if applicable):

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed legacy configuration, patch utility, time arithmetic, and local HDF5 reader APIs.
  * Catalogs no longer support direct patch additions or removals; use directory synchronization instead.
  * Removed redundant patch-dimension access and deprecated HDF5 writer teardown alias.

* **Bug Fixes**
  * Improved catalog refresh behavior for single-file updates.
  * Filesystem indexing now considers files without restricting by extension.
  * Simplified coordinate query handling for supported value kinds.


d-chambers:
Closing this. The sweep grew past what a single PR should decide, and the split between dead-on-dev and dead-since-release deserves a plan rather than a judgment call per name. The findings are recorded; a narrower PR can follow later.

PR #1110, opened 2026-09-02
Slice the DASDAE dataset directly in read_array

d-chambers:
## Description

First `read_array` override, on the contract #1107 published. DASDAE stores each patch as one HDF5 dataset, so the override resolves the waveform group the key names, reads its `_dims` attribute, and slices the dataset. No attrs, coords, or Patch. The base default stays the oracle: a common-IO test checks every overriding format against it.

Median of alternated repeats. `read(select)` is `read` with coordinate kwargs, which already slices in-file; the override saves the per-call attrs/coords/Patch cost, about 4 ms.

| file | window | read + trim | read(select) | override |
| --- | --- | --- | --- | --- |
| example file (601 x 1001, 4.8 MB) | 601 x 50 | 4.6 ms | 4.7 ms | 0.59 ms |
| 2000 x 20000 float64 (320 MB) | 2000 x 1000 (16 MB) | 54 ms | 15 ms | 13 ms |
| 2000 x 20000 float64 (320 MB) | 100 x 100 | 73 ms | 4.7 ms | 0.95 ms |

`get_supported_io_table` gains a `read_array` column. Two existing tests assumed DASDAE inherited the default; the core tests now call the base body explicitly, and the resolver fixtures restore the class's own method instead of deleting it.

## Changelog

- added: DASDAE `read_array` slices the HDF5 dataset directly, so windowed reads skip attribute and coordinate parsing.
- added: `FiberIO.get_supported_io_table` reports which formats implement `read_array`.

## Checklist

I have:

I have (if applicable):


dascore/io/dasdae/utils.py
```diff
@@ -291,6 +293,58 @@ def _get_dims(patch_group):
...
+    """
+    key = normalize_source_patch_key(source_patch_key)
+    waveforms = h5.get("waveforms", {})
+    if key:
+        # h5py reads a key as a path, so "../data" or "/waveforms" would
+        # resolve to something; only a direct child by that name is a patch
+        group = waveforms.get(key) if key in waveforms else None
+        if group is None or group.name != f"{waveforms.name}/{key}":
+            msg = f"No patch named '{key}' in {h5.filename}."
+            raise PatchAttributeError(msg)
```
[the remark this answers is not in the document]
d-chambers: Agreed. b2eb2152 checks for an empty waveform group before resolving the key, so a keyed read of an empty file raises MissingPatchError as the default does; `test_keyed_empty_file_raises_missing` pins it.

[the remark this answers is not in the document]
d-chambers: Confirmed: `<name>/data` resolved under `waveforms` with a name equal to the interpolated string, so the check passed. A group name never holds a separator, so b2eb2152 refuses any key containing one before the membership and name checks; `test_path_like_key_raises` now includes `<name>/data` and `./<name>`.

dascore/io/dasdae/core.py
```diff
@@ -139,6 +144,39 @@ def read(self, resource: H5Reader, source_patch_key=(), **kwargs) -> dc.Spool:
...
+        `FiberIO.read_array` for the window contract. ``source_patch_key``
+        is the waveform group name `scan` reports; a positional index is
+        not accepted, because DASDAE never synthesizes one.
+        """
+        if kwargs:
+            msg = (
+                "read_array takes only windows and source_patch_key; "
+                f"got {sorted(kwargs)}."
+            )
+            raise ParameterError(msg)
```
d-chambers: Make a helper function for this, and make sure it is used everywhere kwargs are defined.

dascore/io/dasdae/utils.py
```diff
@@ -291,6 +293,59 @@ def _get_dims(patch_group):
...
+        return group
+    if len(waveforms) > 1:
+        msg = f"{h5.filename} holds several patches; pass source_patch_key."
+        raise PatchAttributeError(msg)
+    return next(iter(waveforms.values()))
+
+
+def _window_slice(dim, window) -> slice:
```
d-chambers: We already have a helper function very similar to this in utils, please re-use.

dascore/io/dasdae/core.py
d-chambers: Done in 9825548f: `raise_on_extra_kwargs(kwargs, accepted)` in `dascore/utils/misc.py`, with its own tests, and `read_array` calls it. There was no existing helper for this and this override is the only place in the tree that refuses leftover kwargs today (`read` uses them as filters, `scan` and `write` must tolerate the machinery's `timestamp`/`_pre_cast`), so it is the only call site for now; the Terra15 and TDMS overrides will use it. If you meant other sites, say which and I will wire them.

dascore/io/dasdae/utils.py
d-chambers: Done in 9825548f: `_window_slice` is gone; the override now runs each window through `_validate_sample_values` and `_to_slice` from `dascore/utils/misc.py`, the same pair `select(samples=True)` uses, so a float bound gets the same ParameterError the default raises and `None`/`...` open an end the same way.

PR #1161, opened 2026-09-11
Apply OptoDAS dataScale when reading

andreas-wuestefeld:
## Description

The OptoDAS reader previously returned stored sample values without applying header[dataScale], so amplitudes could disagree with the declared physical units.

Apply the scale after time/distance selection for both integer and floating-point data. Preserve floating-point precision through dtype promotion, and leave values unchanged when dataScale is absent. This change decodes storage scaling only; phase-to-strain conversion is unchanged.

Validation: 38 OptoDAS tests pass, covering versions 8–11, storage dtypes, missing/unit/non-unit scales, and subset reads. Verified against an Oslofjord recording; all pre-commit hooks pass.

## Checklist

I have (if applicable):

## Summary by CodeRabbit

* **Bug Fixes**
  * OptoDAS data now correctly applies the header-provided scaling factor to both integer and floating-point samples.
  * Scaling is applied consistently when reading selected subsets of data, while preserving coordinate alignment and units.
  * Data remains unchanged when no scaling factor is provided.

* **Documentation**
  * Clarified that reading performs storage scaling only; phase-to-strain conversion is not included.

### DIFF
--- a/dascore/io/optodas/core.py
+++ b/dascore/io/optodas/core.py
@@ -60,7 +60,12 @@ class OptoDASV8(FiberIO):
         distance: tuple[float | None, float | None] | None = None,
         **kwargs,
     ) -> dc.BaseSpool:
-        """Read a OptoDAS spool of patches."""
+        """Read an OptoDAS spool with values in the declared physical units.
+
+        The header's ``dataScale`` is applied to the selected samples for both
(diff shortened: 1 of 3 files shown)


d-chambers (approved):
LGTM

PR #1167, opened 2026-09-11
Persist coordinate variants and patch counts in the spool index

d-chambers:
## Description

First of a short series making spool index costs follow the requested data rather than the archive size.

Before this PR, every coordinate discovery (`coord_names`, `coord_meta`, and so every coordinate selection's query setup) ran a `DISTINCT` over every whole-coordinate link, and a root spool's `len` counted all patches through the query joins. Projecting a few rows also counted the patches table to choose a fetch strategy. All of these grew with the archive.

Index version 18 adds two maintained summaries and reads them instead:

- a `coord_variants` table, one row per distinct coordinate name, dtype, kind, units, and relativity, with the number of patches stating it. Its primary key is those five as a JSON array, so a tuple with nullable members has a unique identity.
- `meta_data.patch_count`.

SQLite triggers on `patches` and `patch_coords` keep both, in the transaction of the write that changes them. Because the triggers live in the database, every write is counted however it arrives: bulk inserts, replacements, cascaded deletes, and rows written straight through SQL by any connection. Counting inside the backend's own write methods, the first design here, covered only writes that went through them. Validation refuses an index missing a trigger.

**Index versions.** An index of an older version is rebuilt by the directory indexer, as before; the index is a cache whose truth is the files, and version 17 is unreleased, so there is no migration path. An index written by a *newer* DASCore now raises `InvalidIndexError` and is left byte-for-byte intact, rather than being deleted. Released DASCore versions will still delete a version 18 index they open, so directories shared across versions should give each version its own `index_path`.

Measured at 50,000 one-patch sources, three alternated runs against `dev` (each column within 3% across runs), metadata only:

| Operation | dev | this PR |
|---|---|---|
| `coord_meta()` | 60 ms | 0.5 ms |
| Root `len` | 10.6 ms | 0.005 ms |
| Projecting one row | 5.7 ms | 5.7 ms |
| Writing 50,000 sources | 1.6 s | 2.2 s |
| Replacing 1,000 sources | 66 ms | 85 ms |

The triggers cost about 11 µs per patch written, roughly a third of the database write. A real update opens every changed source file, which costs milliseconds each, so this is a small share of an update; it buys counts no connection can bypass.

**Mixed-kind sorting.** Sorting by a non-time, non-distance coordinate used one of its value kinds only: whichever SQLite happened to scan first. For a coordinate stated as numbers in some patches and text in others, the other kind's rows then tied and kept ingestion order, and a filtered view could lose its sort entirely. The order clause now has one key per kind, so each kind's values sort among themselves in any view: numbers, then text, then times. `coord_meta` rows are ordered to make that kind order fixed.

**Not included.** `coord_dims_map` and `associated_coord_names` still scan links; they serve chunk planning and need a deliberate tie rule first.

**Verification.** The work-bound tests count SQLite VM instructions at two archive sizes eight times apart, and fail against `dev` for all four discovery and count paths. A `count(*)` is a single VM instruction, so the one-row projection is checked by its executed SQL instead. The maintenance tests compare the summaries with a recount after insert, replace, delete, move, rollback, reopen, and a raw-SQL write on a heterogeneous random population, and were mutation-checked.

## Changelog

- changed: Spool coordinate discovery and a root spool's `len` read counts the index maintains instead of scanning every patch.
- changed: Opening a spool index written by a newer DASCore raises `InvalidIndexError` and leaves the file intact instead of deleting and rebuilding it.
- fixed: Sorting a spool by a coordinate stated as numbers in some patches and text in others orders each kind's values among themselves, in every view.

## Checklist

I have:

- [x] filled in the Changelog section above (see `dascore/docs/contributing/general_guidelines.qmd`).

I have (if applicable):


d-chambers:
@coderabbitai review

PR #1194, opened 2026-09-21
Support explicit window arrays in spool selection and chunking

d-chambers:
## Description

Allow callers to pass an `(n, 2)` array of absolute coordinate bounds to `Spool.select`, `Spool.chunk`, and `Spool.chunk_plan`. Selection returns the source pieces for each requested interval; chunking assembles each interval under the existing compatibility, tolerance, and filling rules. Overlapping and repeated intervals retain their input order.

`on_incomplete="raise"` makes unmet chunk requests explicit. `"warn"` and `"ignore"` skip them, while the existing `keep_partial=True` permits available partial chunks. Sample-index and relative array selections are rejected for now.

Validation covers the full test suite (including generated documentation examples), coverage, doctests, two clean full lint runs, and documentation rendering. The MiniSEED fixture also uses the writer API’s `start_time` argument so its mixed-dtype check runs correctly.

## Changelog

- added: Explicit coordinate-window arrays for spool selection and chunking, with `on_incomplete` handling and support for existing partial-chunk, gap-tolerance, and fill options.

## Checklist

I have:

- [x] filled in the Changelog section above (see `dascore/docs/contributing/general_guidelines.qmd`).

I have (if applicable):


d-chambers:
Addressed the [partition lookup review](https://github.com/DASDAE/dascore/pull/1194#pullrequestreview-5270817109) in [4aced569](https://github.com/DASDAE/dascore/commit/4aced569ab6e55b905bfae24cf8c5fd1b0cc204c): each partition now collects its available source coordinates once. Bounds calculation and coordinate selection still run independently for each requested window, preserving missing-coordinate filtering and output behavior.

For the [docstring warning](https://github.com/DASDAE/dascore/pull/1194#issuecomment-5766050932), the new catalog helpers now explain revision sharing, coherent snapshot caching, and deferred operations. I left trivial forwarding methods, nested test helpers, and unrelated existing functions unchanged.

The same commit adds the existing `concurrency` category to the thread snapshot test. Pyodide has no OS threads, and the WASM workflow already excludes that category. The test body and every assertion are unchanged: native selection with `-m concurrency` passes, and the other 76 explicit-range tests pass under the WASM marker expression.

Validation: full tests and coverage each pass 13,925 tests (130 skipped, 2 expected failures); 268 doctests pass; two full lint runs and commit hooks pass; all 25 affected API pages render successfully. Both CLI entry-point checks pass after committing.

d-chambers:
Regarding the [incompatible-unit finding](https://github.com/DASDAE/dascore/pull/1194#pullrequestreview-5271132300): no additional catch is needed. [`convert_units`](https://github.com/DASDAE/dascore/blob/4aced569ab6e55b905bfae24cf8c5fd1b0cc204c/dascore/units.py#L260-L263) already catches Pint's `DimensionalityError` and raises DASCore's `UnitError`, so the explicit-bound path preserves the public exception contract.

I verified all three public APIs (`select`, `chunk`, and `chunk_plan`) with seconds on a metre coordinate and metres on a timedelta coordinate. All six cases raise `UnitError`; both chunk APIs include the incompatible source and target units in the message. I am leaving the shared conversion behavior unchanged rather than duplicating its exception handling here.

dascore/utils/explicit_ranges.py
```diff
@@ -0,0 +1,381 @@
...
+        if coord is not None:
+            out[patch_row] = coord
+            continue
+        coords = file_source_coords(resolver, row, files)
+        if coords is not None and name in coords.coord_map:
+            out[patch_row] = coords.coord_map[name]
+    return out
+
+
+def _planned_coordinate(resolver, row, name):
```
d-chambers: `_planned_coordinate` and `_planned_manager` run the same member replay: source lookup → parent residuals → `_modified` trim with unit conversion → convert to the output unit → `_get_merged_coord`. As a check, I forced `_planned_coordinate` to always take its `_planned_manager` branch (line 314). `test_spool_explicit_ranges.py`, `test_spool.py` and `test_spool_gaps.py` still pass (350 tests). So either the dim-only path is redundant, or the thing it does differently isn't tested. That difference is probably using `coord_from_row` via `known_coordinates` so it doesn't scan the file.

Suggestion: keep one replay (`_planned_manager`) and make the fast path only the `coord_from_row` shortcut in front of it. If the fast path is there for performance, add a test that fails when it scans, e.g. monkeypatch `dc.scan_payloads` to raise. That removes about 60 lines and a second copy of the unit/trim logic that could drift.

dascore/utils/explicit_ranges.py
```diff
@@ -0,0 +1,381 @@
...
+
+def _source_manager(resolver, row, files=None):
+    """Recover all source coordinates without loading measurement arrays."""
+    path = str(row.get("source_path") or "")
+    live = resolver.live_entries().get(path)
+    if live is not None:
+        return live.coords
+    if path.startswith("plan://"):
+        plans = getattr(resolver, "plans", {})
+        plan = next(
```
d-chambers, line 189: This prefix lookup (`next(item for prefix, item in plans.items() if path.startswith(prefix))`) is repeated in `_planned_coordinate` (lines 305-310). The two also handle a resolver that is itself a plan slightly differently (`hasattr(resolver, "member_rows")` before vs. after the lookup). A small `_plan_for(resolver, path)` helper would give both one rule.

dascore/utils/explicit_ranges.py
```diff
@@ -0,0 +1,381 @@
...
+            native = coords.coord_map[dim].units
+            unit = member.get(f"_{dim}_units")
+            if native is not None and unit is not None and not pd.isnull(unit):
+                if str(native) != str(unit):
+                    low, high = (
+                        convert_units(x, to_units=native, from_units=unit)
+                        for x in (low, high)
+                    )
+            coords, _ = coords.select(**{dim: (low, high)})
+        unit = member.get(f"_{plan.dim}_units")
```
d-chambers: Unit handling repeats here. `native`/`unit` were already computed inside the `_modified` branch above, and `coords.select` doesn't change units, so read them once before the `if`.

More broadly, the pattern *read `_{dim}_units` → treat NaN as None → compare `str()` to the native unit → convert* appears about 7 times in this PR: here, 224-225, 290-291, 324-325, 350-365, and twice in `exact_coordinate_bounds`. `planned._stated_units` already does the NaN→None step. With that plus a `_units_differ(a, b)` predicate, most of these become one line each.

dascore/utils/explicit_ranges.py
```diff
@@ -0,0 +1,381 @@
...
+    for index in range(len(candidate)):
+        low, high = candidate[index, 0], candidate[index, 1]
+        bounds: list[object] = []
+        for bound in (low, high):
+            if isinstance(bound, np.str_):
+                bound = str(bound)
+            if bound is None or bound is Ellipsis:
+                msg = f"Explicit range row {index} needs two closed bounds."
+                raise ParameterError(msg)
+            value = bound.magnitude if isinstance(bound, Quantity) else bound
```
d-chambers: Nit: `value` shadows the function's `value` argument. That's harmless now because the argument isn't used after `candidate` is built, but it's a trap for the next edit. `magnitude` would read better.

dascore/io/index/explicit.py
```diff
@@ -0,0 +1,162 @@
...
+            rows = []
+            pieces = []
+            requests = []
+            for request, (bounds, candidate) in enumerate(
+                zip(self.ranges.rows, candidate_frames, strict=True)
+            ):
+                for _, projected in candidate.iterrows():
+                    row = by_id.loc[projected["_patch_row"]]
+                    coord = known.get(row["_patch_row"])
+                    if coord is not None:
```
d-chambers: Small simplifications:
- Raise first as a guard clause (`if coord is None: raise MissingPatchError(...)`) and then `actual = exact_coordinate_bounds(...)`. That removes the `if/else` with the `continue` buried in the first branch.
- Line 102: `source.drop_duplicates("_patch_row")` can't drop anything after the `is_unique` assert on line 63, so pass `source` directly.
- Line 112: `self._source_revision = revision` (the lock is held, so the value is the same one already read).

dascore/io/index/planned.py
```diff
@@ -808,6 +830,20 @@ def resolve(self, row: Mapping, **trim) -> dc.Patch:
                 )
         return self._stamp(patch, row)
 
+    def _anchor_metadata_coords(self, row: Mapping) -> CoordManager | None:
```
d-chambers: This reimplements the live/file branch of `explicit_ranges._source_manager`. I replaced the body with `coords = _source_manager(self.loader, row)` and the explicit, spool, gaps and `tests/test_io/test_index` suites still pass (1114 tests).

`_source_manager` also resolves `plan://` anchors from metadata. That might let `_sibling_coords` drop its `self._load_member(...)` fallback, which reads measurement data despite the docstring promising coordinates only. That's worth checking with a nested-plan fill test before removing it, since I don't think one exists.

dascore/io/index/planned.py
```diff
@@ -965,13 +1013,29 @@ def derived_catalog(
...
+                anchor_sources.get("source_path", pd.Series(dtype=str)).astype(str)
+            )
         loader.absorb(parent.resolver, paths=member_paths)
     coord_dims_map = {} if parent is None else parent.backend.coord_dims_map()
     # a coordinate riding a dimension it is not named for, on any member,
     # is associated; the array path rebuilds dimension coordinates only
     aux_coords = frozenset(
         () if parent is None else parent.backend.associated_coord_names()
     )
+    anchors = sources
```
d-chambers, line 1013: This block copies the `member_rows` absolutization at lines 1001-1010. Absolutizing `sources` once, up front, would give one helper and would also make the anchor paths passed to `loader.absorb(...)` at ~1016 use the same absolute spelling as `member_paths`. Right now they come from the unabsolutized `sources`.

Both copies also test absolutes with `str(p).startswith("/")`, while `catalog.resolve_path` and `_absolutize_record` use `Path(p).is_absolute()`. A shared helper using the latter would also handle Windows paths. The first copy predates this PR; the second one repeats that behavior.

dascore/utils/chunk_plan.py
```diff
@@ -1573,15 +1593,38 @@ def build_chunk_plan(
         missing_dim=missing_dim,
         group=_resolve_group_attrs(group, set(df.columns)),
         sampling_group_tolerance=dc.get_config().sampling_group_tolerance,
+        on_incomplete=on_incomplete,
     )
     if df.empty:
+        if explicit is not None and explicit.rows:
+            _report_incomplete(
```
d-chambers: These two `_report_incomplete` blocks are identical to each other, and so are the two early returns. They could collapse to:
```python
if not df.empty:
    df = _prepare_relation(df, name, missing_dim)
if df.empty:
    if explicit is not None:
        _report_incomplete([...], on_incomplete)
    ...
    return ChunkPlan(outputs, empty_members, name, value, params)
```
(`explicit.rows` being empty already makes `_report_incomplete` a no-op, so the `and explicit.rows` guard isn't needed.)

dascore/utils/chunk_plan.py
```diff
@@ -1792,6 +1888,31 @@ def build_chunk_plan(
...
+            columns=[
+                min_name,
+                max_name,
+                "output_id",
+                "_request_row",
+                "_compat_group",
+                "_partition",
+            ]
+        )
+        outputs, empty_members = _finish_explicit_plan(
```
d-chambers, line 1887: `_finish_explicit_plan` is called twice with the same 11 positional arguments (here and at 1991). The empty branch also keeps a hand-written column list that has to match the columns assigned at 1958-1966. It already omits `_anchor_patch_row`.

One option: when `explicit is not None and not fed_counts.sum()`, build the empty outputs/members and fall through to a single `_finish_explicit_plan` call placed just before the final `return`. At minimum, pass keyword arguments. Several of these are adjacent booleans/`None`s (`keep_partial, conflict, on_incomplete, _exact_coords, fill_value`), so a swapped argument would pass silently.

dascore/utils/chunk_plan.py
```diff
@@ -2501,3 +2645,314 @@ def build_subdivision_plan(df: pd.DataFrame, pieces, name: str) -> ChunkPlan:
...
+    for label, sub in cells:
+        groups.append((label, sub[min_name].min(), sub[max_name].max(), sub))
+    groups.sort(key=lambda item: (item[1], str(item[0])))
+    rank = {label: pos for pos, (label, *_rest) in enumerate(groups)}
+    accepted = set()
+    failures = []
+    for request, bounds in enumerate(requests.rows):
+        applicable = False
+        for label, start, stop, sub in groups:
+            unit = _partition_unit(sub, name, 0)
```
d-chambers: In the `request × group` loop, everything except the request bounds depends only on the group: `_partition_unit(sub, ...)`, the `datetime64`/`timedelta64` coercion of `start`/`stop`, `step`, `missing_required`, `known` and `no_grid`. Computing them once when `groups` is built (line ~2745) makes the loop body shorter and removes a quirk: `start, stop` are rebound on every iteration, which only works because the coercion happens to be idempotent.

This is the same shape as the CodeRabbit note that was already addressed in `build_chunk_plan` (`part_coords`), but for `_finish_explicit_plan`.

dascore/utils/chunk_plan.py
```diff
@@ -2501,3 +2645,314 @@ def build_subdivision_plan(df: pd.DataFrame, pieces, name: str) -> ChunkPlan:
...
+                    continue
+                low = min(x[0] for x in selected)
+                high = max(x[1] for x in selected)
+            elif known.all() and selected and fill_value is None:
+                # Exact member labels obey the existing snap/assembly rules.
+                # Another partition in the group may have a shifted origin,
+                # and a joined source may have sub-sample jitter.
+                low = min(x[0] for x in selected)
+                high = max(x[1] for x in selected)
+                expected_low, _, _ = _grid_snapped(
```
d-chambers: The two `_grid_snapped` calls differ only in `origin`, and `_grid_snapped` is plain numpy arithmetic, so one call with an array origin works:
```python
exp_lo, exp_hi, _ = _grid_snapped(
    np.full(2, requested_low), np.full(2, requested_high), np.array([low, high]), abs(step)
)
if exp_lo[0] < low or exp_hi[1] > high:
```

d-chambers (commented):
Review focused on simplifying the code and reducing duplication without losing robustness. The feature and tests look solid: `test_spool_explicit_ranges.py` passes locally (77 tests). Most of the added complexity comes from logic that is repeated rather than from the feature itself. In rough order of payoff:

1. **Coordinate recovery in `explicit_ranges.py`.** `_planned_coordinate` and `_planned_manager` duplicate the member replay, and the plan-prefix lookup and unit-normalization idioms are repeated throughout. Forcing everything through `_planned_manager` still passes the suite, so either about 60 lines can go or a fast-path test is missing.
2. **Explicit branch of `build_chunk_plan` / `_finish_explicit_plan`.** There are duplicated empty-source blocks, two 11-positional-argument calls, a repeated "exact envelope" computation (4 copies), and group-invariant work redone inside the request loop.
3. **`planned.py`.** `_anchor_metadata_coords` can reuse `_source_manager` (also verified: 1114 tests pass), and the path-absolutization block is now copied twice with a `startswith("/")` check that differs from `Path.is_absolute()` used elsewhere.
4. **`spool.py`.** `select`'s `possible` pre-check duplicates `explicit_ranges`' shape detection, and `chunk`/`chunk_plan` now repeat the same planning preamble.

None of these change behavior. The two that I verified by patching a local checkout are marked inline.

dascore/utils/chunk_plan.py
```diff
@@ -1663,7 +1707,53 @@ def build_chunk_plan(
                 break
             if diag is not None:
                 size_diagnostics.append({"first_output_id": next_id, **diag})
-        if merge_mode:
+        if explicit is not None:
+            unit = _partition_unit(sorted_df, name, seg_starts[part])
+            # Source coordinates are invariant across requests in this partition.
+            part_coords = (
```
d-chambers, line 1703: The pattern *run `exact_coordinate_bounds` over known source coords, drop the `None`s, take min of lows and max of highs* appears here (1731-1737), twice in `_finish_explicit_plan` (the `selected` list and the `no_grid` / `known.all()` branches), and in `ExplicitSelectCatalog`. A helper `_exact_envelope(coords, bounds, unit) -> tuple | None` would give one implementation of the margin/unit rules and remove the three repeated `low = min(x[0] ...); high = max(x[1] ...)` pairs.

dascore/core/spool.py
```diff
@@ -535,7 +544,68 @@ def select(
...
+        """
+        # Explicit windows are independent requests, so they need separate
+        # plan outputs even when they name the same source samples.
+        raw = dict(kwargs)
+        if isinstance(_coords, Mapping):
+            raw.update({str(k): v for k, v in _coords.items()})
+        coord_names = self._catalog.backend.coord_names()
+        attr_names = self._catalog.backend.attr_names()
+        tagged = selector_spec_names(_coords)
+        possible = any(
```
d-chambers, line 561: `possible` re-implements the shape detection in `explicit_ranges` (non-1D ndarray, or a non-empty list/tuple whose first element is a list/tuple/ndarray). If the two diverge, `select` will either skip real explicit input or run `_classify_query` for nothing. A non-raising `looks_explicit(value)` predicate in `explicit_ranges.py`, used both here and at the top of `explicit_ranges()`, would keep them in sync.

A one-line comment saying this pre-check exists to avoid `_classify_query` on the common path would also help.

dascore/core/spool.py
```diff
@@ -1684,6 +1754,12 @@ def _plan_frames(
                 base = base[base["_patch_row"].isin(working["_patch_row"])]
         return base.reset_index(drop=True), working.reset_index(drop=True)
 
+    def _known_chunk_coords(self, source_rows, name: str | None, value) -> dict:
```
d-chambers: `_known_chunk_coords` parses the ranges only to get a bool, and `build_chunk_plan` then parses them again. More importantly, `chunk` and `chunk_plan` now share the same ~20 lines: `name = next(iter(kwargs))` → `_plan_frames` → `build_chunk_plan(..., on_incomplete=..., _exact_coords=...)` → `coalesce_runs`. A private `_build_chunk_plan(**params) -> (source_rows, working, plan)` used by both would mean the next chunk option is added in one place, and this helper could be folded into it.

dascore/utils/explicit_ranges.py
d-chambers: Consolidated member replay into `_planned_manager` with optional coordinate projection in a66381b6. Regular file grids still use `coord_from_row` before scanning. `test_regular_file_derived_window_uses_indexed_grid` now fails if nested selection or planning calls `scan_payloads`, and it checks the resulting metadata and loaded samples.

dascore/utils/explicit_ranges.py
d-chambers: The duplicate prefix lookup disappeared with `_planned_coordinate`; `_source_manager` now owns the single plan-dispatch rule. This is included in a66381b6.

dascore/io/index/planned.py
d-chambers: Fill anchors now use `_source_manager`, including nested-plan metadata recovery. I retained the loader fallback because a valid single-patch legacy record can have a blank key while current scan metadata has a native key. `test_fill_only_legacy_blank_key_loads_anchor` exercises that path with real DASDAE files and persisted records, checking filled data, bounds, and an associated coordinate. All changed production modules have 100% local coverage.

dascore/io/index/planned.py
d-chambers: Source paths are now normalized once before deriving members and anchors, using `FileResolver.resolve_path`. This reuses the existing URI, absolute-path, and rooted-relative-path rules, and both inputs to `loader.absorb` use the resolved spelling. Included in a66381b6.

dascore/utils/explicit_ranges.py
d-chambers: Addressed in [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30): native and planner units are read once per recovered member and reused for trimming and normalization. The earlier consolidation already removed the duplicate coordinate replay, so I kept this local rather than adding unit helpers across the planner and resolver modules.

dascore/utils/explicit_ranges.py
d-chambers: Addressed in [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30): the scalar extracted from a bound is now named `magnitude`, so it no longer shadows the input `value`.

dascore/io/index/explicit.py
d-chambers: Addressed the guard clause and redundant deduplication in [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30). The existing uniqueness assertion establishes that `source` can be passed directly. I retained the final revision read: realizing `PatchCatalog.backend` can call `ensure_updated()` and invalidate the parent while this reentrant lock is held. `PatchCatalog.to_df()` also rereads the revision for that reason. A comment now explains why the earlier captured revision is insufficient.

dascore/utils/chunk_plan.py
d-chambers: Addressed in [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30): nonempty input is prepared first, and both empty-input cases now share one incomplete-request report and one empty-plan return. The original empty output/member columns and ordinary chunk behavior are preserved.

dascore/utils/chunk_plan.py
d-chambers: Addressed the positional-argument risk in [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30): both `_finish_explicit_plan` calls now name their arguments. I kept the empty-output return separate because the ordinary assembly path concatenates genuinely empty contribution lists and expects source-backed rows. Folding that path together would need additional schema handling and indentation; the two explicit keyword calls keep the distinction small and visible.

dascore/utils/chunk_plan.py
d-chambers: Addressed in [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30): each compatibility group now computes its normalized envelope, units, step, coordinate availability, and exact-coordinate list once. Original group ordering is retained; request-specific applicability, bound conversion, coordinate selection, and snapping stay inside the request loop.

dascore/utils/chunk_plan.py
d-chambers: Addressed in [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30): the lower and upper origins are passed together to one `_grid_snapped` call, and the corresponding lower/upper results are checked. The temporal and numeric coordinate checks retain their existing results.

dascore/utils/chunk_plan.py
d-chambers: Addressed in [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30): `_exact_envelope` is shared by partition-level and request/group-level aggregation. Each caller still supplies its own bounds. Individual source selection and member trimming keep their separate exact-coordinate calls because their bounds and ownership rules differ.

dascore/core/spool.py
d-chambers: Addressed in [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30): `select` and explicit-range validation share the cheap `looks_explicit` predicate. Selection retains its ordinary empty-list/tuple path, and chunk validation retains its existing malformed-array errors. Scalar and ordinary-range selectors still avoid the additional namespace-classification pass.

dascore/core/spool.py
d-chambers: Addressed the repeated setup in [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30): a private `_build_chunk_plan` now prepares source/working rows, obtains exact metadata when needed, builds the plan, and coalesces runs for both public methods. It returns only source rows and the plan. The builder still receives the original input value so `ChunkPlan.value` and validation behavior remain unchanged; I did not introduce an internal parsed wrapper into that boundary just to eliminate the second validation call.

dascore/utils/explicit_ranges.py
```diff
@@ -0,0 +1,320 @@
...
+                    )
+            coords, _ = coords.select(**{dim: (low, high)})
+        unit = member.get(f"_{plan.dim}_units")
+        native = coords.coord_map[plan.dim].units
+        if native is not None and unit is not None and not pd.isnull(unit):
+            if str(native) != str(unit):
+                coords = coords.convert_units(**{plan.dim: unit})
+        recovered.append(coords)
+    if not recovered:
+        assert not projected, "plan output has no members or reconstructable grid"
```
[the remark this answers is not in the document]
d-chambers: Fixed in [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30): an unavailable projected fill-only grid raises `ChunkError` with the same message. Both existing public-API regression expectations now require that exception. The select and chunk-plan cases also pass in a separate `python -O` process, so the guard no longer disappears under optimization.

d-chambers:
Addressed the [simplification review](https://github.com/DASDAE/dascore/pull/1194#pullrequestreview-5277640329) in [a66381b6](https://github.com/DASDAE/dascore/commit/a66381b637932242c0a028956b6e305eacbaafe5) and [f023d5e8](https://github.com/DASDAE/dascore/commit/f023d5e8512e4f187bf666c6fc3e890e2dee5c30). Coordinate recovery now uses one replay path; chunk and chunk-plan share preparation; contributor metadata and request-group work are computed once; aggregate envelope calculation and source-path normalization are shared. Production code is 82 lines shorter than `4aced569`. All 77 original explicit-window cases remain, with seven additional regression cases; the test file grew by 85 net lines.

Each inline suggestion has a reply explaining the change or the retained distinction. In particular, catalog initialization can advance the revision while the lock is held, so the final revision read remains. Individual member/source bounds and the empty/nonempty finish paths also retain their separate semantics.

The [projected fill-only error finding](https://github.com/DASDAE/dascore/pull/1194#pullrequestreview-5278688128) is fixed: unsupported projected grids raise `ChunkError` with the same message, including under `python -O`.

Validation: full tests and coverage each pass 13,932 tests (130 skipped, two expected failures); all 268 doctests pass. Two full lint runs, commit hooks, both CLI entry-point checks, and all 45 changed API page renders pass. API generation reports 5,336 valid links and no broken links. The before/after public behavior and metadata comparisons match apart from the intentional exception-type improvement.

d-chambers:
@coderabbitai review

dascore/utils/chunk_plan.py
```diff
@@ -2501,3 +2630,336 @@ def build_subdivision_plan(df: pd.DataFrame, pieces, name: str) -> ChunkPlan:
...
+        groups.append(
+            dict(
+                label=label,
+                start=start,
+                stop=stop,
+                unit=_partition_unit(sub, name, 0),
+                step=get_middle_value(sub[step_name].to_numpy()),
+                missing_required=any(coord is None for coord in coords),
+                all_known=len(coords) == len(sub)
+                and all(x is not None for x in coords),
```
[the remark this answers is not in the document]
d-chambers: Fixed in [2b3ac8b6](https://github.com/DASDAE/dascore/commit/2b3ac8b6), now pushed with the updated branch. Coordinate availability is checked only for continuity partitions touched by the request, so an unavailable disconnected partition no longer vetoes a valid window.

The check retains every supporting row within those partitions. Restricting it to raw source-envelope overlap would miss fill-only gap anchors and overlap ownership. The regressions use real indexed files with a legacy blank source key and cover both the valid disconnected request and the cases that must still follow the incomplete-request policy.

Revalidated after merging dev at `b03b2f3a`: all 108 explicit-window tests pass, and the full suite with coverage passes 14,418 tests (130 skipped, two expected failures).

dascore/io/index/planned.py
```diff
@@ -942,9 +963,22 @@ def _sibling_coords(self, output_id: int):
...
+            if output_id not in self._fill_coords:
+                anchor = self._anchor_rows[anchor_id]
+                coords = _source_manager(self.loader, anchor)
+                if coords is not None:
+                    coords = _select_manager(coords, self.parent_residuals)
+                if coords is None:
+                    # Unsupported metadata sources still have a patch loader.
+                    coords = self._load_member({**anchor, "_modified": False}).coords
+                self._fill_coords[output_id] = coords
+            return self._fill_coords[output_id]
```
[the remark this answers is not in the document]
d-chambers: Fixed in [69a2dcef](https://github.com/DASDAE/dascore/commit/69a2dceffe69d50ce3a251495b7375ccbeb71f9f).

Confirmed the failure when the first source of a later fill partition uses feet and the plan uses metres. Fill-anchor coordinates now convert to the output row's units after either metadata recovery or the legacy patch-loading fallback, before being cached. The existing coordinate-manager conversion handles already-matching units without duplicating conversion rules.

The regression covers live patches, indexed files with native keys, and legacy blank-key files that exercise the load fallback. Each returns two samples at 15–16 m with fill value -1; all three fail against the previous function.

dascore/utils/explicit_ranges.py
```diff
@@ -0,0 +1,333 @@
...
+            # Float envelopes omit the expression that rounded each label.
+            # Only integer/tick grids can supply exact samples from the index.
+            if (
+                coord is not None
+                and isinstance(dtype, str)
+                and np.dtype(dtype).kind in "iuMm"
+            ):
+                return _one_coordinate(projection, coord)
+        coords = file_source_coords(resolver, row, files)
+    return None if coords is None else _project_manager(coords, projection)
```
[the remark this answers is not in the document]
d-chambers: Fixed in [69a2dcef](https://github.com/DASDAE/dascore/commit/69a2dceffe69d50ce3a251495b7375ccbeb71f9f).

Confirmed both paths. Metadata recovery now uses the matching route from plan_entries(), so a nested row reaches its owning plan. File-backed collapsed rows scan through the plan's loader, which supplies file path resolution.

The new regressions cover regular floating and uneven file coordinates, plus a nested time-then-distance plan re-planned along distance. They verify plan bounds, loaded coordinate labels, and data. Against the previous function, the file cases raise AttributeError and the nested case incorrectly raises ChunkError; all three pass with the fix.

d-chambers:
Addressed the anchor-retention suggestion in the [latest review](https://github.com/DASDAE/dascore/pull/1194#pullrequestreview-5295123100). The source subset referenced by _anchor_patch_row is now computed once and shared by loader absorption and PlanResolver. Plans without anchor IDs pass None. This avoids keeping a record dictionary for every unrelated source without adding a second filtering pass.

Included in [69a2dcef](https://github.com/DASDAE/dascore/commit/69a2dceffe69d50ce3a251495b7375ccbeb71f9f). Validation: 728 affected tests, 14,424 full-suite tests (also passing with coverage), 267 doctests, and two full lint passes. All changed production lines are covered.

